PR #33, opened 2026-03-10
Harden workflow modules against re-entry and add envelope-level dedup for LLM dispatch

eanzhao:
### Problem

Workflow modules (`WorkflowExecutionKernel`, `LLMCallModule`, `DelayModule`, `WaitSignalModule`) are susceptible to duplicate or inconsistent execution under re-entry scenarios:

1. **Crash between schedule and state persistence**: Callback scheduling (`ScheduleSelfDurableTimeoutAsync`) and state save are not atomic. If the grain crashes after the callback is registered but before the lease is persisted, recovery cannot match the incoming callback — leading to orphaned timeouts or lost progress.
2. **Watchdog / timeout re-fires**: A watchdog or timeout callback re-enters the module with a different inbound envelope. Without a stable dedup key tied to the logical operation (rather than the envelope), the receiving actor treats retried dispatches as distinct requests — causing duplicate LLM calls or duplicate step dispatches.
3. **Pending dispatch not tracked**: `WorkflowExecutionKernel` had no awareness of whether a step dispatch was in-flight; re-activation would attempt to dispatch the same step again.
4. **Eager actor resolution**: `WorkflowRunCommandTargetResolver` resolved (or created) an actor before checking whether projection was enabled, wasting resources on a path that would immediately fail.

#### Commit 1 — State-first persistence and callback-ID fallback matching

Reorder all module scheduling paths to **persist state (including callback ID) before scheduling the durable callback**, so that even if the grain crashes between schedule and lease persistence, recovery can match by callback ID.

**WorkflowExecutionKernel**:
- New proto fields: `current_step_dispatch_pending` (field 9), `current_step_timeout_callback_id` (field 10) on `WorkflowExecutionKernelState`; `callback_id` (field 4) and `dispatch_pending` (field 5) on `RetryBackoffState`.
- `ResumePendingCurrentStepDispatchAsync` detects and resumes in-flight step dispatches on re-entry (timeout, step-completed, retry-backoff handlers all check for pending dispatch first).
- Retry backoff scheduling now persists `CallbackId` + `DispatchPending` flag before scheduling, and clears on completion.

**LLMCallModule**:
- Split monolithic dispatch into explicit phases: create pending → schedule watchdog → dispatch request → mark dispatched.
- New proto fields on `PendingLlmCallState`: `request_dispatched` (field 5), `watchdog_callback_id` (field 6).
- `TryResolvePending` resumes existing pending state on re-entry instead of creating a duplicate.
- `MatchesWatchdog` falls back to callback-ID comparison when lease is missing.

**DelayModule**:
- New proto field: `callback_id` (field 3) on `PendingDelayState`.
- State-first save ordering: persist `PendingDelayState` with callback ID → schedule callback → update lease.
- `MatchesPendingCallback` falls back to callback-ID matching.

**WaitSignalModule**:
- New proto field: `timeout_callback_id` (field 6) on `PendingSignalState`.
- Same state-first pattern; `MatchesTimeout` falls back to callback-ID matching.
- Extracted `CancelPendingAsync` for consistent cancel-before-reschedule.

**WorkflowRunCommandTargetResolver**:
- Check `ProjectionEnabled` before `ResolveOrCreateAsync` (fail-fast, avoid unnecessary actor creation).

#### Commit 2 — Generic metadata passthrough and envelope-level dedup

Even with the `RequestDispatched` flag from commit 1, the non-atomic gap between `DispatchChatRequestAsync` and `MarkRequestDispatchedAsync` can still produce duplicate envelopes that reach the target actor. This commit closes that gap at the receiving side.

**Generic metadata channel**:
- Add optional `IReadOnlyDictionary<string, string>? metadata` parameter to `IEventContext.PublishAsync/SendToAsync` and `IEventPublisher.PublishAsync/SendToAsync`.
- Forwarded through `GAgentBase`, `EventHandlerContext`, `WorkflowExecutionContextAdapter` → publisher implementations → `EnvelopePublishContextHelpers.ApplyPublishContext`, which merges the dictionary into `envelope.Metadata`. Callers that don't pass metadata are unaffected (`null` → no-op).

**DedupOriginId**:
- New metadata key: `EnvelopeMetadataKeys.DedupOriginId` (`__dedup_origin_id`).
- `LLMCallModule` generates a stable `DispatchDedupId` per chat session (proto field 7 on `PendingLlmCallState`), attaches it to outbound `ChatRequestEvent`.
- `RuntimeActorGrain.BuildDedupKey` priority: `DedupOriginId` → `RetryOriginEventId` → `envelope.Id` — all re-entries for the same logical dispatch collapse to the same dedup key.

### Changes

| Category | Files | What changed |
|---|---|---|
| Abstractions | `IEventContext`, `IEventPublisher`, `EnvelopeMetadataKeys` | `metadata` param + `DedupOriginId` key |
| Adapters | `GAgentBase`, `EventHandlerContext`, `WorkflowExecutionContextAdapter` | Metadata pass-through |
| Runtime impl | `LocalActorPublisher`, `OrleansGrainEventPublisher` | Pass-through to `ApplyPublishContext` |
| Envelope assembly | `EnvelopePublishContextHelpers` | Merge caller metadata into envelope |
| Dedup consumer | `RuntimeActorGrain.BuildDedupKey` | Prioritise `DedupOriginId` |
| Workflow kernel | `WorkflowExecutionKernel` | Pending-dispatch tracking, state-first persistence, callback-ID fallback |
| Workflow modules | `LLMCallModule`, `DelayModule`, `WaitSignalModule` | State-first scheduling, callback-ID fallback, dedup ID |
| Command resolution | `WorkflowRunCommandTargetResolver` | Fail-fast on projection disabled |
| Proto | `workflow_state.proto` | 8 new fields across 5 messages |
| Tests | ~20 files | Signature adaptation + new re-entry / dedup coverage |

### Test plan

- [ ] `dotnet build aevatar.slnx --nologo`
- [ ] `dotnet test aevatar.slnx --nologo`
- [ ] `bash tools/ci/architecture_guards.sh`
- [ ] `bash tools/ci/workflow_binding_boundary_guard.sh`
- [ ] `bash tools/ci/test_stability_guards.sh`
- [ ] Verify `RuntimeCallbackEventizationTests` covers watchdog-retry and dedup paths


src/workflow/Aevatar.Workflow.Projection/Orchestration/WorkflowRunEventSessionCodec.cs
```diff
@@ -11,61 +15,326 @@ public sealed class WorkflowRunEventSessionCodec : IProjectionSessionEventCodec<
...
+            return true;
         }
         catch (JsonException)
         {
-            return null;
+            return false;
+        }
+    }
+
+    private static bool TryDeserializeLegacyJson(string eventType, string json, out WorkflowRunEvent? evt)
```
loning: 序列化不要用json吧

PR #181, opened 2026-04-13
Refine scripting accepted and visibility semantics

### LINKED ISSUE #158: [P0] 从 scripting 写路径中移除 authority read-model activation
Parent Epic: #157

## 目标
从 scripting 写路径中移除 authority read-model activation。

## 当前问题
definition command service、catalog command service、runtime capability 路径会直接调用 authority projection activation，导致写路径开始控制 read-model 生命周期。

## 范围
- 从 scripting 写路径 / runtime 服务中移除对 `IScriptAuthorityReadModelActivationPort` 的直接使用
- 保持现有功能仍然通过正常 event / projection 主链完成
- 明确新的成功路径说明

## 验收标准
- definition / catalog command service 不再直接触发 authority read-model activation
- runtime capability 的 create 路径不再直接触发 authority read-model activation
- 不依赖显式 projection activation，写路径仍可正常工作

### LINKED ISSUE #159: [P0] 去掉 catalog promote/rollback 成功路径中的 query-port 追平依赖
Parent Epic: #157

## 目标
去掉 catalog promote / rollback 成功确认路径中对 query-port 追平的依赖。

## 当前问题
catalog promote / rollback 现在会先 dispatch command，然后轮询 query / read-model 状态来判断是否成功。这会把 command 成功语义与 projection 延迟、read-model 可用性直接绑在一起。

## 范围
- 用 interaction receipt、事件确认，或其他写侧确认机制替换 query-port 轮询
- 让成功语义保持明确且可测试

## 验收标准
- promote / rollback 不再通过轮询 query port 判断成功
- command 成功语义与 read-model 追平解耦
- 当无法获得写侧确认时，失败模式是明确可解释的

louis4li:
## What changed
- narrowed scripting write-path success semantics to `accepted for dispatch` and removed write-side coupling to read-model catch-up
- added stable accepted receipts plus save-observation flow for scope/studio save APIs
- updated Studio web and CLI flows so promotion only shows as applied after the existing catalog read model reflects the target revision
- removed wall-clock supersede inference from save observation and switched to durable catalog facts
- aligned the CQRS design doc with the shipped behavior, including reusing existing catalog read APIs instead of adding a dedicated promotion observation endpoint

## Why
The write path had drifted back toward read/write coupling in two places: backend services previously waited for query-visible catalog state, and UI flows treated an accepted promotion decision as if the catalog had already switched. That made `accepted`, `applied`, and `query-visible` semantically blurry and introduced a user-visible race.

## Impact
- command endpoints now honestly return accepted handles instead of implying read freshness
- save and promotion UX now distinguishes "accepted" from "catalog visible"
- read-side visibility continues to come from the existing read-model interfaces

## Root cause
The catalog promote/save paths were mixing write-side dispatch acknowledgement with read-side materialization state. Save observation also used cross-service wall-clock ordering, which is not a reliable authority signal.

## Validation
- `dotnet test test/Aevatar.GAgentService.Integration.Tests/Aevatar.GAgentService.Integration.Tests.csproj --filter ScopeScriptApplicationServicesTests --nologo`
- `pnpm tsc --noEmit` in `apps/aevatar-console-web`
- `pnpm jest src/modules/studio/scripts/ScriptsWorkbenchPage.test.tsx --runInBand` in `apps/aevatar-console-web`

## Issues
- Fixes #158
- Fixes #159


eanzhao:
I don't think the authority-activation dependency has actually been removed yet.

The services were changed to accept `IScriptAuthorityReadModelActivationPort?`, but the default host still registers that port in `AddScriptingProjectionComponents()`, and the write/runtime paths still call `ActivateAsync(...)` when it is present (`RuntimeScriptDefinitionCommandService`, `RuntimeScriptCatalogCommandService`, `ScriptBehaviorRuntimeCapabilities`). In the default production graph this branch still executes, so the write path is still controlling authority projection lifecycle.

To match the design goal here, I think the activation dependency/calls need to be removed from these paths entirely, or the default graph needs to stop wiring the port into them.

eanzhao:
`saveCurrentDraftToScope` captures `resolvedScopeId`, but `resolvedScopeId` is not part of the callback dependency list in `ScriptsWorkbenchPage.tsx`.

This page is rendered without a scope-based remount key from `pages/studio/index.tsx`, so after switching scope the callback can still PUT to the previous scope until something else recreates it. That makes the save path vulnerable to writing into the wrong scope.

I think `resolvedScopeId` needs to be included in the callback dependencies, or the callback needs to resolve the current scope at call time instead of closing over an old value.

eanzhao:
The new save-observation polling treats the first transient query error as a hard save failure.

`observeAcceptedSave` in the web UI and the analogous loop in the CLI both call the observation endpoint without the transient-error retry handling that promotion polling already has. A single timeout / 5xx from `/save-observation` can therefore surface as "save failed" even though the accepted save becomes visible moments later.

I think this should handle transient observation failures the same way promotion polling does: retry within the observation window, and only fail after the retry budget is exhausted.

louis4li:
Addressed the three scripting review comments; the latest pushed commit is `4aa3a760` after rebasing onto the updated branch tip.

1. Removed the remaining authority read-model activation coupling from the write/runtime paths. `RuntimeScriptDefinitionCommandService`, `RuntimeScriptCatalogCommandService`, and `ScriptBehaviorRuntimeCapabilities` no longer accept or call `IScriptAuthorityReadModelActivationPort`, so the write path is no longer controlling authority projection lifecycle.
2. Fixed the Studio save callback to track the latest `resolvedScopeId`, so switching scope no longer leaves `saveCurrentDraftToScope` closing over the previous scope id.
3. Added transient-error retry handling to save observation polling in both Studio web and the CLI frontend, matching the promotion polling behavior. A temporary timeout / 5xx during `/save-observation` no longer immediately surfaces as a save failure if the save becomes visible within the observation window.

Validation run after rebasing onto the latest branch state:
- `dotnet test test/Aevatar.Scripting.Core.Tests/Aevatar.Scripting.Core.Tests.csproj --filter "ScriptAgentLifecycleCapabilitiesTests|ScriptingBranchCoverageTests" --nologo`
- `pnpm jest src/modules/studio/scripts/ScriptsWorkbenchPage.test.tsx --runInBand`
- `pnpm tsc --noEmit` in `apps/aevatar-console-web`

Earlier in the same fix cycle I also ran:

eanzhao (approved):
LGTM

PR #274, opened 2026-04-21
Add channel RFC §14.1 CI guard suite (bash + Architecture.Tests)

### LINKED ISSUE #265: [Channel RFC] CI guards (bash grep + Roslyn analyzer / Architecture.Tests)
**Parent RFC**: #254

## Scope

RFC §14.1 CI guards 分两层落地。

### Layer 1: bash grep guards（`tools/ci/*.sh`）

4 个纯字符串匹配规则，集成到 `architecture_guards.sh`：

1. **`channel_native_sdk_import_guard.sh`**
   - 禁止 `using Larksuite.Oapi` / `using Telegram.Bot` / `using SlackNet` / `using Discord.*` 出现在 `src/**` 和 `agents/Aevatar.GAgents.*`（除 `agents/channels/**` 自身）的 .cs 文件

2. **`channel_mega_interface_guard.sh`**
   - `agents/Aevatar.GAgents.Channel.Abstractions/**/*.cs` 里不得声明同时包含 `StartReceivingAsync` 和 `SendAsync` 方法的 `interface`

3. **`channel_inbox_gagent_guard.sh`**
   - 代码库内不得新增类名 `InboxGAgent` / `ChannelInboxGAgent` / 类似 technical-function actor

4. **`channel_tombstone_proto_field_guard.sh`**
   - `UserAgentCatalogEntry` / `ChannelBotRegistrationEntry` / `DeviceRegistrationEntry` `.proto` 文件必须含 `bool is_deleted` 字段

### Layer 2: Roslyn analyzer + Architecture.Tests

6 个语义级规则，bash grep 做不到：

1. **CanonicalKey purity**: `ConversationReference.CanonicalKey` 生成方法被 `[CanonicalKeyGenerator]` attribute 标记；analyzer 禁止调用 `DateTime.Now` / `Guid.NewGuid` / `Random`
2. **ChatActivity.Id retry-stability**: `[ActivityIdGenerator]` attribute + analyzer 同上
3. **Outbound auth path**: `IChannelOutboundPort.SendAsync` caller 必须属 `agents/channels/**` 或经 `ITurnContext`（Architecture.Tests call graph check）
4. **Proactive actor-to-actor path**: `SkillRunnerGAgent` / workflow trigger / admin endpoint 不得直接调 `IChannelOutboundPort.ContinueConversationAsync`（Architecture.Tests）
5. **Payload redaction type flow**: Roslyn analyzer——`IBlobStore.WriteAsync(..., RawPayload, ...)` 类型流必须经过 `IPayloadRedactor.Redact` 类型转换
6. **Durable inbox provider dependency**: 实现 `IChannelDurableInbox` 契约的类必须依赖 `IAsyncStream<ChatActivity>`，不得用 `ConcurrentQueue` / 内存 `Channel` 作 durable store

### Acceptance

- [ ] 4 个 bash guard 脚本写好并加入 `architecture_guards.sh`
- [ ] 6 个 analyzer / Architecture.Tests 写好并加入 CI 门禁
- [ ] 实施 PR merge 前跑 guard suite 通过

## References

- RFC §14.1 CI Guards（分两层落地）

## Dependencies

- Depends on: #257（接口定义）
- Can run in parallel with adapter migration sub issues

eanzhao:
Closes #265.

Implements the two-layer CI guard coverage for the channel abstractions RFC (§14.1).

## Summary

- **Layer 1 (bash)** — three new shell guards wired into `tools/ci/architecture_guards.sh`; `channel_mega_interface_guard.sh` was already present.
- **Layer 2 (Roslyn-backed Architecture.Tests)** — six rule tests land in `Aevatar.Architecture.Tests`, backed by Roslyn syntax-tree walks over `src/**` + `agents/**` and two marker attributes in `Channel.Abstractions`.

## Layer 1 — bash guards

| Guard | Purpose |
| --- | --- |
| `channel_native_sdk_import_guard.sh` | Rejects `using Larksuite.Oapi` / `using Telegram.Bot` / `using SlackNet` / `using Discord.*` outside `agents/channels/**`. |
| `channel_mega_interface_guard.sh` (pre-existing) | Blocks interfaces that fuse `StartReceivingAsync` and `SendAsync` surfaces. |
| `channel_inbox_gagent_guard.sh` | Blocks technical-function `InboxGAgent` / `ChannelInboxGAgent` classes (CLAUDE.md "Actor 即业务实体"). |
| `channel_tombstone_proto_field_guard.sh` | Enforces `bool is_deleted` on `UserAgentCatalogEntry` / `ChannelBotRegistrationEntry` / `DeviceRegistrationEntry` under `agents/Aevatar.GAgents.Channel.*/**/*.proto`. |

## Layer 2 — Architecture.Tests (6 rules)

All rules live in `test/Aevatar.Architecture.Tests/Rules/ChannelArchitectureTests.cs` and share a `ChannelSourceIndex` helper that discovers the repo root and walks production source trees via `Microsoft.CodeAnalysis.CSharp`.

1. **CanonicalKey purity** — methods annotated with `[CanonicalKeyGenerator]` must not call `DateTime.Now` / `DateTime.UtcNow` / `Guid.NewGuid` / `Environment.TickCount(64)` / `Random.Shared` or `new Random(...)`. `ConversationReference.BuildCanonicalKey` now carries the attribute so the rule has a live anchor.
2. **ActivityId retry-stability** — same analyzer machinery against `[ActivityIdGenerator]`.
3. **Outbound auth path** — callers of `IChannelOutboundPort.SendAsync` must live under `agents/channels/**`, be the interface file itself, or be a `TurnContext*` surface.
4. **Proactive caller** — `SkillRunnerGAgent` / `WorkflowAgentGAgent` / `AdminBroadcast*` / `ChannelBroadcast*` types may not contain `.ContinueConversationAsync(` invocations; they must dispatch through a `ConversationGAgent` command envelope.
5. **Payload redaction** — source files that call `IBlobStore.WriteAsync` with raw payload bytes must also reference `IPayloadRedactor` in the same compilation unit.
6. **Durable inbox provider** — types that reference `IChannelDurableInbox` (other than the interface declaration itself) must depend on `IAsyncStream<ChatActivity>` and must not use `ConcurrentQueue<ChatActivity>` or `System.Threading.Channels.Channel<ChatActivity>` as durable storage.

Rules 5 and 6 remain vacuously satisfied until `IBlobStore` / `IChannelDurableInbox` land; the analyzer is wired so any future implementation trips the guard immediately.

## Validation

- `bash tools/ci/architecture_guards.sh` → all guards pass (including the three new channel guards).
- `dotnet test test/Aevatar.Architecture.Tests/...` → 105 pass, 1 pre-existing skip, 0 failures (6 new channel tests included).
- `dotnet test test/Aevatar.GAgents.Channel.Protocol.Tests/...` → 13/13 pass (regression check for the `[CanonicalKeyGenerator]` annotation on `BuildCanonicalKey`).
- Negative probe: dropping a `[CanonicalKeyGenerator]` method whose body calls `Guid.NewGuid()` produced the expected failure `forbidden 'Guid.NewGuid'` before cleanup restored green.

## Test plan

- [ ] CI runs `bash tools/ci/architecture_guards.sh` end-to-end and all new guards appear in the log as "ok".
- [ ] `dotnet test test/Aevatar.Architecture.Tests/...` passes in CI (6 new Channel tests included).
- [ ] Review PR against issue #265 acceptance criteria (4 bash guards wired + 6 analyzer/Architecture.Tests wired + guard suite green).


eanzhao:
Addressed both Codex review comments in f3409906:

**Comment 1 (P1 — `channel_native_sdk_import_guard.sh` line 22)** — regex broadened to also match `global`, `static`, and alias forms. New pattern:

```
^[[:space:]]*(global[[:space:]]+)?using([[:space:]]+static)?[[:space:]]+([A-Za-z_][A-Za-z0-9_]*[[:space:]]*=[[:space:]]*)?(Larksuite\.Oapi|Telegram\.Bot|SlackNet|Discord\.[A-Za-z0-9_]+)
```

Sanity-tested against plain, `global`, `using static`, aliased, and `global using static Tg = ...` samples. A negative probe with `using Tg = Telegram.Bot;` under a scanned directory now trips the guard.

**Comment 2 (P2 — `ChannelArchitectureTests.cs` proactive-caller rule)** — replaced the `Path.GetFileNameWithoutExtension` filter with a Roslyn `TypeDeclarationSyntax` walk. Partial classes split across arbitrarily-named files (e.g. `WorkflowAgentGAgent.Schedule.cs`) are now matched by declared type identifier, not filename. Negative probe: a `partial class WorkflowAgentGAgent` in a file called `_ProactiveProbe.cs` that calls `port.ContinueConversationAsync(...)` now produces:

```
…_ProactiveProbe.cs:WorkflowAgentGAgent:line 9: port.ContinueConversationAsync(null!, null!, null!, default)
```

Full `bash tools/ci/architecture_guards.sh` and `dotnet test test/Aevatar.Architecture.Tests/...` pass.

test/Aevatar.Architecture.Tests/Rules/ChannelArchitectureTests.cs
```diff
@@ -0,0 +1,509 @@
...
+                continue;
+            }
+
+            var normalized = ChannelSourceIndex.NormalizePath(sourceFile);
+            if (IsAllowedOutboundSendCaller(normalized))
+            {
+                continue;
+            }
+
+            var invocationReport = FindForbiddenInvocations(sourceFile, receiverTypeHint: "IChannelOutboundPort", methodName: "SendAsync");
```
eanzhao: 这条 rule 还是会漏报不少直接调用 `IChannelOutboundPort.SendAsync` 的写法。`FindForbiddenInvocations(...)` 只接受两类 receiver：表达式文本里直接带 `IChannelOutboundPort`，或者 receiver 名字恰好对应同文件里 typed field/property/parameter。像 `this._outbound.SendAsync(...)`、`var port = _outbound; port.SendAsync(...)`、wrapper property 再转发这些都不会命中，所以 item 3 还没真正做到“caller 必须属 agents/channels/** 或经 ITurnContext”。建议这里至少补上 `this.` member access / 局部别名，或者直接做 symbol 级绑定。

test/Aevatar.Architecture.Tests/Rules/ChannelArchitectureTests.cs
```diff
@@ -0,0 +1,509 @@
...
+    [Fact]
+    public void RawPayload_BlobWrites_Must_RouteThrough_PayloadRedactor()
+    {
+        // Rule: any IBlobStore.WriteAsync invocation that forwards a raw payload must receive the
+        // bytes back from IPayloadRedactor.Redact first. Until IBlobStore lands in the repo this
+        // guard is vacuously satisfied; once the type exists, any call site that mentions a raw
+        // payload without also mentioning the redactor fails the guard.
+
+        var violators = new List<string>();
+
```
eanzhao, line 229: 这条还没达到 issue #265 里“类型流必须经过 `IPayloadRedactor.Redact`”的强度。当前实现只要同文件里出现 `IPayloadRedactor` 或 `PayloadRedactor.Redact` 字样就直接放行，即使实际仍然是 `_blobStore.WriteAsync(rawPayload, ...)` 也不会报错。也就是说现在验证的是“同文件提到 redactor”，不是“传给 `WriteAsync` 的值来自 `Redact`”。

eanzhao (commented):
剩余还有 3 个点没有完全收敛到 issue #265 的承诺范围，我按具体代码位置留 inline comments 了。上次提到的 proactive caller 按文件名筛选问题这次已经修掉。

tools/ci/channel_tombstone_proto_field_guard.sh
```diff
@@ -0,0 +1,83 @@
...
+if not root.is_dir():
+    print("channel_tombstone_proto_field_guard: ok (no agents/ tree)")
+    sys.exit(0)
+
+violations: list[str] = []
+
+proto_files = [
+    path
+    for path in root.rglob("*.proto")
+    if any(part.startswith("Aevatar.GAgents.Channel.") for part in path.parts)
```
eanzhao: 这里的匹配只覆盖 `Aevatar.GAgents.Channel.*`，等价于把 `agents/Aevatar.GAgents.ChannelRuntime/**` 整个排除了。但 issue #265 的 item 1.4 没有这个豁免。当前仓库里 `agents/Aevatar.GAgents.ChannelRuntime/channel_runtime_messages.proto` 的 `ChannelBotRegistrationEntry` / `DeviceRegistrationEntry` 仍然没有 `bool is_deleted`，所以如果这个 PR 继续 `Closes #265`，acceptance 其实还没完全满足。

eanzhao:
Second review pass addressed in 1b0a6716.

**Comment on `ChannelArchitectureTests.cs:84` — OutboundSendAsync still missed common patterns.** The textual receiver heuristic (`this._outbound`, local alias, wrapper property) is replaced with a Roslyn walk:

- Per-type receiver set: fields, properties, and parameters whose declared type is `IChannelOutboundPort`.
- Local-alias closure via fixed-point over `VariableDeclaratorSyntax` + `AssignmentExpressionSyntax` — catches `var port = _outbound; var also = port; also.SendAsync(...)`.
- Repo-global member-name set: one pass indexes every identifier anywhere declared as `IChannelOutboundPort` so wrapper-property forwarding across files (`wrapper.Port.SendAsync(...)`) still fails the guard.

Negative probe hit all three forms:

```
ThisOutboundProbe:line 13: this._outbound.SendAsync(null!, null!, default)
LocalAliasProbe:line 29: also.SendAsync(null!, null!, default)
ConsumerOfWrapper:line 53: _wrapper.Port.SendAsync(null!, null!, default)
```

**Comment on `ChannelArchitectureTests.cs:173` — RawPayload rule only checked file-level redactor mention.** Replaced with per-method data-flow analysis:

1. Collect `IBlobStore`-typed receivers in the enclosing type.
2. Build a "redacted provenance" set per method body: seed from declarators / assignments whose RHS is `.Redact(` / `.RedactAsync(` / awaited variants, then expand transitively through variable aliasing.
3. For each `IBlobStore.WriteAsync(args)` invocation, only arguments that look like raw payload (`rawPayload`, `RawPayload`, identifier names containing `raw` / `Raw`) are examined; they must either be an inline `.Redact(` / `.RedactAsync(` call or their identifier must appear in the provenance set.

Negative probe fired on `_blobStore.WriteAsync(rawPayload, ...)` and stayed green for `var safeBytes = await _redactor.RedactAsync(...); _blobStore.WriteAsync(safeBytes, ...);`.

**Comment on `channel_tombstone_proto_field_guard.sh:40` — legacy proto tree silently excluded.** Dropped the `Aevatar.GAgents.Channel.*` filter and added `bool is_deleted = 11` to `ChannelBotRegistrationEntry` + `bool is_deleted = 7` to `DeviceRegistrationEntry` in `agents/Aevatar.GAgents.ChannelRuntime/channel_runtime_messages.proto`. The guard now scans every `.proto` under `agents/` and `ChannelRuntime` tests still pass (229/229).

Full validation:
- `bash tools/ci/architecture_guards.sh` — all guards pass.
- `dotnet test test/Aevatar.Architecture.Tests/...` — 105 pass / 1 skip (pre-existing) / 0 fail.
- `dotnet test test/Aevatar.GAgents.ChannelRuntime.Tests/...` — 229/229 pass.

test/Aevatar.Architecture.Tests/Rules/ChannelArchitectureTests.cs
```diff
@@ -0,0 +1,856 @@
...
+        if (ExtractLeafName(argument.Expression) is { } leaf && redactedProvenance.Contains(leaf))
+        {
+            return true;
+        }
+
+        return false;
+    }
+
+    public static string? ExtractLeafName(ExpressionSyntax expression)
```
eanzhao, line 709: `ExtractLeafName(...)` 还是没有处理 `InvocationExpressionSyntax`，所以 method-return receiver 这一类写法仍然是盲区。具体可复现：我在临时快照里加了一个 `internal` probe，内容是 `ResolvePort().SendAsync(...)`（`ResolvePort()` 返回 `IChannelOutboundPort`），然后只跑 `OutboundSendAsync_Callers_Must_StayInsideAdapters_Or_TurnContext` 这条测试，结果仍然 green。`RawPayload_BlobWrites_Must_RouteThrough_PayloadRedactor` 也受同一问题影响：`ResolveStore().WriteAsync(rawPayload, ct)` 这种 receiver 同样会因为这里返回 `null` 被放过。

eanzhao (commented):
这轮修复把上次的几个点收窄了不少，但我重新做了负例探针后，仍然有 2 个剩余问题。inline comments 里放了具体原因和可复现路径。

test/Aevatar.Architecture.Tests/Rules/ChannelArchitectureTests.cs
```diff
@@ -0,0 +1,856 @@
...
+            forbiddenTypeNames: ImmutableArray.Create("Random"));
+
+        Assert.True(
+            violations.Count == 0,
+            "CanonicalKey generator methods must remain deterministic for retry / replay / dedup. "
+            + "Forbidden calls detected:\n" + string.Join("\n", violations));
+    }
+
+    [Fact]
+    public void ActivityIdGenerator_MethodBodies_ShouldBe_RetryStable()
```
eanzhao, line 46: `ActivityIdGenerator` 这条 rule 现在仍然是 vacuous pass。仓库里我重新扫了一遍，生产代码里还没有任何方法真正标注 `[ActivityIdGenerator]`，所以这条测试目前只是在验证“如果以后有人用了这个 attribute，就会被扫描”，还没有像 `CanonicalKeyGenerator` 那样挂到一个真实锚点上。考虑到 PR 还在 `Closes #265`，这里建议至少补一个 live anchor，或者在 PR 描述里明确把 item 2 改成“先落 analyzer scaffolding，真实 generator 后续 issue 再接入”。

eanzhao:
Third review pass addressed in faafc870.

**Finding 1 (High) — method-return receivers still bypass the guards.** Correct: `ExtractLeafName` did not traverse `InvocationExpressionSyntax`, so `ResolvePort().SendAsync(...)` and `ResolveStore().WriteAsync(rawPayload, ct)` both slipped through. Fixed in two places:

- `ChannelReceiverTracker.ExtractLeafName` now recurses through `InvocationExpressionSyntax` (returning the callee identifier), `ConditionalAccessExpressionSyntax`, and `MemberBindingExpressionSyntax` in addition to the previously handled forms.
- Per-type and repo-global receiver sets now index methods / local functions whose declared return type matches the hint. `Task<IChannelOutboundPort>` and similar wrappers are unwrapped in `MethodReturnTypeMatches` so `async` resolvers also count.

Negative probe:

```
MethodReturnSendProbe:line 23: ResolvePort().SendAsync(null!, null!, default)
MethodReturnWriteProbe:line 40: argument 'rawPayload' passed to IBlobStore.WriteAsync does not trace back to IPayloadRedactor.Redact
```

Both rules fire on the method-return receiver forms after the change.

**Finding 2 (Medium) — vacuous `ActivityIdGenerator` rule.** Added `agents/Aevatar.GAgents.Channel.Abstractions/Transport/ChatActivity.Partial.cs` with a production helper:

```csharp
[ActivityIdGenerator]
public static string BuildActivityId(ChannelId channel, string deliveryKey)
{
    // deterministic channel:delivery_key concatenation
}
```

Parallel structure to `ConversationReference.BuildCanonicalKey`. Negative probe confirmed the analyzer fires when the body is mutated to `System.Guid.NewGuid()`:

```
ChatActivity.Partial.cs:BuildActivityId: forbidden 'Guid.NewGuid'
```

Restoring the helper returns the suite to green. The rule is now backed by a real production anchor, matching the `CanonicalKey` side.

Validation:
- `dotnet test test/Aevatar.Architecture.Tests/...` → 105 pass / 1 pre-existing skip / 0 fail.
- `dotnet test test/Aevatar.GAgents.Channel.Protocol.Tests/...` → 13/13 pass (ChatActivity partial doesn't disturb proto surface tests).
- `bash tools/ci/architecture_guards.sh` → all guards pass.

test/Aevatar.Architecture.Tests/Rules/ChannelArchitectureTests.cs
```diff
@@ -0,0 +1,914 @@
...
+    public static string? ExtractLeafName(ExpressionSyntax expression)
+    {
+        return expression switch
+        {
+            IdentifierNameSyntax id => id.Identifier.ValueText,
+            MemberAccessExpressionSyntax member => member.Name.Identifier.ValueText,
+            // A method-return receiver such as `ResolvePort().SendAsync(...)` or
+            // `this.ResolveStore().WriteAsync(...)` exposes the callee's name: recurse into the
+            // invocation's callee expression so the leaf resolves to the method identifier.
+            InvocationExpressionSyntax invocation => ExtractLeafName(invocation.Expression),
```
eanzhao, line 718: This closes the member-method/local-function receiver gap, but there is still a delegate-based bypass here. In a clean archive, both of these probes still passed:

```csharp
Func<IChannelOutboundPort> resolve = () => throw new NotSupportedException();
return resolve().SendAsync(to, content, ct);
```

```csharp
Func<IBlobStore> resolve = () => throw new NotSupportedException();
return resolve().WriteAsync(RawPayload, ct);
```

`ExtractLeafName(resolve())` resolves to `resolve`, but the tracked receiver/source sets only add typed fields/properties/parameters/methods and then alias those names; they never add local delegate variables whose return type is `IChannelOutboundPort` / `IBlobStore`. So rule 3 and rule 5 are still not fully closed yet.

eanzhao (commented):
Left one inline comment for the remaining delegate-receiver bypass in the channel Architecture.Tests guards.

eanzhao:
Fourth review pass addressed in ab9ab6fe.

**Reviewer's probe** (both were still passing): `Func<IChannelOutboundPort> resolve = () => throw new NotSupportedException(); return resolve().SendAsync(to, content, ct);` and the matching `Func<IBlobStore> resolve; resolve().WriteAsync(RawPayload, ct);`. You're right — `ExtractLeafName(resolve())` resolves to `"resolve"`, but the tracked set only indexed type-level members plus alias-by-initializer expansion. Local delegate variables (and even plain `IFoo local = source;` declarations) were never seeded into the set.

**Fix.** `ChannelReceiverTracker.ExpandLocalAliases` now takes the type hint and pre-seeds the set with:

- every `LocalDeclarationStatementSyntax` whose declared type text matches the hint as a whole-word segment — covers `Func<IFoo>`, `Lazy<IFoo>`, `Task<IFoo>`, `IEnumerable<IFoo>`, and plain `IFoo local = ...;`
- every `LocalFunctionStatementSyntax` whose return type matches the hint

Both the SendAsync rule and the RawPayload rule now call `ExpandLocalAliases(body, seed, typeHint)` so the per-method receiver set closes over these local declarations before the receiver-name check fires. `TypeTextMatches` already handled generic-argument whole-word matching (`Func<IChannelOutboundPort>` → matches `IChannelOutboundPort`), so the change is purely in what nodes `ExpandLocalAliases` inspects.

Negative probe against the exact forms in your comment:

```
FuncOutboundProbe:line 17:     resolve().SendAsync(to, content, ct)
DirectLocalOutboundProbe:line 39: local.SendAsync(to, content, ct)
FuncBlobStoreProbe:line 28:    argument 'rawPayload' passed to IBlobStore.WriteAsync does not trace back to IPayloadRedactor.Redact
```

All three flagged; cleanup returns the suite to green.

Validation:
- `dotnet test test/Aevatar.Architecture.Tests/...` → 105 pass / 1 pre-existing skip / 0 fail.
- `bash tools/ci/architecture_guards.sh` → all guards pass.

Known remaining limitation (for transparency): lambdas bound to `var` (`var resolve = () => ...`) get no explicit type annotation, so the heuristic can't tell their return type from syntax alone. Explicit `Func<IFoo>` typing is what the guard requires — matches the reviewer's example and is the common production shape.

eanzhao:
Fifth review pass addressed in fb88504f.

**Root cause.** `ExpandLocalAliases` short-circuited on `TypeTextMatches(local.Declaration.Type, hint)` — for `var resolve = new Func<IChannelOutboundPort>(...)` the declaration-type text is literally `var`, so nothing inspected the initializer, so `resolve` never entered the receiver set.

**Fix.** `ExpandLocalAliases` now falls back to the initializer expression whenever the declaration type is `var`, dispatching through a new `InitializerExpressionMatchesTypeHint` helper that handles:

- `new Func<IFoo>(...)` / `new Lazy<IFoo>(...)` → `ObjectCreationExpressionSyntax.Type`
- `(IFoo)expr` → `CastExpressionSyntax.Type`
- `default(Func<IFoo>)` → `DefaultExpressionSyntax.Type`
- parenthesized + awaited wrappers
- `cond ? a : b` and `a ?? b` — recurse into each branch

The inline comment on the lambda-to-`var` limit remains, but the two forms you flagged are now covered.

Negative probes, reviewer's exact examples:

```
VarFuncOutboundProbe:line 17:  resolve().SendAsync(to, content, ct)
VarFuncBlobStoreProbe:line 28: argument 'rawPayload' passed to IBlobStore.WriteAsync does not trace back to IPayloadRedactor.Redact
VarCastBlobStoreProbe:line 39: argument 'rawPayload' passed to IBlobStore.WriteAsync does not trace back to IPayloadRedactor.Redact
```

All three flagged. Cleanup returns the suite to green.

**Validation**
- `dotnet test test/Aevatar.Architecture.Tests/...` → 105 pass / 1 pre-existing skip / 0 fail.
- `bash tools/ci/architecture_guards.sh` → all guards pass.

**Known remaining limitation** (documented inline in `ExpandLocalAliases`): `var resolve = () => ...` — lambda expression bound to `var` with no explicit type cast — still carries no syntactic type hint, so the heuristic can't infer its return type without a semantic model. Any explicit typing (`Func<IFoo> resolve = ...`, `var resolve = (Func<IFoo>)(() => ...)`, `var resolve = new Func<IFoo>(...)`) is now enforced.

eanzhao:
Sixth review pass addressed in 499b65e3.

**Root cause.** `var resolve = () => _outbound;` declares `resolve` with inferred natural type. The previous pass only inspected the initializer's declared *type*; it never walked into the lambda body. So `resolve` never joined the receiver set, and `resolve().SendAsync(...)` stayed green.

**Fix.** The `ExpandLocalAliases` fixed-point now has a second arm that recognizes `LambdaExpressionSyntax` and `AnonymousMethodExpressionSyntax` initializers / RHS assignments. If any return expression's leaf name is already in the receiver set, the declarator / LHS identifier is added in the next iteration — so once `_outbound` / `_store` are in the seed (via their field declarations), the lambda-returning local enters on the next pass.

Return collection supports:

- `() => known` expression bodies
- `() => { return known; }` block bodies (descent explicitly stops at nested lambdas / anonymous methods / local functions to avoid mis-attributing inner returns)
- `delegate { return known; }` anonymous methods
- parenthesized / cast wrappers around the lambda

Negative probes against the reviewer's exact forms plus natural variants:

```
VarLambdaOutboundProbe:line 22:    resolve().SendAsync(to, content, ct)
VarLambdaBlobStoreProbe:line 38:   argument 'rawPayload' passed to IBlobStore.WriteAsync does not trace back to IPayloadRedactor.Redact
VarLambdaBlockProbe:line 54:       resolve().SendAsync(to, content, ct)
VarAnonymousMethodProbe:line 70:   resolve().SendAsync(to, content, ct)
```

All four flagged. Cleanup returns the suite to green.

**Validation**
- `dotnet test test/Aevatar.Architecture.Tests/...` → 105 pass / 1 pre-existing skip / 0 fail.
- `bash tools/ci/architecture_guards.sh` → all guards pass.

**Remaining limitation** (narrower than before): lambdas that return receivers **not** yet in the set — e.g. a captured local that was itself obtained through a semantic type-flow the tracker can't follow — still bypass. Every syntactic pattern the reviewer surfaced across the review series is now covered; genuinely semantic cross-file flow requires a full `Compilation` + `SemanticModel` which is out of scope for this guard.

eanzhao:
Seventh review pass addressed in a89fddb2.

**Root cause.** `AnyLambdaReturnInSet` called `ExtractLeafName` on the whole lambda-return expression. `ExtractLeafName` doesn't split `a ?? b` / `cond ? a : b` / `switch { ... }`, so the lambda body resolved to `null` and the lambda-typed local never entered the receiver set. Your two probes and their variants all bypassed.

**Fix.** New helpers `ReturnExpressionMatches` + `EnumerateReceiverBranches`:

- `a ?? b` → enumerate both operands
- `cond ? a : b` → enumerate both branches (cond itself is ignored)
- `switch { arm => expr, ... }` → enumerate every arm expression
- `throw ...` → yields nothing (not a receiver)
- parenthesized / cast / await wrappers → unwrap, recurse

`AnyLambdaReturnInSet` (lambda-body branch) now routes through `ReturnExpressionMatches`. For symmetry the alias fixed-point also swaps `ExtractLeafName` → `ReturnExpressionMatches` so `var alias = _outbound ?? throw ...; alias.SendAsync(...)` is caught without needing a lambda wrapper.

Negative probes:

```
CoalesceThrowOutboundProbe:line 22:     resolve().SendAsync(to, content, ct)
ConditionalThrowBlobStoreProbe:line 39: argument 'rawPayload' passed to IBlobStore.WriteAsync does not trace back to IPayloadRedactor.Redact
CoalesceAliasOutboundProbe:line 55:     alias.SendAsync(to, content, ct)
SwitchExpressionOutboundProbe:line 71:  resolve().SendAsync(to, content, ct)
```

All four flagged. Cleanup returns the suite to green.

**Validation**
- `dotnet test test/Aevatar.Architecture.Tests/...` → 105 pass / 1 pre-existing skip / 0 fail.
- `bash tools/ci/architecture_guards.sh` → all guards pass.

eanzhao:
Eighth review pass addressed in 38e1d08e.

**Root cause.** Rule 5 (RawPayload) had two filters that rule 3 (SendAsync) did not:

1. `if (!text.Contains("IBlobStore")) continue;` — the consumer file never names `IBlobStore` in its own text (the interface and the wrapper property live in `BlobStoreWrapperTypes.cs`), so the whole file was skipped.
2. `if (callableBlobStores.Count == 0) continue;` — even if the file were scanned, the consumer type has no `IBlobStore`-typed member directly, so the per-method receiver set was empty and the type skipped.

Rule 3 already solves the same class of problem via (a) a `.SendAsync` contains check and (b) `GlobalOutboundMemberNames`. Giving rule 5 the same shape:

- **`ChannelSourceIndex.GlobalBlobStoreMemberNames`** — lazy index parallel to `GlobalOutboundMemberNames`, built via the existing `CollectGlobalMemberNamesByType("IBlobStore")` helper. Indexes every field / property / parameter / method-return typed `IBlobStore` across the repo — so wrapper properties declared in one file are reachable from another file's receiver check.
- **Short-circuit swap** — `!text.Contains("IBlobStore")` → `!text.Contains(".WriteAsync")`. Consumer files are now scanned as long as they invoke a `WriteAsync` member.
- **Per-type skip removed** — `callableBlobStores.Count == 0` no longer short-circuits. Receiver-name lookup unions the per-method set with the repo-global index before flagging.

Negative probe (reviewer's exact layout — two files, consumer never mentions `IBlobStore`):

```
_BlobStoreWrapperTypes.cs:    public interface IBlobStore { ... }
                              public sealed class BlobStoreHolder { public IBlobStore Store { get; } ... }

_BlobStoreWrapperConsumer.cs: public System.Threading.Tasks.Task RunAsync(byte[] rawPayload, ...)
                                  => _holder.Store.WriteAsync(rawPayload, "k", ct);
```

Test now reports:
```
_BlobStoreWrapperConsumer.cs:BlobStoreWrapperConsumer:line 13:
argument 'rawPayload' passed to IBlobStore.WriteAsync
does not trace back to IPayloadRedactor.Redact
```

**Validation**
- `dotnet test test/Aevatar.Architecture.Tests/...` → 105 pass / 1 pre-existing skip / 0 fail.
- `bash tools/ci/architecture_guards.sh` → all guards pass.

eanzhao:
Ninth review pass addressed in 21fbbf59.

**Root cause.** Null-conditional `x?.SendAsync(...)`, `x?.WriteAsync(...)`, `x?.ContinueConversationAsync(...)` all come through `ConditionalAccessExpressionSyntax` with the invocation's callee expressed as `MemberBindingExpressionSyntax` (the `.SendAsync` part detaches from the `?.` operator). All three rule bodies pattern-matched `invocation.Expression is not MemberAccessExpressionSyntax` and silently skipped those forms.

**Fix.** New `ChannelReceiverTracker.GetMemberInvocationParts(invocation)` normalises both shapes into `(methodName, receiverExpression)`:

- `x.SendAsync(...)` → `(SendAsync, x)` via `MemberAccessExpressionSyntax`
- `x?.SendAsync(...)` → `(SendAsync, x)` via `MemberBindingExpressionSyntax` + `invocation.FirstAncestorOrSelf<ConditionalAccessExpressionSyntax>().Expression`
- Chained `wrapper?.Port?.SendAsync(...)` → `(SendAsync, MemberBinding(.Port))` → receiver leaf `Port`

Rule 3, rule 4 (proactive caller), and rule 5 all route receiver extraction through the helper. The proactive-caller file-level short-circuit was also widened from `.ContinueConversationAsync(` to `ContinueConversationAsync(` so the `?.` form isn't filtered at the text-contains stage.

Negative probes — reviewer's exact `SkillRunnerGAgentConditionalProbe` plus natural variants:

```
NullCondOutboundProbe:line 20:           _outbound?.SendAsync(to, content, ct)
ChainedNullCondConsumer:line 56:         _wrapper?.Port?.SendAsync(to, content, ct)
NullCondBlobStoreProbe:line 33:          _store?.WriteAsync(rawPayload, ...) — does not trace back to IPayloadRedactor.Redact
SkillRunnerGAgentConditionalProbe:line 69: _outbound?.ContinueConversationAsync(...)
```

All four flagged. Cleanup returns the suite to green.

**Validation**
- `dotnet test test/Aevatar.Architecture.Tests/...` → 105 pass / 1 pre-existing skip / 0 fail.
- `bash tools/ci/architecture_guards.sh` → all guards pass.

eanzhao:
Merged `origin/dev` (commit 577e7355).

**Conflict** was in `agents/Aevatar.GAgents.ChannelRuntime/channel_runtime_messages.proto` — both sides added a tombstone marker to `ChannelBotRegistrationEntry` / `DeviceRegistrationEntry`:

- **My side** (this PR): `bool is_deleted` — minimal marker to satisfy issue #265 item 1.4.
- **Dev side**: `bool tombstoned` + `int64 tombstone_state_version` + matching `*CompactTombstonesCommand` / `*TombstonesCompactedEvent` messages and tombstoned-aware unregister events.

Dev's implementation is strictly richer — it keeps the registration in state for projector watermark coordination (Channel RFC §7.1.1), has accompanying compaction events, and real tests. I **took dev's version** and broadened `tools/ci/channel_tombstone_proto_field_guard.sh` to accept either `bool is_deleted` **or** `bool tombstoned` — the semantic (a tombstone marker field) is what the guard really cares about, and two conventions coexist across the tree (the new dotted `Aevatar.GAgents.Channel.*` path still uses `is_deleted`, the legacy `Aevatar.GAgents.ChannelRuntime` path uses `tombstoned`).

Updated guard scope documented inline; the regex is now `\bbool\s+(is_deleted|tombstoned)\s*=\s*\d+\s*;`.

**Validation post-merge**
- `bash tools/ci/channel_tombstone_proto_field_guard.sh` → ok
- `bash tools/ci/architecture_guards.sh` → all guards pass
- `dotnet test test/Aevatar.Architecture.Tests/...` → 105 pass / 1 pre-existing skip / 0 fail
- `dotnet test test/Aevatar.GAgents.ChannelRuntime.Tests/...` → 255 pass (up from 229 pre-merge as dev added suites)

Issue #264, opened 2026-04-20
[Channel RFC] Conformance + Fault Injection test suite (Aevatar.GAgents.Channel.Testing)


eanzhao:
**Parent RFC**: #254

## Scope

Conformance Suite + Fault Injection Suite —— 本 RFC 架构承诺的可执行契约。

### Deliverables

**Test infrastructure package**:
- `test/Aevatar.GAgents.Channel.Testing`（抽象基类库，不是 production code——放 test/ 不放 agents/）

**Conformance base class** (`ChannelAdapterConformanceTests<TAdapter>`):

```csharp
public abstract class ChannelAdapterConformanceTests<TAdapter>
    where TAdapter : IChannelTransport, IChannelOutboundPort
{
    protected abstract TAdapter CreateAdapter();
    protected abstract WebhookFixture? WebhookFixture { get; }
    protected abstract GatewayFixture? GatewayFixture { get; }
    
    // 14 conformance tests — see RFC §8.1
}
```

**Composer base class** (`MessageComposerUnitTests<TComposer>`):
- 6 composer unit tests — see RFC §8.2

**Fault Injection Suite** (`ChannelAdapterFaultTests<TAdapter>`) — RFC §8.3：
- `CommitCrashConsumeGap_BotStillExecutesActivityAfterRestart`
- `Inbound_SamePayloadRetried_ProducesSameActivityId`
- `Inbound_DuplicateFromPlatformRetry_DedupedExactlyOnce`
- `Inbound_DuplicateFromConcurrentSilos_DedupedExactlyOnce`
- `Gateway_ResumeInvalidated_FallsBackToIdentify`
- `Gateway_PreStopMissing_DetectsEventGap`
- `InboundStream_Saturation_TriggersBackpressureSignal`
- `RawPayload_DoesNotLeakCredentials`
- `EmitResult_ErrorMessage_DoesNotContainVendorRawBody`
- `Streaming_DisposedWithoutComplete_MarksMessageInterrupted`
- `Streaming_IntentDegradesMidway_ReachesTerminalState`
- `Projector_TombstonedEntry_InvokesDispatcherDeleteAsync`
- `Projector_LaggedBehindHousekeeping_DoesNotMissTombstone`
- `Shard_DualSupervisor_OnlyLeaderWritesSessionState`
- `ShardLeaderGrain_StaleWriteRejectedByLeaseEpoch`
- `ShardLeaderGrain_LastSeqMonotonicallyIncreases`
[...]

Issue #257, opened 2026-04-20
[Channel RFC] Core abstractions package (Aevatar.GAgents.Channel.Abstractions)


eanzhao:
**Parent RFC**: #254

## Scope

`Aevatar.GAgents.Channel.Abstractions` package —— 核心接口层。

### Deliverables

**Interfaces** (`Aevatar.GAgents.Channel.Abstractions/`):
- `IChannelTransport` (Runtime side: lifecycle + inbound stream + capabilities)
- `IChannelOutboundPort` (Dispatch side: send/update/delete/continue)
- `IMessageComposer<TNativePayload>` + `IMessageComposer` 通用
- `IBot` + `ITurnContext` (业务 bot logic 接口)
- `IChannelMiddleware` (窄范围横切：logging / tracing / conversation resolver)
- `ICredentialProvider` (see #255)
- `IPayloadRedactor` (含 `HealthCheckAsync`)
- `PerEntryDocumentProjector<TEntry, TDocument>` 抽象基类（base 类抽取见 #259）
- `ISchedulable` marker interface（见 #259）

**Hard constraint**: single-class two-interfaces
- 禁止 "独立两个类实现"——`IChannelTransport` 和 `IChannelOutboundPort` 必须 single class 同时实现（init/auth/capabilities 共享 state）
- CI guard: `channel_mega_interface_guard.sh` 禁止 mega-interface 回归
- Conformance Suite 泛型约束 `where TAdapter : IChannelTransport, IChannelOutboundPort`

**Types** (from proto-generated in #256):
- C# record partial classes 在 proto-generated types 上扩展必要的 helper 方法（deterministic canonical key builders etc.）

**Interface documentation**:
- §5.8 Interface Contracts 全部内联为 XML doc comments（lifecycle / idempotency / ordering）

### Acceptance

- [ ] `Aevatar.GAgents.Channel.Abstractions.csproj` 独立 build 通过
- [ ] 接口 surface 过 Abigail / Shining / Potter / Louis 架构 review（对齐 Commitment 2 盲区覆盖）
- [ ] XML doc 完整，CI 强制 no public API without doc

## References

- RFC §5 核心接口设计（§5.1-§5.8 全章节）
- RFC §5.8 Interface Contracts

## Dependencies

- Depends on: #256

PR #276, opened 2026-04-21
Add channel adapter conformance + fault test base classes (#264)

eanzhao:
## Summary

Introduces `Aevatar.GAgents.Channel.Testing` — the abstract test base class library that channel adapters (Lark / Telegram / future Slack / Discord) derive from to satisfy the RFC §8 Conformance + Fault Injection suites. This PR now targets `dev` directly after the core abstractions landed there, and it closes the shared testing-infrastructure part of [#264](https://github.com/aevatarAI/aevatar/issues/264).

- `ChannelAdapterConformanceTests<TAdapter>` — conformance coverage for inbound / outbound / lifecycle / capability parity, multi-tenant routing, participant canonicalization, typing, reactions, and mention normalization.
- `MessageComposerUnitTests<TComposer>` — intent → native translation tests with adapter-supplied shape assertions so the base class remains channel-agnostic.
- `ChannelAdapterFaultTests<TAdapter>` — fault coverage routed either to adapter fixtures or a `RuntimeFaultHarness`, so adapter-only suites exercise adapter-scoped failures while runtime-coupled cases remain explicitly opt-in.
- Reference stub adapter + composer in `Aevatar.GAgents.Channel.Protocol.Tests` derive from each base class to keep the library honest; the current stub suite passes `77/77`.

## Review Follow-ups

Follow-up commits on this PR tightened several tests after review feedback:

- weak / placeholder fault tests were replaced with probe- or harness-based assertions
- capability claims for reactions / mentions now require the corresponding test hooks instead of silently self-skipping
- gateway-only inbound paths no longer skip duplicate / mention coverage just because `WebhookFixture` is absent
- webhook-specific multi-binding routing checks were narrowed to the webhook fixture path to avoid false coverage and nullability holes

## Design Notes

- The library lives under `test/` because it is test infrastructure, not production code.
- The same-class two-interfaces constraint (`where TAdapter : class, IChannelTransport, IChannelOutboundPort`) is reflected across the suites so the channel mega-interface guard keeps conformance and fault coverage aligned.
- `WebhookFixture` / `GatewayFixture` stay channel-agnostic; adapter suites provide synthetic payloads and bindings without coupling the base library to one platform.
- `RuntimeFaultHarness` is the explicit extension seam for runtime-coupled fault tests such as crash / restart, shard leadership, redactor fail-closed, and retryability semantics.

## Scope / Follow-up

- Parent RFC: #254
- Core abstractions from #257 are already in `dev`, so this PR now rebases directly on `dev`
- Adapter-specific derivations remain in #261 / #262
- Making adapter suites and CI status a `pull_request` hard gate is still follow-up work outside this PR

## Validation

- [x] `dotnet build test/Aevatar.GAgents.Channel.Testing/Aevatar.GAgents.Channel.Testing.csproj`
- [x] `dotnet test test/Aevatar.GAgents.Channel.Protocol.Tests/Aevatar.GAgents.Channel.Protocol.Tests.csproj --nologo` — `77/77` passed after the latest review fixes
- [ ] Lark + Telegram adapters deriving from this suite land in #261 / #262


test/Aevatar.GAgents.Channel.Testing/Fault/ChannelAdapterFaultTests.cs
```diff
@@ -0,0 +1,482 @@
...
+        (resumed.IsResumed == false || GatewayFixture.AuthoritativeSequenceNumber >= 0).ShouldBeTrue();
+    }
+
+    [Fact]
+    public async Task InboundStream_Saturation_TriggersBackpressureSignal()
+    {
+        await using var lifetime = await StartAdapterAsync();
+        var reader = lifetime.Adapter.InboundStream;
+        reader.ShouldNotBeNull();
+        await Task.Yield();
```
eanzhao: This test is still effectively a no-op: it only asserts that `InboundStream` is non-null and then yields. That will pass even if saturation never produces any observable backpressure signal or incorrectly drops work. Consider driving the stream to capacity through a fixture/harness and asserting the actual saturation contract.

test/Aevatar.GAgents.Channel.Testing/Fault/ChannelAdapterFaultTests.cs
```diff
@@ -0,0 +1,482 @@
...
+        failedClosed.ShouldBeTrue();
+    }
+
+    [Fact]
+    public async Task QuarantineOverflow_DropsOldestAndAlerts()
+    {
+        if (RuntimeHarness is null || !RuntimeHarness.SupportsRedactorPipeline)
+            return;
+
+        await Task.Yield();
```
eanzhao: This is also a no-op right now. `Task.Yield()` does not verify either half of the contract in the test name (`drops oldest` and `alerts`), so quarantine overflow can regress silently. It probably needs a dedicated harness method that reports both the eviction result and whether an alert was emitted.

test/Aevatar.GAgents.Channel.Testing/Conformance/ChannelAdapterConformanceTests.cs
```diff
@@ -0,0 +1,575 @@
...
+    {
+        if (WebhookFixture is null)
+            return;
+
+        var secondary = CreateSecondaryBinding();
+        if (secondary is null)
+            return;
+
+        await using var lifetime = await StartAdapterAsync();
+        var activity = await DispatchAsync(InboundActivitySeed.DirectMessage("multi-tenant"));
```
eanzhao: This does not actually exercise multi-tenant routing. The test never dispatches anything through `secondary`, so it only proves that the primary bot id differs from the secondary descriptor. To validate routing, the fixture needs to inject one inbound event for each binding and assert that each activity is attributed to the correct bot instance.

eanzhao (commented):
Left three inline comments for the remaining weak spots in the suite.

Separately, I don't think this fully closes #264 yet. The issue acceptance calls for Lark + Telegram adapters to pass the conformance/composer/fault suites, and for suite pass to be a hard merge gate. This PR still defers the adapter-specific suites, and the current CI workflow does not run the `dotnet test` split matrix on `pull_request`, so the suite is not yet enforced as a PR-time hard gate.

test/Aevatar.GAgents.Channel.Testing/Conformance/ChannelAdapterConformanceTests.cs
```diff
@@ -0,0 +1,575 @@
...
+            CancellationToken.None);
+        emit.Capability.ShouldBeOneOf(ComposeCapability.Exact, ComposeCapability.Degraded, ComposeCapability.Unsupported);
+        emit.Success.ShouldBeTrue();
+    }
+
+    [Fact]
+    public async Task Capabilities_ImplementedInterfaces_AreConsistent()
+    {
+        await using var lifetime = await StartAdapterAsync();
+        var caps = CapabilitiesOf(lifetime.Adapter);
```
eanzhao, line 427: This still doesn't validate "implemented interfaces" consistency. The method only inspects static capability values and `Transport`, while the `AdapterImplements<TCapability>` helper above is never used. As written, an adapter can drift between capability flags and its optional typing/reaction/mention interfaces and still pass. If parity is the goal, this needs to assert against the actual optional surfaces rather than only checking numeric invariants.

test/Aevatar.GAgents.Channel.Testing/Conformance/ChannelAdapterConformanceTests.cs
```diff
@@ -0,0 +1,575 @@
...
+
+    [Fact]
+    public async Task Reaction_SendAndRemove_Succeeds()
+    {
+        await using var lifetime = await StartAdapterAsync();
+        if (!CapabilitiesOf(lifetime.Adapter).SupportsReactions)
+            return;
+
+        var reference = BuildDirectMessageReference(lifetime.Adapter);
+        var emit = await lifetime.Adapter.SendAsync(
```
eanzhao: This test name promises `SendAndRemove` reaction coverage, but the body only calls `SendAsync` on a normal text message. An adapter could have no reaction implementation at all and still pass here. Either drive the actual reaction surface and assert both add/remove behavior, or rename/defer the test so the suite does not claim coverage it doesn't have.

test/Aevatar.GAgents.Channel.Testing/Conformance/ChannelAdapterConformanceTests.cs
```diff
@@ -0,0 +1,575 @@
...
+            CanonicalId = lifetime.Binding.Bot.Bot.Value,
+            DisplayName = "Bot",
+        };
+        var seed = new InboundActivitySeed(
+            ActivityType.Message,
+            ConversationScope.DirectMessage,
+            ConversationKey: "dm-mention",
+            SenderCanonicalId: "user-1",
+            SenderDisplayName: "Test User",
+            Text: "hello",
```
eanzhao: The input text here does not contain any bot mention syntax, so there is nothing for the adapter to strip. That means the test can pass even if bot mentions are never removed from `Content.Text`. To validate the contract in the name, seed a platform-like message that actually mentions the bot and then assert the mention is absent from normalized text while still present in `activity.Mentions`.

eanzhao (commented):
Reviewed the previous fixes on the current head. The gateway-only dispatch fix and the new streaming probe hooks look correct to me.

I did leave three more inline comments for tests that still read stronger than what they actually verify, because they can produce false confidence in the conformance matrix.

test/Aevatar.GAgents.Channel.Testing/Conformance/ChannelAdapterConformanceTests.cs
```diff
@@ -0,0 +1,628 @@
...
+    /// implement the supplied optional interface, and vice versa. Default returns an empty set — adapters with optional
+    /// typing / reaction / mention adapter interfaces override this hook to wire the parity into the suite.
+    /// </remarks>
+    protected virtual IEnumerable<CapabilityInterfaceParity> GetCapabilityInterfaceParities(TAdapter adapter) =>
+        Array.Empty<CapabilityInterfaceParity>();
+
+    /// <summary>
+    /// Returns the adapter-supplied reaction probe, or <see langword="null"/> when the adapter has no reaction surface.
+    /// </summary>
+    protected virtual ReactionProbe? Reactions => null;
```
eanzhao, line 110: Because this defaults to `null`, an adapter can set `Capabilities.SupportsReactions = true` and still pass the suite without ever exercising reaction add/remove behavior: `Reaction_SendAndRemove_Succeeds` just returns when `Reactions` is null. If the goal is conformance, consider making the probe mandatory whenever `SupportsReactions` is true instead of silently self-skipping.

test/Aevatar.GAgents.Channel.Testing/Conformance/ChannelAdapterConformanceTests.cs
```diff
@@ -0,0 +1,628 @@
...
+    /// <summary>
+    /// Builds a synthetic inbound activity seed whose body mentions the bot associated with the supplied binding using
+    /// the adapter's native mention syntax.
+    /// </summary>
+    /// <remarks>
+    /// Default returns <see langword="null"/> so the mention test self-skips. Adapters with platform-specific mention
+    /// syntax override to emit a seed whose <see cref="InboundActivitySeed.Text"/> contains the raw mention token and
+    /// whose <see cref="InboundActivitySeed.Mentions"/> carries the normalized participant.
+    /// </remarks>
+    protected virtual InboundActivitySeed? BuildBotMentionSeed(ChannelTransportBinding binding) => null;
```
eanzhao, line 132: These mention seeds default to `null`, and the mention tests simply return when they are missing. That means an adapter can advertise `SupportsMention = true` yet still pass without any bot-mention stripping or participant-mention preservation coverage. Similar to reactions, this probably needs to fail fast when mention support is claimed but the adapter did not provide the required test hooks.

eanzhao (commented):
Reviewed the latest fix commit `7a3f3675`. The new saturation/quarantine assertions and the earlier multi-tenant / mention / reaction fixes move things in the right direction.

I left two more inline comments for capability paths that can still self-skip even when the adapter advertises support, which would let the conformance suite miss real regressions.

test/Aevatar.GAgents.Channel.Testing/Conformance/ChannelAdapterConformanceTests.cs
```diff
@@ -0,0 +1,644 @@
...
+        added.ShouldBeTrue("Adding a reaction on a channel with SupportsReactions=true must succeed.");
+
+        var removed = await Reactions.RemoveAsync(sent.SentActivityId, ":thumbsup:", CancellationToken.None);
+        removed.ShouldBeTrue("Removing a previously added reaction must succeed.");
+    }
+
+    [Fact]
+    public async Task Mention_BotMention_StrippedFromTextButPresentInMentions()
+    {
+        if (WebhookFixture is null)
```
eanzhao: This still skips mention coverage for gateway-only adapters. `DispatchAsync(...)` already supports `GatewayFixture`, but this test returns immediately whenever `WebhookFixture` is null, so an adapter can advertise `SupportsMention = true` on a gateway transport and still never exercise bot-mention stripping here. The same issue exists in the other-participant mention test below. If mention normalization is meant to be transport-agnostic inbound behavior, this gate should probably be `WebhookFixture is null && GatewayFixture is null` instead.

eanzhao (commented):
Reviewed the latest fix commit `cea3ddee`. The reaction/mention hook tightening is moving in the right direction.

I left one more inline comment for a remaining gap: gateway-only adapters can still bypass mention coverage even when they advertise `SupportsMention=true`.

test/Aevatar.GAgents.Channel.Testing/Conformance/ChannelAdapterConformanceTests.cs
```diff
@@ -0,0 +1,645 @@
...
+                parity.CapabilityFlag,
+                $"Capability flag '{parity.CapabilityName}' must match the presence of its optional interface. "
+                    + $"Flag={parity.CapabilityFlag}, interface implemented={parity.InterfaceImplemented}.");
+        }
+    }
+
+    [Fact]
+    public async Task ParticipantRef_CanonicalId_StableAcrossDisplayNameChanges()
+    {
+        if (WebhookFixture is null)
```
eanzhao: This canonicalization test still self-skips for gateway-only adapters because it gates on `WebhookFixture` rather than on inbound support in general. `DispatchAsync(...)` already supports `GatewayFixture`, so a gateway transport can currently bypass the `CanonicalId` stability check entirely. If this is meant to be transport-agnostic inbound conformance, the guard should be `WebhookFixture is null && GatewayFixture is null`.

test/Aevatar.GAgents.Channel.Testing/Conformance/ChannelAdapterConformanceTests.cs
```diff
@@ -0,0 +1,645 @@
...
+            ConversationKey: "dm-participant",
+            SenderCanonicalId: "user-42",
+            SenderDisplayName: "Beta",
+            Text: "second"));
+
+        first.From.CanonicalId.ShouldBe(second.From.CanonicalId);
+    }
+
+    [Fact]
+    public async Task Typing_KeepaliveWithinInterval()
```
eanzhao, line 474: These typing tests are still mostly capability-shape checks rather than behavior checks. Across `Typing_KeepaliveWithinInterval`, `Typing_AutoStopsAfterTtl`, `Typing_FailureCircuitBreakerAfterConsecutiveFailures`, and `Typing_StartStop_Idempotent`, the suite never drives a typing surface or observes keepalive / TTL / breaker / idempotent start-stop behavior. An adapter can advertise `SupportsTyping=true` and still pass with a broken typing implementation. This probably needs a dedicated typing probe, similar to reactions / streaming, or the test names should be narrowed.

test/Aevatar.GAgents.Channel.Testing/Fault/ChannelAdapterFaultTests.cs
```diff
@@ -0,0 +1,492 @@
...
+        snapshot.ResumeTokenAtStart.ShouldNotBeNullOrWhiteSpace();
+
+        await GatewayFixture.InvalidateResumeTokenAsync();
+        var resumed = await GatewayFixture.StartAsync();
+
+        resumed.IsResumed.ShouldBeFalse();
+    }
+
+    [Fact]
+    public async Task Gateway_PreStopMissing_DetectsEventGap()
```
eanzhao, line 142: This assertion is too weak for the contract in the test name. `(resumed.IsResumed == false || GatewayFixture.AuthoritativeSequenceNumber >= 0)` is effectively always true for any fixture with a non-negative sequence number, so the test does not really verify that a missing pre-stop caused the adapter to detect an event gap. It should assert an observable gap-handling outcome instead of a tautological fallback condition.

eanzhao (commented):
Shared testing infrastructure is in, but I don't think this finishes issue #264 yet.

What still looks missing to me:
- some conformance / fault cases still read stronger than what they actually verify, so adapters can regress and still pass
- gateway-only adapters still bypass part of the inbound matrix
- #264 acceptance is still missing the concrete `Aevatar.GAgents.Channel.Lark.Tests` / `Aevatar.GAgents.Channel.Telegram.Tests` derivations and CI hard-gating of the suite

So the current scope looks closer to "shared testing base classes landed" than "#264 fully completed".

test/Aevatar.GAgents.Channel.Testing/Conformance/ChannelAdapterConformanceTests.cs
```diff
@@ -0,0 +1,678 @@
...
+            return;
+
+        await using var lifetime = await StartAdapterAsync();
+        var seed = InboundActivitySeed.DirectMessage("dedupe-me") with
+        {
+            PlatformMessageId = "platform-msg-1",
+        };
+
+        var first = await DispatchAsync(seed);
+        var replay = await WebhookFixture!.ReplayLastInboundAsync();
```
eanzhao: This conformance case still breaks for gateway-only adapters. The guard now allows either inbound fixture, and `DispatchAsync(seed)` already supports `GatewayFixture`, but the replay step is still hard-wired to `WebhookFixture!.ReplayLastInboundAsync()`. So a gateway-only adapter will get past the guard and then hit a null dereference here instead of either exercising a gateway replay path or self-skipping as not applicable. If duplicate coverage is meant to stay transport-agnostic, `GatewayFixture` needs a replay hook too; otherwise this test should be explicitly webhook-scoped.

eanzhao (commented):
Reviewed the latest fix commit `8c14f739`.

The new `TypingProbe` path and the `EventGapDetected` assertion look like the right direction, and I resolved the three threads those changes addressed.

I did find one remaining gateway-only path that can still fail unexpectedly, so I left one more inline comment for that.

eanzhao (commented):
Reviewed the latest fix commit `1453bf3a`.

The new `ReplayLastAsync()` / `GatewayFixture.ReplayLastEventAsync()` path closes the remaining gateway-only duplicate gap, and the `CapabilityInterfaceParity` change now reads the adapter's real runtime interfaces instead of trusting an author-supplied bool.

I reran `dotnet test test/Aevatar.GAgents.Channel.Protocol.Tests/Aevatar.GAgents.Channel.Protocol.Tests.csproj --nologo` on this head and it still passes (`77/77`). I don't have new findings on this delta.

The earlier scope note still stands: this PR looks good for the shared testing infrastructure slice, while the concrete Lark / Telegram derivations and CI hard-gating remain follow-up work outside this PR.

PR #277, opened 2026-04-21
fix(auth): require authenticated caller by default; drop MockNyxId

eanzhao:
## Summary

First hardening pass toward turning Studio from a local-only tool into a product. Addresses the authentication-layer concerns raised in #251.

- **Authorization FallbackPolicy.** When `Aevatar:Authentication:Enabled = true`, endpoints now default to requiring an authenticated caller (`RequireAuthenticatedUser` on the JWT Bearer scheme). Previously the HTTP surface was "open unless an individual controller opted in with `[Authorize]`", which only `UserConfigController` did.
- **Explicit public opt-outs.** `GET /`, `/health/live`, `/health/ready`, `/api/openapi.json`, `/api/health`, `/api/auth/me`, `/api/app/context`, `/api/studio/context` are marked `.AllowAnonymous()` so the UI can still probe health and the login state without a token.
- **Disabled path unchanged.** When auth is disabled the fallback policy isn't registered, so existing anonymous dev flows continue to work.
- **Remove `Aevatar.Tools.MockNyxId`.** No in-tree consumer, and its `scope` → JWT `scope` claim mapping did not line up with the backend scope resolver (which reads `scope_id`). Removing it prevents future test plans from being written against a misleading fixture.

## What is NOT in this PR (follow-up)

- Phase 2: collapsing the `scope_id → uid → sub → NameIdentifier → *_id` fallback chain in `DefaultAppScopeResolver`.
- Phase 3: `/api/services/**` is still taking `tenantId/appId/namespace` from the request body. Adding `[Authorize]` does not plug that — the identity must come from JWT claims. Tracking separately.
- `ExecutionService.ListAsync/GetAsync` per-scope filtering (SEC-05 at the data layer) — this requires `IStudioWorkspaceStore` changes that deserve their own PR.
- `ChronoStorageServiceException` returns **503**, not 502 as the issue describes. Raw transport exceptions still return 502. Left as a note on #251 rather than a code change.
- Rollout Pause/Resume idempotency — code is correct, only the test assertion in TC-SCOPE-SVC-006 needs updating on the test repo side.

Detailed answers to Q-01 … Q-08 posted in a comment on #251.

## Test plan

- [x] `dotnet build aevatar.slnx --nologo`
- [x] `dotnet test test/Aevatar.Bootstrap.Tests/Aevatar.Bootstrap.Tests.csproj` — 46/46 pass (including 2 new FallbackPolicy tests)
- [x] `dotnet test test/Aevatar.Tools.Cli.Tests/Aevatar.Tools.Cli.Tests.csproj` — 339/339 pass
- [x] `dotnet test test/Aevatar.Studio.Tests/Aevatar.Studio.Tests.csproj` — 203/203 pass
- [ ] Manually hit `/health/live`, `/api/auth/me` against a host with auth enabled to confirm `.AllowAnonymous()` still returns 200 without a token
- [ ] Manually hit `/api/roles` without a token on the same host and confirm 401


eanzhao (commented):
There is already an open thread on the webhook routes under the new fallback policy, so I did not duplicate that one. I am adding one additional concern below: the new `scope_id`-only resolver now returns `null` for some authenticated principals, but several Studio call sites still treat `null` as global/default access instead of failing closed.

src/Aevatar.Studio.Infrastructure/ScopeResolution/AppScopeResolver.cs
```diff
@@ -50,19 +53,9 @@ public DefaultAppScopeResolver(
...
-                        return new AppScopeContext(claimValue, $"claim:{claimType}");
-                }
-
-                var genericIdClaim = user.Claims.FirstOrDefault(claim =>
-                    claim.Type.EndsWith("_id", StringComparison.OrdinalIgnoreCase) &&
-                    !IgnoredGenericIdClaimTypes.Contains(claim.Type) &&
-                    !string.IsNullOrWhiteSpace(claim.Value));
-                if (genericIdClaim != null)
-                    return new AppScopeContext(genericIdClaim.Value.Trim(), $"claim:{genericIdClaim.Type}");
+                var scopeIdValue = user.FindFirst(ScopeIdClaimType)?.Value?.Trim();
```
eanzhao, line 56: Tightening the resolver to the canonical `scope_id` is reasonable, but it is only safe if every consumer fails closed when this lookup misses. That is not true today: `ExecutionService` treats a `null` scope as workspace-global visibility, and several scope-backed stores still map it to the shared `default` actor. So an authenticated token that passes JWT validation but misses provider claim mapping can still read or write shared data instead of getting rejected.

src/Aevatar.Studio.Application/Studio/Services/ExecutionService.cs
```diff
@@ -46,14 +51,40 @@ public async Task<IReadOnlyList<ExecutionSummary>> ListAsync(CancellationToken c
...
+    }
+
+    private string? ResolveCurrentScopeId() =>
+        _scopeResolver?.Resolve()?.ScopeId?.Trim();
+
+    private static bool IsVisibleToCurrentScope(StoredExecutionRecord record, string? scopeFilter)
+    {
+        // When no scope resolver is wired (test doubles, legacy hosts) fall back to the historical
+        // "workspace-global" view so existing callers continue to see every execution.
+        if (scopeFilter is null)
```
eanzhao: This is the concrete fail-open path. After `DefaultAppScopeResolver` stopped guessing from `uid`/`sub`/`*_id`, a missing `scope_id` no longer means only `no resolver wired`; it can also mean `authenticated request with broken claim mapping`. Returning every execution here lets that caller enumerate workspace-global data, and `StartAsync` has the same issue because the mismatch check only runs when `callerScopeId != null`. Once auth is enabled this should deny rather than expand visibility.

src/Aevatar.Authentication.Hosting/AevatarAuthenticationHostExtensions.cs
```diff
@@ -36,7 +37,14 @@ public static WebApplicationBuilder AddAevatarAuthentication(this WebApplication
                 jwt.TokenValidationParameters.ValidateAudience = !string.IsNullOrWhiteSpace(options.Audience);
             });
 
-        builder.Services.AddAuthorization();
+        // When authentication is enabled, endpoints default to requiring an authenticated caller.
+        // Public endpoints must opt out with [AllowAnonymous] / .AllowAnonymous().
+        builder.Services.AddAuthorization(authorization =>
+        {
+            authorization.FallbackPolicy = new AuthorizationPolicyBuilder(JwtBearerDefaults.AuthenticationScheme)
+                .RequireAuthenticatedUser()
```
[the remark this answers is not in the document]
eanzhao: Addressed in 487d5fab. Marked the externally-invoked webhook routes `.AllowAnonymous()` so they stay open under the new fallback policy:

- `POST /api/channels/{platform}/callback/{registrationId}` — platform webhooks (Lark/Telegram)
- `POST /api/device-events/{registrationId}` — NyxID device relay (HMAC-authenticated)
- `POST /api/webhooks/nyxid-relay` plus `/health` and `/diag` siblings

Registration CRUD + the diagnostics endpoint stay behind auth. Confirmed with `Aevatar.GAgents.ChannelRuntime.Tests` (229/229).

src/Aevatar.Studio.Infrastructure/ScopeResolution/AppScopeResolver.cs
eanzhao: Addressed in 487d5fab.

Added `IAppScopeResolver.HasAuthenticatedRequestWithoutScope(HttpContext?)` which returns true when the resolver sees an authenticated principal but no `scope_id` claim — i.e. the provider's claims transformer missed this principal. Callers can now distinguish "out-of-request / CLI" (safe fall-through) from "authenticated request with broken claim mapping" (fail closed).

Studio-side consumers still need their own pass — this commit only rewires `ExecutionService` (see reply on the other thread). The scope-backed stores that treat missing scope as a shared `default` actor are a separate issue; flagging them as a follow-up so we don't expand the PR further.

src/Aevatar.Studio.Application/Studio/Services/ExecutionService.cs
eanzhao: Addressed in 487d5fab.

`ExecutionService` now derives a `ScopeFilter` with an explicit `FailClosed` bit before touching the store. When the resolver reports the request is authenticated but unscoped:

- `ListAsync` returns an empty list (no records enumerated)
- `GetAsync` / `ResumeAsync` / `StopAsync` return null → 404
- `StartAsync` throws `InvalidOperationException` ("Authenticated caller has no resolvable scope") → 400 at the controller boundary

The "no resolver wired" path (tests, CLI, out-of-request) keeps the historical workspace-global view — distinguished by `HasAuthenticatedRequestWithoutScope` returning false when there is no authenticated principal in flight.

New tests cover the three fail-closed variants (`ListAsync_WhenAuthenticatedCallerHasNoScope_ShouldFailClosed`, `GetAsync_WhenAuthenticatedCallerHasNoScope_ShouldReturnNull`, `StartAsync_WhenAuthenticatedCallerHasNoScope_ShouldThrow`). `Aevatar.Tools.Cli.Tests` 350/350, Studio 203/203, Bootstrap 46/46.

src/Aevatar.Studio.Infrastructure/ScopeResolution/AppScopeResolver.cs
eanzhao: Follow-up (7be58c04): ended up expanding the pass to cover the scope-backed catalog stores too. `ResolveScopeIdOrDefault` now throws `InvalidOperationException` when `HasAuthenticatedRequestWithoutScope` is true, so `ActorBackedRoleCatalogStore`, `ActorBackedConnectorCatalogStore`, and `ActorBackedGAgentActorStore` no longer route authenticated-but-unscoped callers into the shared `*-default` actor. The "default" fallback stays only for out-of-request contexts (CLI / tests / unauthenticated dev).

Exception maps cleanly to 400 at the Roles / Connectors controller boundaries (existing `catch InvalidOperationException -> BadRequest`). New test: `ConnectorCatalogStore_SaveCatalog_WhenAuthenticatedWithoutScope_Throws`. All tests still green: Tools.Cli 351/351, Bootstrap 46/46, Studio 203/203.

src/Aevatar.Studio.Application/Studio/Services/ExecutionService.cs
```diff
@@ -20,40 +20,96 @@ public sealed class ExecutionService
...
+
     public async Task<ExecutionDetail> StartAsync(StartExecutionRequest request, CancellationToken cancellationToken = default)
     {
         if (!ShouldUsePublishedWorkflowRun(request))
             throw new InvalidOperationException("scopeId and workflowId are required. Executions must target a registered scope service.");
 
+        var requestedScopeId = request.ScopeId!.Trim();
+        var scope = GetScopeFilter();
+        if (scope.FailClosed)
+            throw new InvalidOperationException("Authenticated caller has no resolvable scope; refuse to start a scoped execution.");
```
eanzhao, line 109: **P2** `StartAsync` now throws for the authenticated-but-unscoped case, but `ExecutionsController.Start` still forwards this call directly with `Ok(await _executionService.StartAsync(...))` and does not catch `InvalidOperationException` the way `RolesController` / `ConnectorsController` do. In the current host pipeline this turns the new fail-closed branch into a 500, not the 400 described in the thread reply. Please either translate this exception at the controller boundary or return a result shape that preserves the intended client error semantics, and add an API-level test for `POST /api/executions` to lock it down.

eanzhao (commented):
Re-reviewed the latest fix. The previous fail-open paths look addressed, but one HTTP boundary issue remains on the execution start path.

src/Aevatar.Studio.Application/Studio/Services/ExecutionService.cs
eanzhao: Addressed in c1428a89.

`ExecutionsController.Start` / `Resume` / `Stop` now wrap `_executionService.*Async` in `try { ... } catch (InvalidOperationException) { return BadRequest(...) }` — matching the pattern already in place on `RolesController` / `ConnectorsController`. The fail-closed and scope-mismatch branches surface as **400** with the original message, not 500.

Added `test/Aevatar.Tools.Cli.Tests/ExecutionsControllerTests.cs` covering `POST /api/executions`:
- `Start_WhenAuthenticatedCallerHasNoScope_ShouldReturnBadRequest` — locks in the fail-closed path
- `Start_WhenRequestedScopeDoesNotMatchAuthenticatedScope_ShouldReturnBadRequest` — locks in the cross-scope guard
- `Start_WhenScopeOrWorkflowMissing_ShouldReturnBadRequest` — locks in the existing contract check

All suites still green: Tools.Cli 354/354, Bootstrap 46/46, Studio 203/203.

PR #280, opened 2026-04-21
Refactor/frontend

potter-sun:


src/Aevatar.Studio.Hosting/Controllers/WorkspaceController.cs
```diff
@@ -119,11 +119,54 @@ public async Task<ActionResult<IReadOnlyList<WorkflowSummary>>> ListWorkflows(
...
-    [HttpGet("workflows/{workflowId}")]
-    public async Task<ActionResult<WorkflowFileResponse>> GetWorkflow(
+    #pragma warning disable CS0618
+    [Obsolete("Use /api/workspace/workflow-drafts.")]
+    [HttpGet("workflows")]
+    public async Task<ActionResult<IReadOnlyList<WorkflowSummary>>> ListWorkflows(
+        [FromQuery] string? scopeId,
+        CancellationToken cancellationToken)
+    {
+        var result = await ListDrafts(scopeId, cancellationToken);
```
eanzhao, line 132: If we keep `GET /api/workspace/workflows` as a compatibility alias, this cannot delegate to `ListDrafts()`. The old scoped behavior came from `_scopeWorkflowService.ListAsync(...)`, which merged committed scope workflows with stored drafts. With the current code, any legacy caller that still uses the old route with `scopeId` will silently lose committed-only workflows.

src/Aevatar.Studio.Hosting/Controllers/WorkspaceController.cs
```diff
@@ -151,15 +194,43 @@ public async Task<ActionResult<WorkflowFileResponse>> GetWorkflow(
...
-        [FromBody] SaveWorkflowFileRequest request,
+    #pragma warning disable CS0618
+    [Obsolete("Use /api/workspace/workflow-drafts/{workflowId}.")]
+    [HttpGet("workflows/{workflowId}")]
+    public async Task<ActionResult<WorkflowFileResponse>> GetWorkflow(
+        string workflowId,
+        [FromQuery] string? scopeId,
+        CancellationToken cancellationToken)
+    {
+        var result = await GetDraft(workflowId, scopeId, cancellationToken);
```
eanzhao, line 211: This is another compatibility regression for the obsolete route: `GetDraft()` only checks the draft store, while the old `GET /api/workspace/workflows/{workflowId}` fell back to committed scope workflows through `AppScopedWorkflowService.GetAsync(...)`. A legacy scoped caller now gets `404` for a published workflow that has no draft yet.

src/Aevatar.Studio.Hosting/Controllers/WorkspaceController.cs
```diff
@@ -249,9 +421,39 @@ public async Task<IActionResult> DeleteWorkflow(
...
+        code = "WORKFLOW_DRAFT_PATH_CONFLICT",
+        message = exception.Message,
+    };
+
+    #pragma warning disable CS0618
+    private static WorkflowFileResponse ToLegacyWorkflowFileResponse(WorkflowDraftResponse? draft)
+    {
+        ArgumentNullException.ThrowIfNull(draft);
+
+        return new WorkflowFileResponse(
```
eanzhao, line 445: This helper no longer preserves the legacy `WorkflowFileResponse` shape. The previous route returned parsed `document` and `findings`; here they are hard-coded to `null` / empty, so any caller still depending on the obsolete path loses validation and graph/source metadata. If the alias stays, it needs to parse the YAML before projecting back to `WorkflowFileResponse`.

eanzhao (changes requested):
The new draft-only routes make sense, but the obsolete `/api/workspace/workflows` aliases are no longer preserving the old contract. In scoped mode they now stop surfacing committed-only workflows, and the legacy `WorkflowFileResponse` path now drops `document/findings`. That makes the alias unsafe for any staged migration.

src/Aevatar.Studio.Hosting/Controllers/WorkspaceController.cs
louis4li: Tracked in #284.

This comment is about preserving the old `/api/workspace/workflows` scoped contract if we keep the obsolete alias as a staged migration layer. The current direction is not to preserve that alias long-term. `refactor/frontend` still has active frontend callers on `/api/workspace/workflows*`, so the plan is:

1. migrate frontend callers to `/api/workspace/workflow-drafts*`
2. merge that frontend work
3. delete the obsolete alias routes entirely

So we should not invest in re-implementing the legacy merged scoped behavior here unless we decide to keep the alias around permanently.

src/Aevatar.Studio.Hosting/Controllers/WorkspaceController.cs
louis4li: Tracked in #284.

Same classification here: this is a legacy alias compatibility concern, not a problem with the new draft route itself. The current frontend branch still calls `/api/workspace/workflows*`, so removing the alias immediately would break that in-flight branch. Once frontend is moved to `/api/workspace/workflow-drafts*`, the intended cleanup is to remove the obsolete alias instead of restoring committed-workflow fallback on the old route.

src/Aevatar.Studio.Hosting/Controllers/WorkspaceController.cs
louis4li: Tracked in #284.

Agreed on the premise: if we keep the obsolete `WorkflowFileResponse` alias, it should preserve the old response contract instead of dropping `document/findings`.

The planned direction is to avoid carrying that compatibility layer forward. Frontend callers are still on `/api/workspace/workflows*` today, so we need to migrate them first; after that, the cleaner follow-up is to remove the obsolete alias routes and their legacy response adapter entirely.

louis4li:
Follow-up tracked in #284.

The review comments on `/api/workspace/workflows*` are compatibility-contract concerns around the obsolete alias routes. Current direction:

1. migrate active frontend callers on `refactor/frontend` from `/api/workspace/workflows*` to `/api/workspace/workflow-drafts*`
2. merge that frontend work
3. remove the obsolete alias routes entirely instead of rebuilding full legacy scoped semantics on them

That keeps the intended API surface on the new draft routes and avoids carrying a half-compatibility layer forward.

louis4li (approved):

eanzhao (approved):
同意使用新issue追踪遗留问题：https://github.com/aevatarAI/aevatar/issues/284

PR #311, opened 2026-04-22
Fix service identity spoofing for /api/services endpoints

### LINKED ISSUE #281: /api/services/** takes tenant/app/namespace from request body, allowing identity spoofing
## Background

Discovered during the #251 API design review.

## Problem

`Aevatar.GAgentService.Hosting.Endpoints.ServiceServingEndpoints` (and its sibling command/query endpoints under `/api/services/**`) take tenant / app / namespace from the **request body or query string**, not from the authenticated principal's JWT claims:

`src/platform/Aevatar.GAgentService.Hosting/Endpoints/ServiceServingEndpoints.cs`:
- `HandleDeployRevisionAsync` — line 37: `ToIdentity(request.TenantId, request.AppId, request.Namespace, serviceId)`
- `HandleReplaceServingTargetsAsync` — line 76
- `HandleStartRolloutAsync` — line 101
- `HandleAdvanceRolloutAsync` — line 121
- …and every other handler in the file.

The same pattern exists in `ServiceEndpoints.cs`, `ScopeServiceEndpoints.cs`, and in the governance endpoints (`ServicePolicyEndpoints.cs`, `ServiceEndpointCatalogEndpoints.cs`, `ServiceBindingEndpoints.cs`).

PR #277 adds an authorization fallback policy so these endpoints now require **an authenticated** caller. But an authenticated caller can still cross-tenant: they just write whatever `tenantId` / `appId` / `namespace` they want into the request body, and the handler happily builds a `ServiceIdentity` from it.

## Impact

- Cross-tenant read & write on the entire service serving / rollout surface.
- Any `[Authorize]` style fix that doesn't change the identity wiring is cosmetic.
- Blocks cross-tenant isolation assertions in #250 test plan (`TC-SCOPE-SVC-*`).

## Proposed direction

1. Introduce standard claims for tenant / app / namespace (mirroring `scope_id`). The auth provider's claims transformer (e.g. `NyxIdClaimsTransformer`) maps provider-specific JWT claims into them.
2. Add a minimal resolver — parallel to `IAppScopeResolver` — that returns a `ServiceIdentityContext` from the current `HttpContext`.
3. Change every `/api/services/**` handler: drop `TenantId` / `AppId` / `Namespace` from the HTTP request body / query, replace with the resolver.
4. For admin-level overrides (if any legitimate use cases exist), gate them behind an explicit admin policy.
5. Update clients that were sending tenant/app/namespace in the body (SDK, bridge, test fixtures).

## Why this is separate from #277

Touches every `/api/services/**` handler (~30+ call sites across hosting + governance), plus SDK/test fixtures that assume body-based identity. Worth batching into its own PR with explicit migration notes.

## Relates to
- #251 — Q-07
- #250 — TC-SCOPE-SVC-* cross-tenant isolation

eanzhao:
## Summary
- enforce `tenant_id` / `app_id` / `namespace` claims as the authoritative identity for authenticated `/api/services/**` and governance requests
- reject authenticated requests with missing or ambiguous service identity claims instead of accepting spoofed body/query identity
- update CLI callers and bundled frontend assets to stop depending on `tenantId/appId/namespace` query parameters for service listing and bindings lookup
- add integration coverage for authenticated claim precedence and missing-claim denial paths across service and governance endpoints

## Context
- follows the API design review context in #251
- closes #281

## Verification
- `dotnet test test/Aevatar.GAgentService.Integration.Tests/Aevatar.GAgentService.Integration.Tests.csproj --filter "FullyQualifiedName~ServiceEndpointsTests|FullyQualifiedName~GovernanceEndpointsTests" --nologo`
- `dotnet build src/platform/Aevatar.GAgentService.Hosting/Aevatar.GAgentService.Hosting.csproj --nologo`
- `dotnet build tools/Aevatar.Tools.Cli/Aevatar.Tools.Cli.csproj --nologo`
- `npm ci`
- `npm run build`


eanzhao:
Reviewed. Two real issues and one question:

### 1. `HandleInvokeAsync` still trusts caller identity from the body (security)

`src/platform/Aevatar.GAgentService.Hosting/Endpoints/ServiceEndpoints.cs:374-379`

```csharp
Caller = new ServiceInvocationCaller
{
    ServiceKey = request.CallerServiceKey ?? string.Empty,
    TenantId   = request.CallerTenantId   ?? string.Empty,
    AppId      = request.CallerAppId      ?? string.Empty,
},
```

This PR hardens the **target** `tenant/app/namespace` via claims, but the **caller** identity on the same endpoint is still taken verbatim from the request body. `Caller.ServiceKey` then flows into `DefaultInvokeAdmissionEvaluator` (`src/platform/Aevatar.GAgentService.Governance.Infrastructure/Admission/DefaultInvokeAdmissionEvaluator.cs:58-67`):

```csharp
var callerServiceKey = request.Caller?.ServiceKey ?? string.Empty;
if (policy.InvokeAllowedCallerServiceKeys.Contains(callerServiceKey))
    continue;
// …otherwise caller_not_allowed
```

So an authenticated user can bypass `InvokeAllowedCallerServiceKeys` by sending any string in `callerServiceKey`. That's the same spoofing class the PR targets; it needs to be derived from claims (or from an authenticated caller-service identity), not taken from the body. Also no integration test covers this path — it's worth one analogous to the new `CreateServiceAsync_WhenAuthenticatedIdentityConflictsWithBody_ShouldUseClaimIdentity`.

### 2. Silent tenant rewrite of `BoundServiceRef` (question / DTO hygiene)

`src/platform/Aevatar.GAgentService.Governance.Hosting/Endpoints/ServiceBindingEndpoints.cs:152-167` + test `GovernanceEndpointsTests.BindingEndpoints_WhenAuthenticatedIdentityConflictsWithBody_ShouldUseClaimIdentityForOwnerAndBoundService`

When authenticated, the bound service's `TenantId/AppId/Namespace` are silently overwritten with the caller's claims, **even if the client explicitly set `request.Service.TenantId/AppId/Namespace` to a different tenant**. The body value `serviceId` is kept, so the resulting binding may point at a non-existent service in the caller's own tenant.

If cross-tenant bindings are illegal by design, fine — then please (a) drop `TenantId/AppId/Namespace` from `BoundServiceHttpRequest` (the fields are dead weight and mislead clients), or (b) return 400 on explicit mismatch instead of silently rewriting. The current behavior is encoded in the test but surprises a caller who thought they were binding across tenants.

### 3. Silent override vs explicit 400 on claim/body mismatch (minor)

`ServiceIdentityEndpointAccess.TryResolveContext` / the various `TryResolveIdentity` callers silently let the claim identity win whenever body/query disagree. A 400 "body identity does not match authenticated claims" would make client bugs loud without weakening the security posture; right now, a misconfigured client cross-posting to the wrong tenant gets 202 and writes to the authenticated tenant.

### Nits
- `IServiceIdentityContextResolver.Resolve()` is called twice per request on bindings / catalog / policy create/update (once in `TryResolveContext`, again in `ToSpec`). Pass the resolved `ServiceIdentityContext` through instead.
- `DefaultServiceIdentityContextResolver.TryGetSingleClaimValue` uses `StringComparison.OrdinalIgnoreCase` for claim type matching. JWT claim names are case-sensitive by convention; case-insensitive matching is harmless today but slightly surprising.
- `ServiceEndpoints.cs:391` `ResolveServiceIdentity` builds an otherwise-consumed `ServiceRevisionSpec` even when denied — not a bug (denied is checked before use), just wasted work.

eanzhao (commented):
Re-reviewed latest head `2415119`.

1. `src/platform/Aevatar.GAgentService.Hosting/Endpoints/ServiceEndpoints.cs:384-388` still trusts `callerServiceKey` / `callerTenantId` / `callerAppId` from the request body. `DefaultInvokeAdmissionEvaluator` authorizes against `request.Caller.ServiceKey`, so an authenticated caller can still spoof an allowed caller key and bypass `InvokeAllowedCallerServiceKeys`. This PR hardens the target service identity, but the caller identity on the same invoke path is still body-controlled. Please derive caller identity from authenticated service claims (or another authenticated caller-service contract) and add an integration test for the authenticated conflict case.

2. `src/platform/Aevatar.GAgentService.Governance.Hosting/Endpoints/ServiceBindingEndpoints.cs:152-164` still silently rewrites `request.Service.{TenantId,AppId,Namespace}` to the caller claims while keeping `serviceId`. The new test `GovernanceEndpointsTests.BindingEndpoints_WhenAuthenticatedIdentityConflictsWithBody_ShouldUseClaimIdentityForOwnerAndBoundService` locks this in. If cross-tenant bindings are illegal, this should reject explicit mismatches or those fields should be removed from the DTO; otherwise a client can think it bound to another tenant's service while the server actually creates an in-tenant reference to `serviceId` that may not exist.

PR #344, opened 2026-04-23
Bridge Day One slash commands over Nyx relay (#341)

### LINKED ISSUE #341: [Bridge] Day One over Nyx relay without NyxID changes or stored full_key
## Context

`aevatarAI/aevatar#328` 的正统目标是把 Nyx relay 降级为 transport adapter，统一回到 `ConversationGAgent` 主干。

但这条线当前被 Nyx relay outbound credential 卡住：

- Aevatar provisioning 现在只保存 `nyx_agent_api_key_id`，不保存 Nyx 一次性返回的 `full_key`
- 我们不打算让 Aevatar 持久化 `full_key`
- NyxID 侧的正式增强提案在 `ChronoAIProject/NyxID#469`

产品上不能等完整重构完成后再修。Day One 当前最急的用户可见需求，是 **Nyx relay 生产路径上的文本命令面可用且不幻觉**。

## Goal

在 **不改 NyxID**、**Aevatar 不存 `full_key`** 的前提下，交付一条 **bridge 模式**，让 Nyx relay 生产路径满足 Day One 的最小可用需求。

## Non-goals

这个 issue **不是** `#328` 的替代品，不追求以下能力：

- 不把 Nyx relay endpoint 瘦成纯 shim
- 不把生产 inbound 切回 `ConversationGAgent`
- 不获得 `ConversationTurnCompletedEvent` / projection / unified dedup 主链收益
- 不实现 durable async reply / 进程重启后补发 / 长时重试
- 不引入 Aevatar 侧 secret store 来保存 Nyx agent `full_key`

### 1. Day One slash / text-command surface works on production Nyx relay path

以下文本命令在 Nyx relay 路径上应走 **确定性短路**，不落 LLM：

- `/daily-report`
- `/social-media`
- `/templates`
- `/agents`
- `/agent-status <agent_id>`
- `/run-agent <agent_id>`
- `/disable-agent <agent_id>`
- `/enable-agent <agent_id>`
- `/delete-agent <agent_id> confirm`

### 2. chat_type is propagated from Nyx callback into Day One logic

Nyx callback 已带 `conversation.type`（`private` / `group` / `channel` / `device`）。bridge 模式下，endpoint 必须把它翻译成 Day One 当前消费的 `channel.chat_type`：

- `private` -> `p2p`
- `group` -> `group`
- `channel` -> `group`（Day One bridge 阶段先不新增第三种 chat_type）
- `device` -> 不进入 Day One command flow

### 3. unknown slash command must short-circuit

对以 `/` 开头但未命中已注册命令的文本：

- 不交给 LLM
- 返回 deterministic unknown-command usage reply

这一步是为了修 `/daily_report` 这类输入被 LLM 误套模板的问题。

### 4. free-text LLM fallback stays on the current endpoint-local path

对不属于 Day One 命令面的自由文本：

- 继续沿用现有 `/api/webhooks/nyxid-relay` -> `NyxIdChatGAgent` -> `SendChannelRelayTextReplyAsync` 路径
- 继续使用 callback 自带的 `X-NyxID-User-Token`
- 不把该 token 持久化为长期凭证

## Hard Boundary Of This Bridge

这条 bridge 的能力边界必须写清楚：

- **reply capability 只在当前 callback 派生出的短生命周期内有效**
- **不能承诺** 进程重启后补发
- **不能承诺** 超过 relay JWT TTL 后仍可回包
- **不能承诺** endpoint 之外的 durable continuation / replay-safe outbound

也就是说，它是为了交付 Day One 用户面，不是为了交付架构正确性。

### A. Keep business orchestration inside `NyxIdChatEndpoints` for now

在 bridge 阶段，允许 `/api/webhooks/nyxid-relay` 继续承载 relay 业务编排；不要强行把这条路径改造成 `#328` 想要的 endpoint shim。

### B. Add a relay-to-Day-One bridge helper before the LLM path

在 relay endpoint 里新增一个 bridge helper，顺序如下：

1. 验证 `X-NyxID-User-Token`
2. 解析 relay payload
3. 从 `conversation.type` 派生 `channel.chat_type`
4. 构造 `ChannelInboundEvent`
5. 先调用 `NyxRelayAgentBuilderFlow.TryResolve(...)`
6. 若命中：
   - 用 `AgentToolRequestContext` 注入
     - `LLMRequestMetadataKeys.NyxIdAccessToken = X-NyxID-User-Token`
     - `ChannelMetadataKeys.ChatType`
     - `ChannelMetadataKeys.ConversationId`
     - `ChannelMetadataKeys.MessageId`
     - `ChannelMetadataKeys.Platform`
     - `scope_id`
   - 复用 `AgentBuilderTool`
   - 把结果格式化后，直接走当前 relay async reply path 回 Nyx
7. 若未命中：落回现有 `NyxIdChatGAgent` 自由对话路径

### C. Widen slash detection before p2p gating

现有 `NyxRelayAgentBuilderFlow` 入口先硬卡 `p2p`，这会让 group unknown slash 直接掉回 LLM。

bridge 阶段应调整为：

1. 先识别“是否是 slash command”
2. 再按 `chat_type` 决定：
   - 已知 Day One 命令在 group/channel 下返回 private-chat restriction
   - 未知 slash 返回 unknown-command usage
   - 非 slash 文本才允许 fall through 到 LLM

### D. Do not introduce full_key storage or relay-token cache service

bridge 阶段禁止：

- 保存 Nyx agent `full_key`
- 新增进程级 `message_id -> token` 事实缓存
- 把 callback token 写进 read model / projection / event store

token 只能活在当前 request / 当前短生命周期后台任务里。

### User-visible

- Lark p2p 发 `/daily-report alice`：命中 Day One 命令流，不落 LLM
- Lark group 发 `/daily-report alice`：返回 private-chat restriction，不落 LLM
- Lark p2p 发 `/daily_report` 或 `/foobar`：返回 unknown-command usage，不落 LLM
- 自由文本 `hello`：继续落现有 LLM fallback

### Security / boundary

- Aevatar 不存 Nyx agent `full_key`
- Aevatar 不新增长期 reply credential 存储
- bridge 逻辑只消费 callback 自带的 `X-NyxID-User-Token`

### Tests

新增或调整测试至少覆盖：

- relay payload `conversation.type` -> `channel.chat_type` 映射
- p2p known slash -> deterministic tool/direct reply
- group known slash -> private-chat restriction
- unknown slash -> deterministic unknown-command reply
- non-slash text -> falls through to LLM path
- bridge path不需要 `full_key`

## Related

- `aevatarAI/aevatar#328` 正统 unified inbound backbone
- `ChronoAIProject/NyxID#469` NyxID 侧 reply-token 增强

## Exit Criteria

当 `NyxID#469` 或等价能力落地后，这条 bridge issue 的产物可以作为过渡实现，后续再由 `#328` 吸收或删除。

kaihuei0114:
Resolves aevatarAI/aevatar#341.

## Summary
- Intercept Day One slash commands (`/daily`, `/social-media`, `/agents`, …) on the Nyx relay production path before they reach the LLM. Free-text keeps falling through to the existing `NyxIdChatGAgent` flow.
- Rename `/daily-report` → `/daily` across the bridge flow, card flow, help text, system prompt and tests. `/create-daily-report` alias dropped per the deletion-first rule.
- Widen `NyxRelayAgentBuilderFlow.TryResolve` to gate on slash prefix first, then `chat_type`:
  - known command in `group`/`channel` → deterministic private-chat restriction
  - unknown `/foobar`, `/daily_report`, etc. → deterministic unknown-command usage
  - non-slash text → fall through to LLM
- Add `INyxRelayDayOneBridge` (NyxidChat) with sync `ShouldHandle` gate + async `HandleAsync` resolver, implemented by `NyxRelayDayOneBridge` (ChannelRuntime). Maps Nyx callback `conversation.type` → `channel.chat_type` (`private → p2p`, `group/channel → group`, `device → skip`) and reuses `AgentBuilderTool` with a per-request `AgentToolRequestContext`.
- Wire the bridge into `/api/webhooks/nyxid-relay`: when the bridge owns a message the endpoint skips actor dispatch, returns `202` with `bridge=day_one_command`, and delivers the reply through the current callback's relay reply channel.

## Non-goals (from #341)
- No Aevatar-side storage of the Nyx agent `full_key`.
- No long-term reply-credential cache, no projection/read-model write for callback tokens, no durable async-reply continuation.
- Bridge stays a disposable helper so `#328` can absorb or remove it once `NyxID#469` (or equivalent) unlocks the proper outbound credential story.

## Comment follow-ups
- Bridge is an extractable flow (dedicated class, thin endpoint wiring) so a future unified inbound backbone can lift it without rewriting from scratch.
- Slash surface stays text-first. Bootstrap-card UX for first-time `/daily` without saved GitHub username is intentionally **not** added here — it needs owner-scoped Day One preference storage (not conversation-scoped), which belongs with `#336` / the unified backbone rather than in this bridge.
- Group/channel known slash returns a deterministic DM handoff instead of letting the LLM make something up.

## Test plan
- [x] `dotnet test test/Aevatar.GAgents.ChannelRuntime.Tests` — 193 pass (2 pre-existing `ServiceCollectionExtensionsTests` failures unrelated to this PR, verified by stashing)
- [x] `dotnet test test/Aevatar.AI.Tests --filter NyxIdChatEndpointsCoverageTests` — 40/40 pass (includes new bridge-intercept + defer-to-actor coverage)
- [x] `bash tools/ci/architecture_guards.sh` — pass
- [x] `bash tools/ci/test_stability_guards.sh` — pass

### DIFF
--- a/agents/Aevatar.GAgents.ChannelRuntime/NyxRelayDayOneBridge.cs
+++ b/agents/Aevatar.GAgents.ChannelRuntime/NyxRelayDayOneBridge.cs
@@ -1,0 +1,108 @@
+using Aevatar.AI.Abstractions.LLMProviders;
+using Aevatar.AI.Abstractions.ToolProviders;
+using Aevatar.GAgents.NyxidChat.Relay;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Aevatar.GAgents.ChannelRuntime;
+
+internal sealed class NyxRelayDayOneBridge : INyxRelayDayOneBridge
+{
+    private const string PrivateChatType = "p2p";
+    private const string GroupChatType = "group";
+    private const string DeviceConversationType = "device";
+
+    private readonly IServiceProvider _services;
+
+    public NyxRelayDayOneBridge(IServiceProvider services)
+    {
+        _services = services ?? throw new ArgumentNullException(nameof(services));
+    }
+
+    public bool ShouldHandle(NyxRelayBridgeRequest request)
+    {
+        ArgumentNullException.ThrowIfNull(request);
+
+        if (string.IsNullOrWhiteSpace(request.Text))
+            return false;
+        if (!request.Text.TrimStart().StartsWith('/'))
+            return false;
+        if (IsDevice(request.ConversationType))
+            return false;
+
+        return true;
+    }
+
+    public async Task<string> HandleAsync(NyxRelayBridgeRequest request, CancellationToken ct)
+    {
+        ArgumentNullException.ThrowIfNull(request);
+
+        var chatType = MapChatType(request.ConversationType);
+        var inboundEvent = new ChannelInboundEvent
+        {
+            Text = request.Text.TrimStart(),
+            ConversationId = request.ConversationId ?? string.Empty,
+            MessageId = request.MessageId ?? string.Empty,
+            ChatType = chatType,
+            Platform = request.Platform ?? string.Empty,
+            SenderId = request.SenderId ?? string.Empty,
+            SenderName = request.SenderName ?? string.Empty,
+            RegistrationScopeId = request.ScopeId,
+        };
+
+        if (!NyxRelayAgentBuilderFlow.TryResolve(inboundEvent, out var decision) || decision is null)
+        {
+            throw new InvalidOperationException(
+                "NyxRelayDayOneBridge.HandleAsync was invoked for a message that ShouldHandle did not own.");
+        }
+
+        if (!decision.RequiresToolExecution)
+            return decision.ReplyPayload;
+
+        var previousMetadata = AgentToolRequestContext.CurrentMetadata;
+        try
+        {
+            AgentToolRequestContext.CurrentMetadata = BuildToolMetadata(request, chatType);
+            var tool = ActivatorUtilities.CreateInstance<AgentBuilderTool>(_services);
+            var toolResult = await tool.ExecuteAsync(decision.ToolArgumentsJson!, ct);
+            return NyxRelayAgentBuilderFlow.FormatToolResult(decision, toolResult);
+        }
+        finally
+        {
+            AgentToolRequestContext.CurrentMetadata = previousMetadata;
+        }
+    }
+
+    private static bool IsDevice(string? conversationType) =>
+        !string.IsNullOrWhiteSpace(conversationType) &&
+        string.Equals(conversationType.Trim(), DeviceConversationType, StringComparison.OrdinalIgnoreCase);
+
+    private static string MapChatType(string? conversationType)
+    {
+        if (string.IsNullOrWhiteSpace(conversationType))
+            return PrivateChatType;
+
+        return conversationType.Trim().ToLowerInvariant() switch
+        {
+            "private" => PrivateChatType,
+            "group" or "channel" => GroupChatType,
+            _ => PrivateChatType,
+        };
+    }
+
+    private static IReadOnlyDictionary<string, string> BuildToolMetadata(
+        NyxRelayBridgeRequest request,
+        string chatType)
+    {
+        return new Dictionary<string, string>(StringComparer.Ordinal)
+        {
+            [LLMRequestMetadataKeys.NyxIdAccessToken] = request.NyxIdAccessToken,
+            [ChannelMetadataKeys.Platform] = request.Platform ?? string.Empty,
+            [ChannelMetadataKeys.SenderId] = request.SenderId ?? string.Empty,
+            [ChannelMetadataKeys.SenderName] = request.SenderName ?? string.Empty,
+            [ChannelMetadataKeys.ConversationId] = request.ConversationId ?? string.Empty,
+            [ChannelMetadataKeys.MessageId] = request.MessageId ?? string.Empty,
+            [ChannelMetadataKeys.ChatType] = chatType,
+            ["scope_id"] = request.ScopeId,
+        };
+    }
+}
--- a/agents/Aevatar.GAgents.ChannelRuntime/NyxRelayDayOneBridge.cs
+++ b/agents/Aevatar.GAgents.ChannelRuntime/NyxRelayDayOneBridge.cs
@@ -76,16 +76,19 @@ internal sealed class NyxRelayDayOneBridge : INyxRelayDayOneBridge
         !string.IsNullOrWhiteSpace(conversationType) &&
         string.Equals(conversationType.Trim(), DeviceConversationType, StringComparison.OrdinalIgnoreCase);
 
+    // Fail-closed: only an explicit "private" maps to p2p. Missing or unrecognized types fall
+    // back to group so the known-command path returns a private-chat restriction instead of
+    // silently executing Day One side effects on an unexpected surface.
     private static string MapChatType(string? conversationType)
     {
         if (string.IsNullOrWhiteSpace(conversationType))
-            return PrivateChatType;
+            return GroupChatType;
 
         return conversationType.Trim().ToLowerInvariant() switch
         {
             "private" => PrivateChatType,
             "group" or "channel" => GroupChatType,
-            _ => PrivateChatType,
+            _ => GroupChatType,
         };
     }
 
--- a/agents/Aevatar.GAgents.NyxidChat/NyxIdChatEndpoints.cs
+++ b/agents/Aevatar.GAgents.NyxidChat/NyxIdChatEndpoints.cs
@@ -7,6 +7,7 @@ using Aevatar.AI.Abstractions.LLMProviders;
 using Aevatar.AI.ToolProviders.NyxId;
 using Aevatar.Foundation.Abstractions;
 using Aevatar.Foundation.Abstractions.Streaming;
+using Aevatar.GAgents.NyxidChat.Relay;
 using Google.Protobuf.WellKnownTypes;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Http;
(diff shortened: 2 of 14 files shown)


agents/Aevatar.GAgents.ChannelRuntime/NyxRelayDayOneBridge.cs
```diff
@@ -0,0 +1,108 @@
...
+        }
+    }
+
+    private static bool IsDevice(string? conversationType) =>
+        !string.IsNullOrWhiteSpace(conversationType) &&
+        string.Equals(conversationType.Trim(), DeviceConversationType, StringComparison.OrdinalIgnoreCase);
+
+    private static string MapChatType(string? conversationType)
+    {
+        if (string.IsNullOrWhiteSpace(conversationType))
```
eanzhao, line 84: This maps a missing or future `conversation.type` to `p2p`, and `ShouldHandle()` also accepts those requests. That makes the private-chat restriction fail open: any relay payload that omits `type` (we already have tests using that shape) or introduces a new non-DM type will be executed as a private Day One command instead of being rejected / handed off. I think this needs to fail closed and only treat an explicit `private` as `p2p`.

agents/Aevatar.GAgents.NyxidChat/NyxIdChatEndpoints.cs
```diff
@@ -838,6 +839,54 @@ private static async Task<IResult> HandleRelayWebhookAsync(
...
+                    ScopeId: scopeId,
+                    NyxIdAccessToken: userToken);
+
+                if (dayOneBridge.ShouldHandle(bridgeRequest))
+                {
+                    logger.LogInformation(
+                        "Relay Day One bridge owning message: platform={Platform}, conversation={ConversationId}, message_id={MessageId}",
+                        platform, conversationId, message.MessageId);
+
+                    _ = FinalizeDayOneBridgeReplyAsync(
```
eanzhao, line 893: This bridge path skips the actor/session path that was explicitly using `conversationId-messageId` for idempotent relay retries. If Nyx redelivers the same webhook, we enqueue another `HandleAsync` immediately; for `/daily` or `/social-media`, that reaches `AgentBuilderTool`, which generates a fresh actor id and API key on every invocation. Result: duplicate agents / keys from a single user command. We need some message-id scoped dedupe or receipt before firing these side effects.

eanzhao (changes requested):
Two blocking regressions stood out in this bridge path.

1. Missing or unrecognized `conversation.type` is currently treated as `p2p`, so the private-chat guard becomes fail-open.
2. The bridge bypasses the existing message-scoped idempotent retry/session path, so a retried Nyx webhook can execute `AgentBuilderTool` twice and create duplicate agents/API keys.

I left both points inline.

kaihuei0114:
Addressed both review points in 2960f2fd.

**1. Fail-closed chat-type mapping** — `MapChatType` now only maps the explicit `"private"` to `p2p`. Missing, empty, or unrecognized `conversation.type` values fall back to `group`, so the known-command branch returns a private-chat restriction instead of executing tool side effects. Added a `Theory` covering `null` / `""` / `"unknown-future-type"` in `NyxRelayDayOneBridgeTests`.

**2. Message-id scoped dedupe** — before firing `FinalizeDayOneBridgeReplyAsync`, the endpoint calls `IEventDeduplicator.TryRecordAsync("nyxid-relay-bridge:{scope}:{messageId}")`. A retried webhook returns 202 with `"dedupe":"duplicate"` and does not re-enter `AgentBuilderTool`, so `/daily` / `/social-media` no longer double-create agents + API keys. If no `IEventDeduplicator` is registered (unlikely in production — both Local and Orleans runtimes register `MemoryCacheDeduplicator`), the bridge still works but falls back to the prior behavior. Added an endpoint-level test that fires the same callback twice using a stubbed deduplicator and a `TaskCompletionSource` sync point (no polling wait, guard passes).

agents/Aevatar.GAgents.NyxidChat/NyxIdChatEndpoints.cs
eanzhao: Re-reviewing `2960f2f`: the fail-closed mapping is fixed, but I don’t think the dedupe is actually safe under concurrent retries yet. At this call site we’re relying on `IEventDeduplicator.TryRecordAsync(...)` as if it were an atomic first-writer-wins guard, but the concrete implementation here is `MemoryCacheDeduplicator`, which does a non-atomic `TryGetValue` + `Set`. That was fine in the runtime paths because those checks happen under mailbox/grain serialization; this webhook endpoint is fully concurrent. Two simultaneous deliveries of the same callback can still both observe the key as absent and both continue into `FinalizeDayOneBridgeReplyAsync`, so duplicate agent/API-key creation is still possible. This needs an atomic boundary-level dedupe/receipt, plus a test that exercises duplicates concurrently rather than sequentially.

eanzhao:
Fixed in 283399d6.

You were right — the previous commit leaned on `IEventDeduplicator`, whose `MemoryCacheDeduplicator` does a non-atomic `TryGetValue` + `Set`. That pair is safe inside a grain mailbox but not at a concurrent HTTP boundary.

**Replaced with an explicit atomic boundary guard** — `INyxRelayBridgeIdempotencyGuard` / `NyxRelayBridgeIdempotencyGuard` in `Aevatar.GAgents.NyxidChat.Relay`, backed by `ConcurrentDictionary.TryAdd` / `TryUpdate`:

- `TryAdd` is atomic: of N concurrent callers, exactly one sees the insert succeed.
- TTL reclaim uses `TryUpdate` keyed on the previously observed value, so an expired entry also cannot be claimed twice under a race.
- Registered in `AddNyxIdChat` as a singleton; consumed by `HandleRelayWebhookAsync` in place of `IEventDeduplicator`.

**Tests now exercise concurrency:**
- `NyxRelayBridgeIdempotencyGuardTests` — direct guard coverage including a 32-way concurrent race through a `ManualResetEventSlim` release gate; asserts exactly one winner. Also covers TTL expiry via an injected `TimeProvider`.
- `HandleRelayWebhookAsync_ShouldDedupeBridgeSideEffects_UnderConcurrentRetries` — replaces the previous sequential test. Fires eight simultaneous webhook deliveries for the same `message_id`, all released together. After `Task.WhenAll`, assertions: every delivery enters `ShouldHandle`, exactly one fires `HandleAsync` (side-effects), seven responses contain `"dedupe":"duplicate"`.

Cross-node replay is out of scope for this per-process bridge; doc-commented that explicitly. Guards + test stability guard still pass.

agents/Aevatar.GAgents.NyxidChat/Relay/INyxRelayBridgeIdempotencyGuard.cs
```diff
@@ -0,0 +1,75 @@
...
+    /// <summary>
+    /// Atomically claim <paramref name="key"/>. Returns true only for the first caller to
+    /// observe the key as absent (or expired); every concurrent caller sees false.
+    /// </summary>
+    bool TryClaim(string key);
+}
+
+internal sealed class NyxRelayBridgeIdempotencyGuard : INyxRelayBridgeIdempotencyGuard
+{
+    private readonly ConcurrentDictionary<string, DateTimeOffset> _claims = new(StringComparer.Ordinal);
```
eanzhao, line 30: This new singleton guard never evicts old keys. `_ttl` is only used to decide whether the *same* key can be reclaimed later, but for the normal case here each `message_id` is unique and only claimed once, so `_claims` grows monotonically for the lifetime of the host. On a long-running relay process this becomes an unbounded memory leak proportional to total slash-command traffic. We need actual expiration/removal semantics here (background sweep, bounded cache with TTL eviction, or remove-on-expiry on some timed path), not just an expiry timestamp stored in the value.

eanzhao (commented):
The concurrent first-writer-wins race looks fixed in `283399d6`, but I found one remaining issue in the new guard implementation and left it inline.

eanzhao:
Fixed in 71b6c2e6.

You're right — the previous guard used `_ttl` only for reclaim detection, never for actual eviction, so every processed `message_id` lived in `_claims` forever.

**Folded an opportunistic sweep into `TryClaim`:**

- Sweep interval = `max(1s, ttl / 4)`, gated on wall-clock delta from the injected `TimeProvider` — a quiet host does no work, a busy host sweeps on whichever thread first crosses the interval.
- Sweep slot is CAS-claimed via `Interlocked.CompareExchange` so concurrent callers do not duplicate the walk.
- Removal uses `ConcurrentDictionary.TryRemove(KeyValuePair)` so a concurrent reclaim on the same key cannot be clobbered by the sweep.
- Added an internal `ClaimCount` property for diagnostics / tests (the class is already internal; `AI.Tests` sees it via `InternalsVisibleTo`).

**New test** (`TryClaim_ShouldEvictExpiredEntries_ToBoundMemory`): claim 500 keys, advance time past `ttl + sweep_interval`, issue one more claim — assert `ClaimCount == 1`. Existing concurrent-winner and reclaim-after-TTL tests still pass (4/4 guard tests, 46/46 endpoint tests).

eanzhao (approved):
Re-reviewed `71b6c2e6`. The opportunistic sweep closes the remaining unbounded-growth issue in the singleton bridge guard, and the targeted guard + concurrent endpoint tests pass locally. No further findings from me on this PR.

Issue #354, opened 2026-04-23
Fail fast when ChannelRuntime registration projections fall back to InMemory in production


eanzhao:
## Problem

`Aevatar.GAgents.ChannelRuntime/ServiceCollectionExtensions.cs` currently resolves the registration read-model provider independently from the stricter projection provider policy used by scripting and GAgentService.

For `ChannelBotRegistrationDocument` and related ChannelRuntime read models, when Elasticsearch is not effectively configured, the host silently registers `AddInMemoryDocumentProjectionStore(...)` and continues booting. The code even logs that registration data will be lost on restart.

This creates a production failure mode where:

1. Nyx relay ingress still reaches Aevatar.
2. `channel-bot-registration-store` may still hold authoritative event-sourced state.
3. The query-side `channel-bot-registrations` read model is volatile and disappears on restart.
4. `/api/channels/registrations` returns `[]`.
5. Normal inbound messages fail with `registration_not_found`, so the bot becomes completely silent.

## Why this matters

Other projection hosts already enforce an explicit provider policy:

- exactly one provider enabled
- no silent production fallback to InMemory when policy/environment forbids it
- invalid Elasticsearch config throws during startup

ChannelRuntime currently behaves differently, which makes registration durability depend on deployment drift instead of an honest boot failure.

## Expected behavior

For ChannelRuntime document projections in production/distributed environments:

- do not silently fall back to `InMemory`
- enforce the same production projection policy used by the other hosts
- fail startup clearly when Elasticsearch is required but unavailable/misconfigured

## Suggested scope

- Align `Aevatar.GAgents.ChannelRuntime/ServiceCollectionExtensions.cs` with the stricter provider-selection/policy enforcement pattern already used in scripting and GAgentService.
- Honor `Projection:Policies:DenyInMemoryDocumentReadStore` and production environment checks.
- Require an explicit valid Elasticsearch document provider configuration for production channel runtime hosts.
- Add regression tests covering:
  - production/distributed config rejecting InMemory fallback
  - invalid/missing ES endpoints failing startup/service registration honestly

## Related context

A local fix is being added separately to provide an explicit registration projection rebuild path from the authoritative `channel-bot-registration-store` actor state. That mitigates recovery, but it does not remove this production configuration hazard.

loning:
## 🤖 structural solver — propose: ChannelRuntime 接入统一投影 provider policy

### TL;DR
[...]

PR #356, opened 2026-04-23
Add relay registration repair and projection rebuild

eanzhao:
## 问题
Lark relay 入口已经能收到 NyxID webhook，但当 Aevatar 本地 `channel registration` mirror / projection 丢失时，聊天链路会直接 silent fail。与此同时，`nyxid-chat` 侧现有工具无法完整携带 Lark 注册所需字段，也缺少恢复本地 mirror 和重建 registration projection 的正式入口。

这轮 review 还补充暴露了三个风险点：
- `repair_lark_mirror` 之前会把 `CredentialRef` 写空，导致 registration 看起来恢复了，但 relay ingress 仍可能因为拿不到 per-registration HMAC 而失败
- `repair_lark_mirror` 之前不会先向 Nyx 校验 bot / api key / route，错误或过期的 Nyx 资源也可能被镜像成本地“健康” registration
- `rebuild` 入口之前会先读 query-side，再 dispatch rebuild；当 projection / document reader 本身坏掉时，手工恢复入口反而会失效

## 方案
- 为 Nyx relay / Lark 注册链补齐 `verification_token` 透传
- 新增 `repair_lark_mirror`，允许基于已有 Nyx bot / route / api key 补回 Aevatar 本地 registration mirror
- 新增 `rebuild_projection` 能力，支持手工和启动时重建 `channel registration` projection
- 暴露 `POST /api/channels/registrations/rebuild` 宿主入口
- 更新 `nyxid-chat` prompt / 配置，让它优先使用 `rebuild_projection` 和 `repair_lark_mirror` 做自救和诊断
- 为上述链路补测试

## 本轮补强
- `repair_lark_mirror` 现在会先向 Nyx 查询并校验 relay API key、channel bot、conversation route，再把确认过的远端事实写入本地 mirror
- `repair_lark_mirror` 不再写空 `CredentialRef`；现在必须保留已有 relay credential，通过显式 `credential_ref` 或复用旧 `registration_id` 对应的 `vault://.../relay-hmac` secret，找不到就 fail closed
- `ChannelBotRegistrationStoreCommands` 统一改成 `IActorRuntime` 只负责 bootstrap / lookup，实际 envelope 投递走 `IActorDispatchPort`
- HTTP / tool 两条 `rebuild_projection` 路径都改成先 dispatch authoritative rebuild，再 best-effort 读取 query-side observation，因此 query-side 不健康时也还能触发恢复
- `nyxid-chat` prompt 同步更新，repair 场景要求补 `webhook_base_url`，并在需要时显式保留 `credential_ref`

## 影响路径
- `agents/Aevatar.GAgents.ChannelRuntime`
- `agents/Aevatar.GAgents.NyxidChat`
- `src/Aevatar.AI.ToolProviders.NyxId`
- `test/Aevatar.GAgents.ChannelRuntime.Tests`
- `test/Aevatar.AI.Tests`

## 验证

结果：
- Channel runtime tests: 256 passed
- Stability guard: passed

## 后续
- #354 Fail fast when ChannelRuntime registration projections fall back to InMemory in production
- #355 Do not start ChannelRuntime in degraded mode when registration projection activation or rebuild fails


agents/Aevatar.GAgents.ChannelRuntime/NyxLarkProvisioningService.cs
```diff
@@ -198,6 +224,63 @@ await RegisterLocalMirrorAsync(
...
+
+        try
+        {
+            await RegisterLocalMirrorAsync(
+                registrationId,
+                nyxProviderSlug,
+                webhookUrl,
+                request.ScopeId?.Trim() ?? string.Empty,
+                request.NyxAgentApiKeyId.Trim(),
+                string.Empty,
```
eanzhao: `repair_lark_mirror` rebuilds the local registration with `CredentialRef = string.Empty`. That makes the repaired registration impossible to resolve in `NyxIdRelayRegistrationCredentialResolver`, so `NyxIdRelayAuthValidator` cannot recover the per-registration relay HMAC and webhook signature validation falls back to global config or fails outright. In other words, this path can make `list` look healthy while relay ingress still stays broken.

eanzhao (commented):
I found two blocking issues in the new repair/rebuild path.

agents/Aevatar.GAgents.ChannelRuntime/ChannelBotRegistrationStoreCommands.cs
```diff
@@ -0,0 +1,67 @@
...
+            {
+                RegistrationId = registrationId ?? string.Empty,
+            },
+            ct);
+
+    private static async Task DispatchAsync<TCommand>(
+        IActorRuntime actorRuntime,
+        TCommand command,
+        CancellationToken ct)
+        where TCommand : class, IMessage
```
eanzhao, line 55: This helper now hard-couples command dispatch to `IActorRuntime.Get/Create + IActor.HandleEventAsync`. That mixes lifecycle lookup with delivery and bypasses the repo's `IActorDispatchPort` abstraction, so the new rebuild/delete path is no longer transport-neutral. Given the architecture rules in this repo, this should go through a dispatch port (runtime for lookup/bootstrap only) instead of calling the actor directly.

eanzhao (commented):
One more issue in the repair flow.

agents/Aevatar.GAgents.ChannelRuntime/NyxLarkProvisioningService.cs
```diff
@@ -198,6 +224,63 @@ await RegisterLocalMirrorAsync(
...
+        var registrationId = string.IsNullOrWhiteSpace(request.RequestedRegistrationId)
+            ? Guid.NewGuid().ToString("N")
+            : request.RequestedRegistrationId.Trim();
+        var nyxBaseUrl = _nyxOptions.BaseUrl.TrimEnd('/');
+        var nyxProviderSlug = string.IsNullOrWhiteSpace(request.NyxProviderSlug)
+            ? DefaultNyxProviderSlug
+            : request.NyxProviderSlug.Trim();
+        var webhookUrl = $"{nyxBaseUrl}/api/v1/webhooks/channel/lark/{Uri.EscapeDataString(request.NyxChannelBotId.Trim())}";
+
+        try
```
eanzhao, line 258: This `repair` flow never verifies anything against Nyx: it derives `webhookUrl` locally and then persists the caller-provided bot/api-key/route IDs straight into the actor. That means a typo, stale route, or mismatched bot/api-key pair produces an apparently healthy local mirror even though the authoritative Nyx resources do not match. Since the PR description says this repairs from existing Nyx resources, the implementation should first query Nyx with the access token and only mirror confirmed remote facts.

agents/Aevatar.GAgents.ChannelRuntime/NyxLarkProvisioningService.cs
eanzhao: Fixed in `41f97772`. `repair_lark_mirror` no longer writes an empty `CredentialRef`: repair now preserves an existing relay credential via explicit `credential_ref` or the reused `registration_id`'s derived `vault://.../relay-hmac` ref, and it fails closed if that secret cannot be recovered. Added coverage in `NyxLarkProvisioningServiceTests`.

agents/Aevatar.GAgents.ChannelRuntime/ChannelBotRegistrationStoreCommands.cs
eanzhao: Fixed in `41f97772`. `ChannelBotRegistrationStoreCommands` now bootstraps the well-known store actor via `IActorRuntime` only, then dispatches envelopes through `IActorDispatchPort`. I updated startup/rebuild/delete/register call sites and tests to keep the path transport-neutral.

agents/Aevatar.GAgents.ChannelRuntime/NyxLarkProvisioningService.cs
eanzhao: Fixed in `41f97772`. `repair_lark_mirror` now verifies Nyx state before mirroring: it queries the relay API key, channel bot, and route, validates callback/bot/api-key consistency, and only persists confirmed remote facts into the local mirror.

agents/Aevatar.GAgents.ChannelRuntime/NyxLarkProvisioningService.cs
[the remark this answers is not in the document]
eanzhao: Addressed in `41f97772`. Repair now preserves a live relay credential reference and refuses to report success when the per-registration relay secret cannot be recovered, so ingress auth cannot be left silently broken after mirror repair.

agents/Aevatar.GAgents.ChannelRuntime/ChannelCallbackEndpoints.cs
```diff
@@ -163,6 +181,26 @@ private static async Task<IResult> HandleListRegistrationsAsync(
         return Results.Ok(result);
     }
 
+    private static async Task<IResult> HandleRebuildRegistrationsAsync(
+        [FromServices] IActorRuntime actorRuntime,
+        [FromServices] IChannelBotRegistrationQueryPort queryPort,
+        CancellationToken ct)
+    {
+        var registrations = await queryPort.QueryAllAsync(ct);
```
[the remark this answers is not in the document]
eanzhao: Addressed in `41f97772`. The rebuild HTTP/tool paths now dispatch the authoritative rebuild command first and treat query-side observation as best-effort, so operators can still trigger recovery when the projection reader is degraded.

agents/Aevatar.GAgents.ChannelRuntime/NyxLarkProvisioningService.cs
```diff
@@ -304,25 +419,261 @@ private async Task RegisterLocalMirrorAsync(
...
+        var listResponse = await _nyxClient.ListConversationRoutesAsync(accessToken, expectedChannelBotId, ct);
+        if (LooksLikeErrorEnvelope(listResponse))
+            throw new InvalidOperationException($"channel_route_list_failed {ExtractErrorDetail(listResponse)}");
+
+        var matches = ParseConversationRoutes(listResponse)
+            .Where(route =>
+                string.Equals(route.ChannelBotId, expectedChannelBotId, StringComparison.Ordinal) &&
+                string.Equals(route.AgentApiKeyId, expectedApiKeyId, StringComparison.Ordinal))
+            .ToList();
+        if (matches.Count == 0)
```
eanzhao, line 521: This still treats "no matching route exists in Nyx" as a successful repair path by returning `null` and later persisting an empty `NyxConversationRouteId`. The PR description, prompt text, and the normal provisioning flow all treat the bot+route+api-key triple as the supported relay setup. If the route is missing, `repair_lark_mirror` currently returns `accepted` even though Nyx routing is still incomplete and the relay can stay silent. This should fail instead of green-lighting the repair.

eanzhao (commented):
I re-checked the repair hardening. The earlier credential-ref and dispatch-port issues are fixed, but there is still one behavior gap.

agents/Aevatar.GAgents.ChannelRuntime/NyxLarkProvisioningService.cs
eanzhao: Fixed in `389e6232`. `repair_lark_mirror` no longer treats a missing Nyx conversation route as a successful repair: `ResolveConfirmedConversationRouteAsync(...)` now fails with `missing_matching_nyx_conversation_route`, and the repaired mirror only persists a confirmed non-empty `NyxConversationRouteId`. I also added coverage in `NyxLarkProvisioningServiceTests` for the no-route case.

eanzhao (commented):
I re-reviewed the latest fix commit. The missing-route false positive is fixed, but there are still two recovery-path issues in `repair_lark_mirror`'s tool entry that can block or misreport repair in degraded states.

agents/Aevatar.GAgents.ChannelRuntime/ChannelRegistrationTool.cs
```diff
@@ -153,33 +226,144 @@ private async Task<string> RegisterLarkViaNyxAsync(
...
+        var nyxConversationRouteId = GetStr(args, "nyx_conversation_route_id")?.Trim() ?? string.Empty;
+        if (string.IsNullOrWhiteSpace(nyxChannelBotId))
+            return """{"error":"'nyx_channel_bot_id' is required for repair_lark_mirror"}""";
+        if (string.IsNullOrWhiteSpace(nyxAgentApiKeyId))
+            return """{"error":"'nyx_agent_api_key_id' is required for repair_lark_mirror"}""";
+
+        var queryPort = _serviceProvider.GetService<IChannelBotRegistrationQueryPort>();
+        if (queryPort is not null)
+        {
+            var registrations = await queryPort.QueryAllAsync(ct);
```
eanzhao: `repair_lark_mirror` is supposed to recover when the local mirror/read model is missing or unhealthy, but this unconditional `queryPort.QueryAllAsync(...)` makes the tool depend on query-side health before it even calls `INyxLarkProvisioningService.RepairLocalMirrorAsync(...)`. If the registration projection or document reader is degraded, this throws and the repair path becomes unusable in exactly the broken state it is meant to fix. This preflight check should be best-effort at most, or skipped entirely.

agents/Aevatar.GAgents.ChannelRuntime/ChannelRegistrationTool.cs
```diff
@@ -153,33 +226,144 @@ private async Task<string> RegisterLarkViaNyxAsync(
...
+        string nyxChannelBotId,
+        string nyxAgentApiKeyId,
+        string nyxConversationRouteId)
+    {
+        static bool Match(string left, string right) =>
+            !string.IsNullOrWhiteSpace(left) &&
+            !string.IsNullOrWhiteSpace(right) &&
+            string.Equals(left, right, StringComparison.Ordinal);
+
+        return Match(entry.NyxAgentApiKeyId, nyxAgentApiKeyId) ||
```
eanzhao: The `already_registered` short-circuit only requires one of `NyxAgentApiKeyId`, `NyxChannelBotId`, or `NyxConversationRouteId` to match. That means a stale local mirror with the same bot ID but a different relay API key / route is treated as "matching" and repair never runs. Since the returned payload says "Matching local Aevatar mirror already exists", this needs stricter identity matching (at least all supplied non-empty IDs) or an authoritative verification path instead of `||`.

eanzhao (commented):
I cross-checked the actual NyxID implementation under `~/Code/NyxID`. That surfaced one more contract mismatch in the route lookup path.

agents/Aevatar.GAgents.ChannelRuntime/NyxLarkProvisioningService.cs
```diff
@@ -304,25 +419,261 @@ private async Task RegisterLocalMirrorAsync(
...
+        }
+    }
+
+    private static IReadOnlyList<ConfirmedConversationRoute> ParseConversationRoutes(string response)
+    {
+        try
+        {
+            using var document = JsonDocument.Parse(response);
+            var routes = new List<ConfirmedConversationRoute>();
+            foreach (var item in EnumerateObjects(document.RootElement, "routes", "channel_conversations", "items", "data"))
```
eanzhao: I checked the real NyxID handler in `~/Code/NyxID/backend/src/handlers/channel_conversations.rs`: `GET /api/v1/channel-conversations` returns `{ "conversations": [...] }`, not `routes` / `channel_conversations` / `items` / `data`. This parser never looks at `conversations`, so the optional route-discovery path in `repair_lark_mirror` will parse an empty list against the real NyxID API and fail with `missing_matching_nyx_conversation_route` even when a default route exists. If `nyx_conversation_route_id` is meant to stay optional, this needs to accept NyxID's actual `conversations` field.

test/Aevatar.GAgents.ChannelRuntime.Tests/NyxLarkProvisioningServiceTests.cs
```diff
@@ -170,6 +180,191 @@ public async Task ProvisionAsync_ShouldRollbackRemoteResources_WhenLocalMirrorRe
...
+            .DispatchAsync(default!, default!, default);
+    }
+
+    [Fact]
+    public async Task RepairLocalMirrorAsync_ShouldReject_WhenNoMatchingConversationRouteExistsInNyx()
+    {
+        var handler = new RecordingHandler();
+        handler.Enqueue(HttpMethod.Get, "/api/v1/api-keys/key-123", """{"id":"key-123","callback_url":"https://aevatar.example.com/api/webhooks/nyxid-relay"}""");
+        handler.Enqueue(HttpMethod.Get, "/api/v1/channel-bots/bot-456", """{"id":"bot-456","platform":"lark","webhook_url":"https://nyx.example.com/api/v1/webhooks/channel/lark/bot-456"}""");
+        handler.Enqueue(HttpMethod.Get, "/api/v1/channel-conversations", """{"routes":[]}""");
```
eanzhao: This test shape no longer matches the real NyxID API. `~/Code/NyxID/backend/src/handlers/channel_conversations.rs` returns `{"conversations": [...]}` for `GET /api/v1/channel-conversations`, so keeping the mock as `{"routes":[]}` hides the parser mismatch in `ParseConversationRoutes(...)`. Please add coverage with the real NyxID list payload shape, including a positive case where `nyx_conversation_route_id` is omitted and the default route is discovered from `conversations`.

agents/Aevatar.GAgents.ChannelRuntime/ChannelRegistrationTool.cs
eanzhao: Fixed in `5fa30b64`. `repair_lark_mirror` now treats the local query-side preflight as best-effort: `QueryAllAsync(...)` is wrapped so projection/document-reader failures no longer block `INyxLarkProvisioningService.RepairLocalMirrorAsync(...)`. I added `ChannelRegistrationToolTests.ExecuteAsync_RepairLarkMirror_StillRepairsWhenQuerySideIsUnavailable` to cover the degraded read-model case.

agents/Aevatar.GAgents.ChannelRuntime/ChannelRegistrationTool.cs
eanzhao: Fixed in `5fa30b64`. The `already_registered` short-circuit now requires every supplied non-empty Nyx identity field to match, instead of accepting any single `bot/api-key/route` match. That prevents stale partial mirrors from suppressing repair. Covered in `ChannelRegistrationToolTests.ExecuteAsync_RepairLarkMirror_DoesNotShortCircuitOnPartialNyxIdentityMatch`.

agents/Aevatar.GAgents.ChannelRuntime/NyxLarkProvisioningService.cs
eanzhao: Fixed in `5b1fe974`. `ParseConversationRoutes(...)` now accepts NyxID's real `{"conversations": [...]}` list shape in addition to the older aliases, so omitted-route repair can discover the default conversation route from the live Nyx API response.

test/Aevatar.GAgents.ChannelRuntime.Tests/NyxLarkProvisioningServiceTests.cs
eanzhao: Fixed in `5b1fe974`. The tests now use the real NyxID list payload shape (`{"conversations": [...]}`), update the negative no-route case to `{"conversations":[]}`, and add a positive omitted-route case that discovers the default route from the conversation list.

PR #382, opened 2026-04-24
Improve Mission Control discovery

### LINKED ISSUE #318: Make Mission Control and Event Stream discoverable from the main operator flow
## Background
`/runtime/runs` and `/runtime/mission-control` are real features, but they are still hidden routes. The product has live-ops capability, but discoverability is weak.

## Goal
Promote `Mission Control` and `Event Stream` into the obvious operator path instead of leaving them as hidden tools.

## Scope
- Add contextual entry points from Team Detail and/or Runs
- Add an obvious "open in Mission Control" path from a selected live run or active team context
- Review whether Event Stream should remain hidden in navigation or become a clear operator destination
- Keep this as frontend-only discoverability and navigation work

## Acceptance
- A user can reach Mission Control from an existing operator flow without manually editing the URL
- Event Stream becomes easier to find from the main runtime workflow
- Navigation stays semantically aligned with `Teams / Platform / Settings`

## Notes
Relevant code:
- `apps/aevatar-console-web/config/routes.ts`
- `apps/aevatar-console-web/src/pages/MissionControl/`
- `apps/aevatar-console-web/src/pages/runs/`
- `apps/aevatar-console-web/src/pages/teams/`

potter-sun:
## Summary
- Promote Event Stream into the Platform navigation as the main runtime operator entry.
- Add Mission Control deep-link helpers and contextual entry points from Team Detail and active Event Stream runs.
- Add Mission Control recovery actions back to Event Stream or Team context when opened without a live run.

## Validation
- npm --prefix apps/aevatar-console-web test -- --runTestsByPath src/shared/navigation/runtimeRoutes.test.ts src/routesConfig.test.ts src/pages/runs/index.test.tsx src/pages/teams/detail.test.tsx --runInBand
- npm --prefix apps/aevatar-console-web run tsc

## Related Issue
Closes #318

### DIFF
--- a/apps/aevatar-console-web/src/pages/runs/index.tsx
+++ b/apps/aevatar-console-web/src/pages/runs/index.tsx
@@ -13,6 +13,7 @@ import {
 import {
   AppstoreOutlined,
   ArrowLeftOutlined,
+  ControlOutlined,
   DeploymentUnitOutlined,
   InfoCircleOutlined,
   SendOutlined,
@@ -27,6 +28,7 @@ import { sanitizeReturnTo } from "@/shared/auth/session";
 import { buildTeamDetailHref } from "@/shared/navigation/teamRoutes";
 import {
   buildRuntimeExplorerHref,
+  buildRuntimeMissionControlHref,
   buildRuntimeWorkflowsHref,
 } from "@/shared/navigation/runtimeRoutes";
 import {
@@ -1203,6 +1205,30 @@ const RunsPage: React.FC = () => {
   ]);
   const actorId = session.context?.actorId;
   const commandId = session.context?.commandId ?? "";
+  const canOpenMissionControl = Boolean(activeScopeId.trim() && session.runId);
+  const handleOpenMissionControl = useCallback(() => {
+    const runId = session.runId?.trim();
+    const scopeId = activeScopeId.trim();
+    if (!scopeId || !runId) {
+      return;
+    }
+
+    history.push(
+      buildRuntimeMissionControlHref({
+        actorId: actorId || undefined,
+        endpointId: activeEndpointId || undefined,
+        runId,
+        scopeId,
+        serviceId: activeServiceOverrideId || undefined,
+      }),
+    );
+  }, [
+    activeEndpointId,
+    activeScopeId,
+    activeServiceOverrideId,
+    actorId,
+    session.runId,
+  ]);
   const payloadTypeUrl =
     activePayloadTypeUrl.trim() ||
     composerFormRef.current?.getFieldValue("payloadTypeUrl")?.trim?.() ||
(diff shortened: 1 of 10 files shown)


apps/aevatar-console-web/src/pages/runs/index.tsx
```diff
@@ -1203,6 +1205,30 @@ const RunsPage: React.FC = () => {
...
+  const canOpenMissionControl = Boolean(activeScopeId.trim() && session.runId);
+  const handleOpenMissionControl = useCallback(() => {
+    const runId = session.runId?.trim();
+    const scopeId = activeScopeId.trim();
+    if (!scopeId || !runId) {
+      return;
+    }
+
+    history.push(
+      buildRuntimeMissionControlHref({
```
AbigailDeng, line 1219: This deep link drops the active prompt, so Mission Control never enables live streaming when you open it from Runs. `useMissionControlRuntime` only sets `streamEnabled` when `routeContext.prompt` is present; with the URL built here the page falls back to polling-only mode, which means an active Event Stream run stops updating in real time as soon as the operator switches to Mission Control. Please carry the current prompt (and ideally the `autoStream` intent) into the href.

apps/aevatar-console-web/src/pages/teams/detail.tsx
```diff
@@ -1485,6 +1486,38 @@ const TeamDetailPage: React.FC = () => {
...
+    }
+
+    const actorId =
+      trimText(lens.currentRun?.actorId) ||
+      trimText(lens.playback.rootActorId) ||
+      trimText(lens.graph.focusActorId) ||
+      undefined;
+
+    history.push(
+      buildRuntimeMissionControlHref({
```
AbigailDeng, line 1508: This Team Detail entry point has the same regression: the Mission Control URL is built without the playback prompt, so the destination page opens in polling-only mode instead of continuing the live run stream. `handleOpenPlaybackRun` already has a reconstructed prompt (`launchPrompt || prompt || summary`); passing the same value through here would preserve the live Mission Control experience instead of degrading it to periodic refresh.

AbigailDeng (commented):
I found two functional issues in the new Mission Control entry points. Both comments are inline with suggested follow-up.

apps/aevatar-console-web/src/pages/runs/index.tsx
potter-sun: Fixed in 13b4cd94. Runs now passes the current active prompt and autoStream=true when opening Mission Control, so the destination can continue the live stream instead of falling back to polling-only mode. Added assertions in src/pages/runs/index.test.tsx for prompt and autoStream.

apps/aevatar-console-web/src/pages/teams/detail.tsx
potter-sun: Fixed in 13b4cd94. Team Detail now waits for a real run prompt/input before showing the Mission Control action, then passes that prompt plus autoStream=true in the deep link. This avoids starting Mission Control streaming with the placeholder playback summary. Covered by src/pages/teams/detail.test.tsx.

PR #383, opened 2026-04-24
Carry relay reply_token through inbox + GC stale pending LLM requests

eanzhao:
## Summary

Production logs after #380 unblocked the deferred LLM dispatch chain showed two follow-on failures keeping the Lark bot silent:

1. `Deferred LLM reply failed: code=adapter_not_found` — `NeedsLlmReplyEvent`s persisted to actor state during pre-#380 days kept replaying after every actor activation, but the user access token (`X-NyxID-User-Token`, ~15 min TTL) embedded in `Activity.TransportExtras.NyxUserAccessToken` was long expired by the time the LLM call ran, so NyxID 401'd every attempt with `token_expired`.
2. `Deferred LLM reply failed: code=reply_token_missing_or_expired` — fresh inbound turns captured the relay reply_token into `ConversationGAgent._nyxRelayReplyTokens` (in-memory), but the actor deactivated between `HandleNyxRelayInboundActivityAsync` and `HandleLlmReplyReadyAsync` (inbox round-trip + LLM streaming exceeds Orleans activation idle in this deployment). On reactivation the dict was empty and the outbound relay reply could not produce its credential.

Issue #366 §3 already flagged this risk:

> 如果确认 actor 在一次 turn 内也可能跨节点迁移，次选方案是使用抽象化分布式短期状态服务（Redis / Orleans transient grain 等）

This PR carries the relay credential through the inbox messages themselves so it survives actor deactivation, prunes stale pending requests so old credentials never burn an LLM round, and adds the diagnostic log line we needed an hour ago.

## Approach

The persisted-state invariant from #366 §4 (`reply_token` not in event store / projection / read model) still holds: the actor strips the credential from the copy passed to `PersistDomainEventAsync` and only attaches it to the inbox-bound copy. The inbox subscriber echoes it straight back inside `LlmReplyReadyEvent`, and the actor consumes it from there.

```
HandleNyxRelayInboundActivityAsync
  ├─ capture reply_token in actor-runtime dict (legacy fast path)
  ├─ runner.RunInboundAsync(activity, runtimeContext)
  │      └─ BuildLlmReplyRequest sets ReplyToken / ExpiresAtUnixMs on the event
  ├─ persistedCopy = inboxCopy.Clone(); persistedCopy.ReplyToken = ""
  ├─ PersistDomainEventAsync(persistedCopy)         ← never carries token
  └─ inbox.EnqueueAsync(inboxCopy)                  ← carries token

ChannelLlmReplyInboxRuntime.ProcessAsync
  ├─ stale gate: drop if RequestedAtUnixMs > 5 min ago
  ├─ relay gate: drop if relay turn but ReplyToken empty
  ├─ LLM call
  └─ dispatch LlmReplyReadyEvent { …, ReplyToken, ReplyTokenExpiresAtUnixMs }

HandleLlmReplyReadyAsync
  ├─ BuildNyxRelayRuntimeContextForReply: prefer evt.ReplyToken,
  │      fall back to actor-runtime dict
  ├─ log replyTokenSource={inbox-echo|actor-runtime-dict|none}
  └─ runner.RunLlmReplyAsync(..., runtimeContext)
```

## Changes

- **`conversation_events.proto`**: add transient `reply_token` + `reply_token_expires_at_unix_ms` to `NeedsLlmReplyEvent` (fields 7/8) and `LlmReplyReadyEvent` (fields 10/11). Doc comments forbid persistence; the actor strips them from the persisted copy.
- **`ChannelConversationTurnRunner.BuildLlmReplyRequest`**: copy reply_token + expires_at from `runtimeContext.NyxRelayReplyToken` onto the inbox-bound `NeedsLlmReplyEvent`.
- **`ConversationGAgent.HandleInboundActivityCoreAsync`**: clone the runner's request, blank reply_token + expires_at, persist the stripped copy, dispatch the original with token to the inbox.
- **`ConversationGAgent.HandleLlmReplyReadyAsync`**: new `BuildNyxRelayRuntimeContextForReply` prefers `evt.ReplyToken` over the in-memory dict; entry log line `replyTokenSource=...` tags which path produced the credential.
- **`ConversationGAgent.SchedulePendingLlmReplyDispatchesAsync`**: walk a snapshot of `State.PendingLlmReplyRequests`, drop entries older than `PendingLlmReplyRequestMaxAge = 5min` by emitting `ConversationContinueFailedEvent { ErrorCode = "stale_pending_request_dropped", NotRetryable }`. The state mutator removes them from `PendingLlmReplyRequests`, so a pod restart no longer re-queues hours-old messages.
- **`ChannelLlmReplyInboxRuntime.ProcessAsync`**: stale-age gate (`MaxInboxRequestAgeMs = 5min`) + relay-credential gate. Both drop with an explanatory log instead of burning an LLM round on a doomed request. Echo `request.ReplyToken` / `RequestedAtUnixMs` into the dispatched `LlmReplyReadyEvent`.

## Validation

- `dotnet test test/Aevatar.GAgents.Channel.Protocol.Tests/Aevatar.GAgents.Channel.Protocol.Tests.csproj` — 105 pass (3 new regression tests for strip-on-persist, inbox-echo preferred over dict, direct dispatch).
- `dotnet test test/Aevatar.GAgents.ChannelRuntime.Tests/Aevatar.GAgents.ChannelRuntime.Tests.csproj` — 286 pass (3 new tests for token echo, missing-token drop, stale-age drop).
- `dotnet test test/Aevatar.AI.Tests/Aevatar.AI.Tests.csproj` — 482 pass.
- `bash tools/ci/architecture_guards.sh` — pass through buf lint (local has no `buf`).
- `bash tools/ci/test_stability_guards.sh` — pass.

## Test plan

- [x] Unit + integration tests pass on dev base.
- [ ] Deploy to mainnet host. Send a Lark message. Expected new log lines:
  - `Enqueued LLM reply request to inbox: correlation=...`
  - `Processing LLM reply request: correlation=... target=...`
  - `Received LLM reply ready: correlation=... terminal=Completed replyTokenSource=inbox-echo`
  - `Completed deferred LLM reply: correlation=... sent=...`
- [ ] Confirm bot now actually replies in the Lark DM.

### DIFF
--- a/agents/Aevatar.GAgents.Channel.Runtime/Conversation/ConversationGAgent.cs
+++ b/agents/Aevatar.GAgents.Channel.Runtime/Conversation/ConversationGAgent.cs
@@ -33,6 +33,11 @@ public sealed partial class ConversationGAgent : GAgentBase<ConversationGAgentSt
     // sub-minute schedules are unreliable. The inbox dispatch happens inline via
     // IChannelLlmReplyInbox; the durable timer is reserved for retry/rehydration.
     private static readonly TimeSpan DeferredLlmDispatchRetryDelay = TimeSpan.FromSeconds(60);
+    // Pending LLM reply requests older than this are considered stale on rehydration:
+    // the user gave up, the relay reply_token (~30 min TTL) is likely already expired,
+    // and the user access token (~15 min TTL) used for the LLM call is definitely gone.
+    // Drop them rather than burn an LLM round and reply hours late.
+    private static readonly TimeSpan PendingLlmReplyRequestMaxAge = TimeSpan.FromMinutes(5);
     private readonly Dictionary<string, NyxRelayReplyTokenContext> _nyxRelayReplyTokens = new(StringComparer.Ordinal);
 
     /// <summary>
@@ -104,8 +109,15 @@ public sealed partial class ConversationGAgent : GAgentBase<ConversationGAgentSt
         var nowMs = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds();
         if (result.LlmReplyRequest is not null)
         {
-            await PersistDomainEventAsync(result.LlmReplyRequest);
-            await DispatchPendingLlmReplyAsync(result.LlmReplyRequest, CancellationToken.None);
+            // The transient inbox copy keeps reply_token + expiry so the LLM worker can
+            // echo them back inside LlmReplyReadyEvent; the persisted state copy must
+            // not carry the credential into the event store / projection / read model.
+            var inboxCopy = result.LlmReplyRequest;
+            var persistedCopy = inboxCopy.Clone();
+            persistedCopy.ReplyToken = string.Empty;
+            persistedCopy.ReplyTokenExpiresAtUnixMs = 0;
+            await PersistDomainEventAsync(persistedCopy);
+            await DispatchPendingLlmReplyAsync(inboxCopy, CancellationToken.None);
             Logger.LogInformation(
                 "Accepted inbound activity for deferred LLM reply: activity={ActivityId} conversation={Key}",
                 activity.Id,
@@ -215,10 +227,17 @@ public sealed partial class ConversationGAgent : GAgentBase<ConversationGAgentSt
             return;
         }
 
+        var runtimeContext = BuildNyxRelayRuntimeContextForReply(evt, pendingRequest?.Activity);
+        Logger.LogInformation(
+            "Received LLM reply ready: correlation={CorrelationId} terminal={TerminalState} replyTokenSource={Source}",
+            evt.CorrelationId,
+            evt.TerminalState,
+            DescribeReplyTokenSource(evt, runtimeContext));
+
         var runner = ResolveRunner();
         var result = await runner.RunLlmReplyAsync(
             evt,
-            BuildNyxRelayRuntimeContext(evt.CorrelationId, pendingRequest?.Activity ?? evt.Activity),
+            runtimeContext,
             CancellationToken.None);
 
         var nowMs = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds();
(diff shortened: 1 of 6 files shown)


agents/Aevatar.GAgents.Channel.Runtime/Conversation/ConversationGAgent.cs
```diff
@@ -104,8 +109,15 @@ private async Task HandleInboundActivityCoreAsync(
...
-            await DispatchPendingLlmReplyAsync(result.LlmReplyRequest, CancellationToken.None);
+            // The transient inbox copy keeps reply_token + expiry so the LLM worker can
+            // echo them back inside LlmReplyReadyEvent; the persisted state copy must
+            // not carry the credential into the event store / projection / read model.
+            var inboxCopy = result.LlmReplyRequest;
+            var persistedCopy = inboxCopy.Clone();
+            persistedCopy.ReplyToken = string.Empty;
+            persistedCopy.ReplyTokenExpiresAtUnixMs = 0;
+            await PersistDomainEventAsync(persistedCopy);
+            await DispatchPendingLlmReplyAsync(inboxCopy, CancellationToken.None);
```
eanzhao, line 120: [P1] Durable retry replays a tokenless relay request

The persisted `NeedsLlmReplyEvent` intentionally strips `ReplyToken`, but every recovery path that later reads `State.PendingLlmReplyRequests` now dispatches that stripped copy. If the first inbox enqueue fails, or a durable retry/rehydration runs before the original inbox copy is processed, `ChannelLlmReplyInboxRuntime` will drop it as a relay request without `ReplyToken`, so the retry path cannot actually recover the turn.

agents/Aevatar.GAgents.ChannelRuntime/ChannelLlmReplyInboxRuntime.cs
```diff
@@ -92,6 +94,32 @@ internal async Task ProcessAsync(NeedsLlmReplyEvent request)
...
+        // Relay credential gate: relay turns require a fresh reply_token to send the
+        // outbound. A relay request with no inbox-carried token (e.g., rehydrated from
+        // persisted state after a pod restart that lost the original capture) cannot
+        // be delivered, so skip the LLM call entirely.
+        if (IsRelayRequest(request) && string.IsNullOrWhiteSpace(request.ReplyToken))
+        {
+            _logger.LogWarning(
+                "Dropping relay LLM reply request without inbox-carried reply_token: correlation={CorrelationId}",
+                request.CorrelationId);
+            return;
```
eanzhao, line 123: [P2] Inbox drops leave actor pending state uncleared

Both stale and missing-token gates return without notifying `ConversationGAgent`. That means the actor keeps the `PendingLlmReplyRequests` entry until a later rehydration happens to emit `stale_pending_request_dropped`; if the actor stays active, this becomes a silent stuck pending request with no terminal event or cleanup.

eanzhao (commented):
Reviewed the relay token inbox flow. It fixes the fresh fast path, but I found two recovery/cleanup gaps that are worth addressing before relying on it in production.

eanzhao:
Both review points addressed:

**P1 — durable retry replays a tokenless relay request** (`bebaef26`)
`DispatchPendingLlmReplyAsync` now re-enriches the inbox copy from the actor's in-memory `_nyxRelayReplyTokens` dict before enqueueing, so any retry / durable-reminder / OnActivate dispatch that reads the stripped persisted `NeedsLlmReplyEvent` from `State.PendingLlmReplyRequests` still emits an inbox copy with the credential — as long as the activation is alive.

- inbound direct path: enrichment is a no-op (request already has token)
- retry after transient inbox outage, same activation: enrichment copies from the dict, inbox gate passes
- rehydration after pod restart: enrichment no-op (dict empty), inbox gate drops as intended (token would have died with the process and the 30-min NyxID TTL would have killed it anyway)

Enqueue log now tags `replyTokenSource={inbound-direct|actor-runtime-dict|none}` so the retry-enrichment path is observable. Regression test `HandleDeferredLlmReplyDispatchRequestedAsync_ReEnrichesStrippedPendingRequestWithActorRuntimeToken` covers the retry path.

**P2 — inbox drops leave actor pending state uncleared** (`e6e432b5`)
Added a dedicated `DeferredLlmReplyDroppedEvent` (proto). Each inbox-side gate now dispatches it to the actor with a distinct reason code:
- `malformed_deferred_llm_reply_request`
- `stale_inbox_request_dropped`
- `missing_relay_reply_token`

The actor's new `HandleDeferredLlmReplyDroppedAsync` handler emits a `NotRetryable` `ConversationContinueFailedEvent` carrying the same reason code, which routes through the existing `ApplyContinueFailed` state matcher branch and retires the pending entry from `State.PendingLlmReplyRequests` plus clears the runtime reply-token dict. Same retirement path as a permanent runner failure — no new state semantics.

Tests cover the actor handler in isolation (`HandleDeferredLlmReplyDroppedAsync_RetiresPendingRequestWithNotRetryableFailure`, `HandleDeferredLlmReplyDroppedAsync_IgnoresUnknownCorrelationId`) and each inbox drop path asserts the dispatched event carries the right reason code.

Local: AI 482 / Channel.Protocol 108 / ChannelRuntime 286 pass; arch + stability guards pass.

PR #368, opened 2026-04-24
Fix NyxID relay callback authentication

eanzhao:
## Summary

Fixes #366.

This replaces the NyxID relay callback HMAC/local-secret authentication path with a dedicated callback JWT verified through JWKS. The callback token validates `aud`, expiry with clock skew, `jti`, `api_key_id`, `message_id`, `platform`, and byte-exact `body_sha256` binding before the relay payload is accepted.

The relay reply token is not persisted in `ChatActivity`, actor state, read models, or outbound protobuf contracts. It is carried only on the transient actor-inbox relay envelope, captured into `ConversationGAgent` runtime memory for the current turn, and cleaned on success, failure, duplicate delivery, token expiry, or actor deactivation. Outbound delivery persists only `reply_message_id` and `correlation_id`; the old `reply_access_token` protobuf field is reserved.

## Rollout Note

This PR intentionally does not restore HMAC fallback or `IAevatarSecretsStore` usage in production paths. Aevatar should be deployed after NyxID callback JWT forwarding is available; callbacks without `X-NyxID-Callback-Token` are rejected as `callback_jwt_missing`.

## Root Cause

The previous relay HMAC flow depended on `IAevatarSecretsStore`, which stores secrets locally on each node. In multi-pod deployment, provisioning can write the relay secret on one pod while callbacks land on another pod, causing signature verification to use a missing key and fail as `invalid_signature`.

## Changes

- Add callback JWT validation for NyxID relay callbacks, including `kid` miss JWKS refresh cooldown and single-flight behavior.
- Read the raw request body for byte-exact `body_sha256` validation.
- Add `correlation_id` to the relay callback payload and require it to match JWT `jti`.
- Register the relay replay guard in production DI and reject replayed callback `jti` / `message_id` values.
- Add `callback_jwt_validation_failures_total{reason}` metrics.
- Move reply token handling into actor-owned runtime state and schedule expiry cleanup.
- Reserve old relay secret / reply-token protobuf fields (`credential_ref`, `reply_access_token`) and remove dead credential-ref plumbing.
- Remove relay HMAC credential resolver and ChannelRuntime production `IAevatarSecretsStore` registration.
- Add an architecture guard preventing service/agent DI from default-registering `IAevatarSecretsStore`.
- Update endpoint, relay transport/outbound delivery, provisioning, and regression tests.

## Validation

- `dotnet build test/Aevatar.AI.Tests/Aevatar.AI.Tests.csproj --nologo --no-restore`
[...]

PR #385, opened 2026-04-24
Fix relay scope resolution + use bot owner LLM config

eanzhao:
## Problems

Two production bugs surfaced after the JWT-based callback auth (#368) + inline dispatch (#380, #383) landed:

**1. 401 ScopeId not produced** — NyxID's relay callback JWT (per `ChronoAIProject/NyxID#504`) only carries channel-routing claims (`api_key_id`, `message_id`, `platform`, `body_sha256`, `jti`). It never emits `sub` / `scope_id` / `nameid`, so the validator's claim-based scope extraction always returned null and the relay endpoint short-circuited to 401 before the inbound message reached the conversation actor.

**2. LLM `token_expired` 401 on relay turns** — `ChannelLlmReplyInboxRuntime.BuildEffectiveMetadata` copied `Activity.TransportExtras.NyxUserAccessToken` (the inbound caller's NyxID session token, ~15 min TTL) straight into `LLMRequestMetadataKeys.NyxIdAccessToken` + `NyxIdOrgToken`. NyxID's LLM proxy validated it against the user session, so the moment the inbound user's session lapsed the deferred bot reply 401'd with `token_expired` — even though the bot owner had a valid pre-configured LLM route the whole time.

## Solution

Single thread tying both fixes together: **resolve the bot owner's Aevatar scope id from the channel bot registration by `nyx_agent_api_key_id`**, and use that scope id wherever the bot owner's identity is needed.

- New `INyxIdRelayScopeResolver` port in NyxidChat. ChannelRuntime implements it via `IChannelBotRegistrationQueryByNyxIdentityPort.GetByNyxAgentApiKeyIdAsync(...)` → `entry.ScopeId`.
- **Auth path (Fix #1)**: relay endpoint falls back from JWT-derived scope to the resolver, keeping the JWT-derived scope as first choice.
- **LLM config path (Fix #2)**: `ChannelLlmReplyInboxRuntime` resolves the bot owner's scope id the same way and loads the owner's `UserConfig` (`DefaultModel`, `PreferredLlmRoute`, `MaxToolRounds`) via `IUserConfigQueryPort.GetAsync(scopeId)` — the same store that backs nyxid-chat preferences. The inbound-user-token leak into LLM auth metadata is removed; the bot's LLM call now uses the bot owner's pre-configured route + model + the host-level `NyxIdLLMProvider` access token accessor (whatever the host is configured with), independent of any inbound session lifecycle.

## Impact

- Lark / NyxID relay callbacks without `scope_id`/`sub` claims now reach the scoped `ConversationGAgent` as long as the bot registration has the matching `NyxAgentApiKeyId` and `ScopeId`.
- Deferred relay LLM replies use the bot owner's pre-configured model + route the same way nyxid-chat does for that owner — no more `token_expired` because of inbound session TTL.
- Both new dependencies are optional ctor params on `ChannelLlmReplyInboxRuntime`, so existing test fixtures and non-relay code paths continue to work without changes.

## Validation

- `dotnet test test/Aevatar.AI.Tests/Aevatar.AI.Tests.csproj --nologo` — 483 pass
- `dotnet test test/Aevatar.GAgents.ChannelRuntime.Tests/Aevatar.GAgents.ChannelRuntime.Tests.csproj --nologo` — 290 pass (incl. `ProcessAsync_ShouldApplyBotOwnerLlmConfig_FromUserConfigQueryPort`, `ProcessAsync_ShouldNotLeakUserAccessTokenIntoLlmAuthMetadata`)
- `dotnet test test/Aevatar.GAgents.Channel.Protocol.Tests/Aevatar.GAgents.Channel.Protocol.Tests.csproj --nologo` — 108 pass
- `bash tools/ci/architecture_guards.sh` — pass through buf lint (local has no `buf`)
- `bash tools/ci/test_stability_guards.sh` — pass

## Test plan

- [x] Unit tests pass.
- [ ] Deploy to mainnet host. Send a Lark message. Expected log lines on the success path:
  - `Relay callback authentication failed: code=callback_jwt_*` does NOT appear
  - `Resolved relay callback scope id from relay scope resolver: ...` appears (since NyxID#504 doesn't emit `sub`)
  - `Applied bot owner LLM config: scopeId=... model=... route=...` appears in the LLM dispatch
  - No `Upstream LLM route 'nyxid' rejected ... token_expired` follow-up
  - `Completed deferred LLM reply: correlation=... sent=...` finishes the turn

### DIFF
--- a/agents/Aevatar.GAgents.ChannelRuntime/ChannelLlmReplyInboxRuntime.cs
+++ b/agents/Aevatar.GAgents.ChannelRuntime/ChannelLlmReplyInboxRuntime.cs
@@ -219,11 +219,24 @@ internal sealed class ChannelLlmReplyInboxRuntime :
         // Apply the bot owner's pre-configured LLM route + model. The relay callback
         // identifies the bot by api_key_id (in activity.Bot.Value); we resolve that to
         // the owner's Aevatar scope id and load the same UserConfig the owner uses
-        // when chatting through nyxid-chat themselves. This decouples the bot's LLM
-        // call from any inbound user session — the bot reply succeeds whenever the
-        // owner has a valid configuration, regardless of whether the relay callback's
-        // X-NyxID-User-Token has expired since dispatch.
+        // when chatting through nyxid-chat themselves, then pin ModelOverride /
+        // NyxIdRoutePreference / MaxToolRoundsOverride from that configuration.
         await ApplyBotOwnerLlmConfigAsync(request, metadata, ct);
+
+        // The inbound callback's X-NyxID-User-Token is the bot owner's NyxID session
+        // JWT (freshly issued by NyxID for each callback). It is the bot owner's own
+        // credential for LLM calls — the same thing that would authorize them in
+        // nyxid-chat. The short TTL (~15 min) is mitigated by the direct-enqueue
+        // dispatch (#380), the inbox-echoed token flow (#383), and the stale pending
+        // request GC, so the token is still valid when the LLM call actually fires
+        // for any non-stale request. If the downstream provider rejects it, the
+        // classifier surfaces a real user-facing error via NyxIdRelayErrorClassifier.
+        var userAccessToken = request.Activity?.TransportExtras?.NyxUserAccessToken?.Trim();
+        if (!string.IsNullOrWhiteSpace(userAccessToken))
+        {
+            metadata[LLMRequestMetadataKeys.NyxIdAccessToken] = userAccessToken;
+            metadata[LLMRequestMetadataKeys.NyxIdOrgToken] = userAccessToken;
+        }
 
         return metadata;
     }
--- a/agents/Aevatar.GAgents.ChannelRuntime/ChannelLlmReplyInboxRuntime.cs
+++ b/agents/Aevatar.GAgents.ChannelRuntime/ChannelLlmReplyInboxRuntime.cs
@@ -4,6 +4,8 @@ using Aevatar.AI.Abstractions.LLMProviders;
 using Aevatar.GAgents.Channel.Abstractions;
 using Aevatar.GAgents.Channel.Runtime;
 using Aevatar.GAgents.Channel.NyxIdRelay;
+using Aevatar.GAgents.NyxidChat;
+using Aevatar.Studio.Application.Studio.Abstractions;
 using Google.Protobuf.WellKnownTypes;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
@@ -21,7 +23,9 @@ internal sealed class ChannelLlmReplyInboxRuntime :
     private readonly IActorRuntime _actorRuntime;
     private readonly IConversationReplyGenerator _replyGenerator;
     private readonly IInteractiveReplyCollector? _interactiveReplyCollector;
-    private readonly NyxIdRelayOptions? _relayOptions;
+    private readonly Aevatar.GAgents.Channel.NyxIdRelay.NyxIdRelayOptions? _relayOptions;
+    private readonly INyxIdRelayScopeResolver? _scopeResolver;
+    private readonly IUserConfigQueryPort? _userConfigQueryPort;
     private readonly ILogger<ChannelLlmReplyInboxRuntime> _logger;
     private IAsyncDisposable? _subscription;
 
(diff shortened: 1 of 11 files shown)


agents/Aevatar.GAgents.ChannelRuntime/ChannelLlmReplyInboxRuntime.cs
```diff
@@ -202,18 +210,88 @@ internal async Task ProcessAsync(NeedsLlmReplyEvent request)
...
-        metadata[LLMRequestMetadataKeys.NyxIdAccessToken] = userAccessToken;
-        metadata[LLMRequestMetadataKeys.NyxIdOrgToken] = userAccessToken;
+        // Apply the bot owner's pre-configured LLM route + model. The relay callback
+        // identifies the bot by api_key_id (in activity.Bot.Value); we resolve that to
+        // the owner's Aevatar scope id and load the same UserConfig the owner uses
+        // when chatting through nyxid-chat themselves. This decouples the bot's LLM
+        // call from any inbound user session — the bot reply succeeds whenever the
+        // owner has a valid configuration, regardless of whether the relay callback's
+        // X-NyxID-User-Token has expired since dispatch.
+        await ApplyBotOwnerLlmConfigAsync(request, metadata, ct);
```
eanzhao, line 224: [P1] This decouples the relay LLM call from the expired inbound user token, but it does not provide any replacement credential for the NyxID LLM provider. In the host registration, `BuildNyxIdFactory` wires NyxID providers with `static () => null`, and `NyxIdLLMProvider.ResolveAccessToken` throws when `nyxid.access_token` is absent. This means relay replies will move from upstream `token_expired` to local `NyxIdAuthenticationRequiredException` unless this path also supplies a durable bot-owner or service credential before `GenerateReplyAsync` reaches the provider. Loading the owner’s model/route/tool-round config is useful, but by itself it is not enough to make the LLM call authenticate.

eanzhao (commented):
Found one blocker in the latest commit. The direction is right: relay LLM calls should not depend on the inbound user's short-lived token. But the current patch removes that token without supplying a bot-owner/service credential, so the NyxID LLM provider still cannot authenticate the request.

eanzhao:
P1 on LLM credential — addressed in `7a39ab5f`.

Walking the NyxID source (`~/Code/NyxID/backend/src/crypto/jwt.rs:318` `generate_relay_access_token` and `handlers/channel_webhooks.rs:~440`) shows `X-NyxID-User-Token` **is** the right credential, already auto-issued per callback without any bot-owner configuration:

- `sub = bot_owner_user_id` (NyxID DB's bot record `bot.user_id`)
- `scope = "openid profile email proxy"` — explicitly includes the `proxy` scope
- `relay: true` + agent-scope restrictions
- signed by NyxID with `jwt_access_ttl_secs` (default 900s)

NyxID's auth middleware (`mw/auth.rs:412`) detects `relay: true` and sets `AuthMethod::Relay`; `llm_gateway.rs:70` calls `ensure_llm_proxy_access()` which passes because `proxy` scope is present. So the token authenticates as the bot owner against the LLM proxy — no service credential, no manual configuration.

The 15-min TTL stopped being a blocker once #380 dispatches inline (no 100ms-turned-1min Reminder hop) and #383 echoes the token through the inbox envelope (survives actor deactivation). For any non-stale request the LLM call now fires within seconds of the callback, well inside the TTL. Stale pending requests (> 5min old) get GC'd by #383's inbox gate before they can trip `token_expired` at the upstream.

Re-looking at the production logs you showed earlier: the `token_expired` cases were all stale pending from the pre-#380 era; the fresh message (`e4082981`) actually failed with `reply_token_missing_or_expired` (reply path, not LLM auth). The LLM bearer path was fine for fresh turns.

So `7a39ab5f` restores the metadata writes that thread `NyxUserAccessToken` into `LLMRequestMetadataKeys.NyxIdAccessToken` + `NyxIdOrgToken` — the callback-issued relay access token authenticates the LLM call. The new `ApplyBotOwnerLlmConfigAsync` path still pins `ModelOverride` / `NyxIdRoutePreference` / `MaxToolRoundsOverride` from the bot owner's `UserConfig` so the call hits the route + model the owner picked in nyxid-chat. No new configuration, no service tokens.

Tests:
- `ProcessAsync_ShouldApplyBotOwnerLlmConfig_FromUserConfigQueryPort` — full happy path: UserConfig's model/route applied AND bot-owner session token threaded as bearer.
- `ProcessAsync_ShouldThreadBotOwnerSessionTokenAsLlmBearer` — no-resolver fallback still carries the bearer.

Local: AI 483 / Channel.Protocol 108 / ChannelRuntime 293 pass; arch + stability guards pass.

Issue #416, opened 2026-04-25
fix(lark): malformed v2 card schema causes nyx-api 502, /agents silently drops reply


eanzhao:
## Summary

Sending `/agents` to the production Lark bot results in **no reply at all** to the user. The backend logs the request, formats an interactive card, dispatches it to NyxID `POST /api/v1/channel-relay/reply` — and gets back **502 with body `error code: 502`** (Cloudflare's origin-error page). The reply token is single-use, so the request is marked `relay_reply_token_consumed` → `PermanentFailure` and no fallback text is sent.

Other commands work:
- `daily` (error path) → ✅ 200 — uses `TextContent(...)`
- normal LLM chat → ✅ 200 — text streaming reply

## Root cause

`LarkMessageComposer.Compose()` (`agents/platforms/Aevatar.GAgents.Platform.Lark/LarkMessageComposer.cs:120-137`) produces a hybrid v1/v2 card payload:

```json
{
  "schema": "2.0",
  "config": { "wide_screen_mode": true },
  "header": { "title": {...}, "template": "blue" },
  "elements": [...]            ← top-level elements
}
```

This is **not a valid Lark v2 card**. Per [Lark Open Platform v2 card spec](https://open.feishu.cn/document/feishu-cards/quick-start/getting-started-with-message-cards), v2 cards require `elements` nested under `body.elements`:

```json
{
  "schema": "2.0",
  "config": {...},
  "header": {...},
  "body": {
    "elements": [...]
  }
}
```

NyxID's lark adapter (`backend/src/services/channel_adapters/lark.rs:384-396`) transparently forwards `metadata.card` to Lark `POST /open-apis/im/v1/messages` as a JSON-string `content` field. Lark either rejects or stalls on the malformed card; NyxID's reqwest call has no explicit timeout, the origin hangs, and Cloudflare returns its standard `error code: 502` page after the idle timeout.

NyxID's existing tests (`lark.rs:1526-1547`) only cover v1 cards (no `schema` field) — v2 cards from this codebase are the first ones to actually reach Lark in production.

## Why this manifested only on `/agents`

`FormatListAgentsCard` always returns a `MessageContent` populated with `Cards` + `Actions` (interactive). All other DM commands either route through text fallbacks or LLM streaming text, so no card was actually sent until `/agents`.

| Command | Format function | Payload | Result |
|---|---|---|---|
[...]

Issue #411, opened 2026-04-25
SkillRunner daily_report GitHub proxy 403s at runtime


eanzhao:
## Problem

A scheduled `daily_report` SkillRunner can be created successfully, but at execution time every GitHub request through NyxID `api-github` proxy returns 403. On retry, the model also emitted one malformed `nyxid_proxy` call with `slug=api-github` and no `path`.

This is separate from the Lark reaction permission noise and from the Lark `open_id cross app` delivery error.

## Evidence

From `aevatar-console-backend-75fd899df8-s6mp7_aevatar-console-backend.log`:

- First run produced three GitHub tool calls, all 403:
  - `GET api-github /search/commits?q=author:eanzhao+author-date:>=2026-04-24T05:30:47Z` -> 403
  - `GET api-github /search/issues?q=author:eanzhao+updated:>=2026-04-24T05:30:47Z` -> 403
  - `GET api-github /search/issues?q=commenter:eanzhao+updated:>=2026-04-24T05:30:47Z` -> 403
- Retry produced:
  - `[nyxid_proxy] Missing path. slug=api-github, raw={\"slug\":\"api-github\",\"method\":\"GET\"}`
  - `GET api-github /users/eanzhao/events/public` -> 403
  - `GET api-github /users/eanzhao` -> 403

## Relevant code path

- `AgentBuilderTemplates.TryBuildDailyReportSpec` tells the runner to use `nyxid_proxy` with service slug `api-github` and suggests the search endpoints.
- `AgentBuilderTool.CreateDailyReportAgentAsync` verifies that GitHub appears connected, resolves required service ids, then creates a new NyxID API key.
- `AgentBuilderTool.BuildCreateApiKeyPayload` only sets `scopes=proxy`, `platform=generic`, and `allowed_service_ids`.
- `SkillRunnerGAgent.BuildExecutionMetadata` passes the generated API key as `NyxIdAccessToken` during scheduled execution.
- `NyxIdApiKeysTool` exposes a `bind` action using `/api/v1/api-keys/{keyId}/bindings`, but the daily report builder does not bind the generated key to the user's GitHub credential, and it does not preflight GitHub proxy access with the generated key.
- `NyxIdApiClient.SendAsync` returns the proxy response body to the tool result, but the server log only records URL/status, so the log does not show the actual NyxID/GitHub 403 body.

## Current hypothesis

[...]

PR #412, opened 2026-04-25
fix(channel-runtime): chat_id-first outbound + fallback retry + consumed-token PermanentFailure + #411 GitHub preflight

eanzhao:
## Summary

This PR landed in three commits because review caught real production-blocking gaps each time. The current scope spans **four behavior changes plus one issue fix**, all on the SkillRunner / channel-runtime outbound delivery path. Reviewer ([4318563419](https://github.com/aevatarAI/aevatar/pull/412#issuecomment-4318563419)) called out that the prior PR body was severely stale; this rewrite documents what actually shipped.

### 1. p2p outbound: chat_id primary + persisted union_id fallback with runtime retry on `230002`

`LarkConversationTargets.BuildFromInbound` now picks `chat_id` first for ALL conversation types (was: `union_id` for p2p, `chat_id` for groups). Production showed `union_id` getting rejected with `99992364 user id cross tenant` — the relay-side ingress and `s/api-lark-bot` outbound apps live in different Lark tenants. `chat_id` is the only Lark identifier that survives both cross-app and cross-tenant boundaries when the same Lark app is on both ends of the relay.

To avoid regressing cross-app same-tenant deployments (where the outbound app is NOT a member of the inbound DM and chat_id fails with `230002 bot not in chat`), the new `BuildFromInboundWithFallback` returns `(primary, optional fallback)`. Fallback is captured ONLY for p2p with a chat_id primary AND a union_id surfaced at ingress; groups skip the fallback (chat_id is tenant-scoped — either the outbound app is in the group or no user-id helps).

**New runtime retry**: `SkillRunnerGAgent.SendOutputAsync` and `FeishuCardHumanInteractionPort.SendMessageAsync` try the primary, on Lark `230002` (`LarkBotErrorCodes.BotNotInChat`) retry exactly once with the fallback typed pair. Other Lark codes propagate immediately so users see actionable hints for the actual failure mode.

**Persistence**: 14 new fields across 7 proto messages (`SkillRunnerOutboundConfig`, `UserAgentCatalogEntry`, `UserAgentCatalogDocument`, `UserAgentCatalogUpsertCommand`, `WorkflowAgentState`, `InitializeWorkflowAgentCommand`, `WorkflowAgentInitializedEvent`), mirrored end-to-end through `UserAgentCatalogProjector.Materialize` + `UserAgentCatalogQueryPort.ToEntry` per the typed-field-projection-mirror lesson.

The full Lark identifier failure ladder, in case future debugging needs the table:

[...]

PR #418, opened 2026-04-25
fix(agent-builder): use UserService.id for api-key allowed_service_ids (#417)

### LINKED ISSUE #417: fix: api-key allowed_service_ids uses catalog DownstreamService.id, NyxID enforcement expects UserService.id → /daily fails with spurious github_proxy_access_denied
## Summary

`/daily` (and any other `agent_builder` flow that mints a new NyxID API key) consistently fails preflight with `github_proxy_access_denied` even when the user's GitHub OAuth connection at NyxID is healthy. The `BuildGitHubAuthorizationResponseAsync` check passes (NyxID reports the GitHub provider connected, OAuth token valid), but the very next request — `GET /api/v1/proxy/s/api-github/rate_limit` using the freshly minted api-key — returns **403** from NyxID's `ApiKeyScopeForbidden` enforcement.

The preflight added under #411 catches this 403 and surfaces it to the user with a hint to "verify the GitHub OAuth provider is connected at NyxID". That hint is **misleading** — reconnecting GitHub does not fix the issue, because the underlying cause is in our api-key payload, not in NyxID's GitHub binding.

## Root cause

`AgentBuilderTool.cs:BuildCreateApiKeyPayload` populates `allowed_service_ids` with **the wrong ID type**.

- We currently call `GET /api/v1/proxy/services` (NyxID catalog list) and read each item's `id`.
- `id` on that endpoint is **`DownstreamService.id`** — a global catalog UUID shared across all users.
- NyxID's proxy enforcement (`backend/src/handlers/proxy.rs:1030-1055`) compares the api-key's `allowed_service_ids` against **`UserService.id`** — a per-user instance UUID created when a user first connects to that catalog service.
- These two UUIDs are different. The api-key carries the catalog UUID, NyxID enforces against the user-instance UUID, the `contains()` check fails, request is denied as `ApiKeyScopeForbidden` → 403.

NyxID's data model (per [`backend/src/models/user_service.rs`](https://github.com/ChronoAIProject/NyxID/blob/main/backend/src/models/user_service.rs)):

```rust
pub struct UserService {
    pub id: String,                         // ← per-user instance UUID — what enforcement checks
    pub slug: String,                       // e.g. "api-github"
    pub catalog_service_id: Option<String>, // ← DownstreamService.id — what we currently fill
    ...
}
```

Enforcement excerpt (NyxID `proxy.rs:1030`):

```rust
if let Some(ref us_id) = pre.user_service_id      // ← UserService.id
    && !auth_user.allow_all_services
    && !auth_user.allowed_service_ids.contains(us_id)
{
    return Err(AppError::ApiKeyScopeForbidden(...)); // → 403
}
```

## Why this looked like an OAuth issue

Issue #411 added the GitHub preflight precisely to "fail fast with an actionable error rather than persisting an agent that will never produce a usable report." The hypothesis at the time was that the api-key was created with `allowed_service_ids=api-github` but the GitHub credential binding was missing, so every scheduled run would 403.

That diagnosis was wrong about *why* the 403 happens. The 403 is from `ApiKeyScopeForbidden`, not from upstream GitHub denying the call. The fix landed in #411 wraps the symptom (`403 from /rate_limit`) in a human-readable hint pointing to NyxID OAuth re-connection — which has no effect because the OAuth side is already correct. The `BuildGitHubAuthorizationResponseAsync` check upstream of the preflight is what actually catches genuine "OAuth not connected" cases, and it works.

## Affected paths

- `agents/Aevatar.GAgents.ChannelRuntime/AgentBuilderTool.cs:871-919` (`ResolveProxyServiceIdsAsync`) — reads catalog `id` from `/proxy/services` response
- `agents/Aevatar.GAgents.ChannelRuntime/AgentBuilderTool.cs:664-678` (`BuildCreateApiKeyPayload`) — passes those ids straight into `allowed_service_ids`
- `agents/Aevatar.GAgents.ChannelRuntime/AgentBuilderTool.cs:1454-1524` (`PreflightGitHubProxyAsync`) — issues `github_proxy_access_denied` with a misleading hint
- `src/Aevatar.AI.ToolProviders.NyxId/NyxIdApiClient.cs:DiscoverProxyServicesAsync` — currently hits `/api/v1/proxy/services` (catalog endpoint, wrong layer)

Affects every code path that creates an agent api-key via NyxID:

- `/daily` (`create_daily_report`) — file `AgentBuilderTool.cs:166-249`
- `/social-media` and `social-media` alias (`create_social_media`) — file `AgentBuilderTool.cs:~358`
- Any future `agent_builder` template that requires NyxID proxy services

## Proposed fix (P0)

Use NyxID's user-services endpoint to obtain `UserService.id` instead of `DownstreamService.id`.

1. Add `ListUserServicesAsync(token, ct)` to `Aevatar.AI.ToolProviders.NyxId.NyxIdApiClient` calling `GET /api/v1/user-services`. Response shape (`UserServiceResponse`, NyxID `user_services_handler.rs:90-130`):
   ```json
   [{"id":"<UserService.id>","slug":"api-github","catalog_service_id":"<catalog uuid>", "is_active": true, ...}]
   ```
2. Rewrite `ResolveProxyServiceIdsAsync` to:
   - call the user-services endpoint
   - filter rows by `slug` matching each required slug
   - require `is_active == true`
   - extract each row's `id` (which is the per-user `UserService.id`)
   - return that list as the api-key payload's `allowed_service_ids`
3. Distinguish two failure modes that today are conflated:
   - **slug not connected for this user** (no `UserService` row matches) → return a structured `error: "service_not_connected"` and a hint to connect the provider at NyxID. This is the case the preflight thought it was catching.
   - **slug exists but is `is_active == false`** → `error: "service_inactive"`.

Once `allowed_service_ids` carries the correct `UserService.id`, the api-key passes enforcement and the GitHub `/rate_limit` probe will return 200 (or whatever upstream GitHub actually says).

## Cleanup follow-up (P2, after P0 lands)

- Drop `PreflightGitHubProxyAsync` (or repurpose it) — once the api-key is minted with the right scope, the only legitimate sources of `403` from `/proxy/s/api-github` are GitHub-side issues (revoked token, scope downgrade) that `BuildGitHubAuthorizationResponseAsync` should catch one step earlier. Keeping a probe-and-revoke dance for hypothetical NyxID bindings is no longer load-bearing.
- Rephrase the user-facing error message. The current text — *"Verify the GitHub OAuth provider is connected at NyxID and that the key picks up the binding"* — is now actively misleading.

## Optional follow-up at NyxID (P1)

Could be raised as a separate issue at `ChronoAIProject/NyxID`: let api-key `allowed_service_ids` also accept catalog `DownstreamService.id` or service `slug` as aliases, so future SDK consumers don't repeat this footgun. Not required if we land the backend-side fix.

## Acceptance

- `ResolveProxyServiceIdsAsync` calls `/api/v1/user-services`, returns `UserService.id` values
- `/daily` happy path with a healthy GitHub connection at NyxID succeeds end-to-end (api-key minted, GitHub rate_limit probe 200, daily report agent persisted, first run scheduled)
- `/daily` when the user has *not* connected GitHub at NyxID returns the existing `oauth_required`/`credentials_required` error from `BuildGitHubAuthorizationResponseAsync` — never `github_proxy_access_denied` from the preflight
- Same for `/social-media`
- Unit test: given a fake `/user-services` response with two slugs, `ResolveProxyServiceIdsAsync` returns the per-user instance UUIDs, not the `catalog_service_id` values
- Unit test: missing slug surfaces `service_not_connected`, not the generic "could not parse" error

## Related

- #411 — original preflight that misdiagnosed the 403
- #416 — concurrent fix for v2 card schema causing `/agents` 502 (different root cause, same incident)

eanzhao:
## Summary

- Fixes #417: `/daily` (and any `agent_builder` flow that mints a child NyxID API key) consistently fails with `github_proxy_access_denied` even when the user's GitHub OAuth is healthy. Root cause is that we populate the new key's `allowed_service_ids` with catalog `DownstreamService.id` values, but NyxID's proxy enforcement (`proxy.rs:1030`) compares against per-user `UserService.id`. The mismatch is silently accepted at create-time and surfaces as 403 `ApiKeyScopeForbidden` on every proxy call.
- Routes service resolution through `GET /api/v1/user-services` and populates `allowed_service_ids` with each per-user `UserService.id`. When the same slug has multiple rows (mixed personal + org bindings, stale rows, etc.), prefers the most eligible one — only emits `service_inactive` / `service_org_viewer_only` errors when no eligible row exists.
- Pins `allow_all_services = false` in the api-key payload so NyxID's enforcement actually consults `allowed_service_ids` (the field defaults to `true`, which short-circuits enforcement and silently grants broad scope; see "Why this was invisible" below).
- Retains `BestEffortRevokeApiKeyAsync`: the GitHub preflight failure path still revokes the freshly minted key so retries don't accumulate orphans. Hint reworded to point at re-authorizing the GitHub provider, not at api-key bindings (the previous hint was based on the misdiagnosis that #411 fixed).

## Why this was invisible until production

Session-token-minted API keys default to `allow_all_services=true`, which short-circuits the enforcement check (NyxID `proxy.rs:1030`). A developer reproducing `POST /api-keys` + `/proxy/s/api-github/rate_limit` from a CLI never tripped the bug. The agent path mints child keys via the channel-relay delegation token; NyxID forces those children to inherit `allow_all_services=false` from the parent, which is what activates the enforcement check and makes the ID mismatch fatal. See #417 review comment for the full repro matrix.

This PR also makes the narrow scope first-class — `BuildCreateApiKeyPayload` sends `allow_all_services = false` explicitly, so the resolver's output is enforced regardless of what the parent's setting happens to be. As a side benefit this also turns on NyxID's `validate_service_ids` at create-time (`key_service.rs:183`), so a malformed `UserService.id` fails fast at `POST /api-keys` instead of silently passing through and 403'ing every later proxy call.

## Changes

- `src/Aevatar.AI.ToolProviders.NyxId/NyxIdApiClient.cs` — add `ListUserServicesAsync` (`GET /api/v1/user-services`).
- `agents/Aevatar.GAgents.ChannelRuntime/AgentBuilderTool.cs`:
  - Rewrite `ResolveProxyServiceIdsAsync` to use `/user-services` and return per-user `UserService.id`. When the same slug has duplicate rows, prefer the most eligible (`is_active && !(org && allowed != true)`) instead of freezing the first row seen. Specific `service_inactive` / `service_org_viewer_only` errors are only emitted when no eligible row exists for a slug.
  - Failure contract is now pre-formed JSON envelopes with stable `error` keys (`service_not_connected`, `service_inactive`, `service_org_viewer_only`, `user_services_unavailable`, `user_services_parse_failed`, `no_required_slugs`). Both call sites return the envelope verbatim instead of double-wrapping.
  - `BuildCreateApiKeyPayload` adds `allow_all_services = false` (per review #4175529548). `allow_all_nodes` left at the NyxID default — this flow does not restrict node routing.
  - `PreflightGitHubProxyAsync` doc comment + user-facing `hint` rewritten to point at re-authorizing the GitHub provider at NyxID, not at api-key bindings. `BestEffortRevokeApiKeyAsync` retained for the preflight failure path so retries don't accumulate orphan keys.
- `test/Aevatar.GAgents.ChannelRuntime.Tests/AgentBuilderToolTests.cs`:
  - Update all 10 stub sites from `/proxy/services?per_page=100` to `/user-services` with the new row shape (`id`, `slug`, `is_active`, `credential_source`).
  - Update `_FailsClosed_When_RequiredProxyServices_AreMissing` to assert the structured `service_not_connected` envelope instead of free-text.
  - Update `_FailsClosed_When_GithubProxyDeniedForNewKey` to pin the new hint wording AND that the api-key IS revoked (DELETE on `/api/v1/api-keys/key-403`).
  - Existing `allow_all_services` assertions flipped from `TryGetProperty(...).Should().BeFalse()` (field absent) to `GetProperty(...).GetBoolean().Should().BeFalse()` (field present and `false`).
  - Add four new tests:
    - `_FailsClosed_When_RequiredSlug_IsInactive` — `is_active: false` row → `service_inactive`.
    - `_FailsClosed_When_OrgSharedSlug_IsViewerOnly` — org-shared `allowed: false` → `service_org_viewer_only`.
    - `_AllowedServiceIds_AreUserServiceIds_NotCatalogIds` — regression pin: stub `id` distinct from `catalog_service_id`, assert the api-key payload carries the per-user `id`.
    - `_PicksEligibleRow_When_DuplicateSlugRowsExist` — duplicate slug rows with the ineligible one first; assert the resolver still picks the eligible one.

## Test plan

- [x] `dotnet test test/Aevatar.GAgents.ChannelRuntime.Tests/Aevatar.GAgents.ChannelRuntime.Tests.csproj` — 421/421 passing.
- [x] `bash tools/ci/architecture_guards.sh` — clean (only flag is missing `buf` for proto lint, unrelated).
- [ ] Production smoke after merge: `/daily alice` from a Lark DM should now succeed end-to-end with a healthy GitHub OAuth connection. With a revoked GitHub token the response should be the `github_proxy_access_denied` error pointing at re-authorization (not at api-key bindings).

## Follow-ups

- Optional: open a NyxID issue to make `allowed_service_ids` accept catalog `DownstreamService.id` or `slug` as aliases. Would prevent future SDK consumers from repeating this footgun. Not required if this fix lands.
- The malformed Lark v2 card schema causing `/agents` 502 is tracked separately at #416.

### DIFF
--- a/agents/Aevatar.GAgents.ChannelRuntime/AgentBuilderTool.cs
+++ b/agents/Aevatar.GAgents.ChannelRuntime/AgentBuilderTool.cs
@@ -226,8 +226,8 @@ public sealed class AgentBuilderTool : IAgentTool
 
         var providerSlug = (args.Str("nyx_provider_slug") ?? "api-lark-bot").Trim();
         var requiredServiceIds = await ResolveProxyServiceIdsAsync(nyxClient, token, templateSpec!.RequiredServiceSlugs, ct);
-        if (requiredServiceIds.error != null)
-            return JsonSerializer.Serialize(new { error = requiredServiceIds.error });
+        if (requiredServiceIds.errorJson != null)
+            return requiredServiceIds.errorJson;
 
         var agentId = string.IsNullOrWhiteSpace(args.Str("agent_id"))
             ? SkillRunnerDefaults.GenerateActorId()
@@ -244,21 +244,16 @@ public sealed class AgentBuilderTool : IAgentTool
         if (!TryParseApiKeyCreateResponse(createKeyResponse, out var apiKeyId, out var apiKeyValue, out var apiKeyError))
             return JsonSerializer.Serialize(new { error = apiKeyError });
 
-        // Issue #411: the new agent API key is allowed_service_ids=api-github but might lack a
-        // bound GitHub credential, in which case every scheduled run hits 401/403 from
-        // proxy/s/api-github and the user sees no useful daily report. Preflight a safe
-        // GitHub endpoint with the freshly minted key and fail the create with an actionable
-        // error rather than persisting an agent that will never produce a usable report.
-        //
-        // Reviewer (PR #412 r3141699756): on the fail-fast path the freshly created NyxID API
-        // key would be left behind — repeated `/daily` attempts that hit GitHub preflight
-        // accumulate orphan proxy keys. Best-effort revoke before returning.
+        // Issue aevatarAI/aevatar#411 / #417 follow-up: catch in-flight GitHub OAuth revocation.
+        // The earlier `BuildGitHubAuthorizationResponseAsync` check covers the "no provider token
+        // at all" case; this preflight only earns its keep when the OAuth grant was revoked or
+        // had its scopes downgraded between connect-time and create-time. The api-key itself is
+        // healthy at this point (`#417` fixed `allowed_service_ids` to carry per-user
+        // `UserService.id`s), so we do NOT revoke it on preflight failure — the failure mode is
+        // upstream GitHub denying the stored credential, not a malformed proxy scope.
         var preflight = await PreflightGitHubProxyAsync(nyxClient, apiKeyValue!, providerSlug, ct);
         if (preflight is not null)
-        {
-            await BestEffortRevokeApiKeyAsync(nyxClient, token, apiKeyId!, "github_preflight_failed", ct);
             return preflight;
-        }
 
         var actor = await actorRuntime.GetAsync(agentId)
                     ?? await actorRuntime.CreateAsync<SkillRunnerGAgent>(agentId, ct);
@@ -373,8 +368,8 @@ public sealed class AgentBuilderTool : IAgentTool
 
         var providerSlug = (args.Str("nyx_provider_slug") ?? "api-lark-bot").Trim();
         var requiredServiceIds = await ResolveProxyServiceIdsAsync(nyxClient, token, [providerSlug], ct);
-        if (requiredServiceIds.error != null)
-            return JsonSerializer.Serialize(new { error = requiredServiceIds.error });
+        if (requiredServiceIds.errorJson != null)
+            return requiredServiceIds.errorJson;
 
         var agentId = string.IsNullOrWhiteSpace(args.Str("agent_id"))
             ? WorkflowAgentDefaults.GenerateActorId()
@@ -998,55 +993,151 @@ public sealed class AgentBuilderTool : IAgentTool
         }
     }
 
-    private async Task<(IReadOnlyList<string>? value, string? error)> ResolveProxyServiceIdsAsync(
+    /// <summary>
+    /// Resolves the per-user <c>UserService.id</c> values that the new agent's API key needs in
+    /// <c>allowed_service_ids</c> to reach each required catalog slug through the NyxID proxy.
+    /// </summary>
+    /// <remarks>
+    /// <para>Issue aevatarAI/aevatar#417. The previous implementation called
+    /// <c>GET /api/v1/proxy/services</c> (the <em>catalog</em> list) and pulled out each row's
+    /// <c>id</c>, which is a <c>DownstreamService.id</c> — a global catalog UUID shared across
+    /// all users. NyxID's proxy enforcement (<c>backend/src/handlers/proxy.rs:1030</c>) checks the
+    /// API key's <c>allowed_service_ids</c> against the per-user <c>UserService.id</c>, not the
+    /// catalog id. The mismatch silently passed at <c>POST /api-keys</c> creation time, then
+    /// surfaced as <c>403 ApiKeyScopeForbidden</c> on every proxy call.</para>
+    /// <para>Why the old code looked correct in development: <c>allow_all_services=true</c>
+    /// short-circuits the enforcement check. Session-token-minted API keys default to
+    /// <c>allow_all_services=true</c>, so a developer reproducing the create-key + proxy-call
+    /// dance from a CLI never tripped the bug. The agent path mints child keys via the
+    /// channel-relay delegation token; NyxID forces those children to inherit
+    /// <c>allow_all_services=false</c> from the parent, which is when enforcement kicks in.</para>
+    /// <para>The fix: use <c>GET /api/v1/user-services</c>, which lists this user's
+    /// <c>UserService</c> instances. For each instance the response carries the per-user
+    /// <c>id</c> (what enforcement actually checks) plus <c>slug</c>, <c>is_active</c>, and a
+    /// <c>credential_source</c> envelope. We filter to active rows whose slug matches a required
+    /// slug, and skip org-shared rows the caller cannot use as a proxy target — those would later
+    /// surface as a less-actionable <c>org_role_insufficient</c> error.</para>
+    /// </remarks>
+    private async Task<(IReadOnlyList<string>? value, string? errorJson)> ResolveProxyServiceIdsAsync(
         NyxIdApiClient client,
         string token,
         IReadOnlyList<string> requiredSlugs,
         CancellationToken ct)
     {
         if (requiredSlugs.Count == 0)
-            return (null, "At least one required Nyx proxy service slug must be provided.");
+        {
+            return (null, JsonSerializer.Serialize(new
+            {
+                error = "no_required_slugs",
+                hint = "At least one required Nyx proxy service slug must be provided.",
+            }));
+        }
 
-        var response = await client.DiscoverProxyServicesAsync(token, ct);
+        var response = await client.ListUserServicesAsync(token, ct);
         if (IsErrorPayload(response))
-            return (null, "Could not discover required Nyx proxy services.");
+        {
+            return (null, JsonSerializer.Serialize(new
+            {
+                error = "user_services_unavailable",
+                hint = "Could not list connected Nyx user-services. Try again or check NyxID availability.",
+            }));
+        }
 
         try
         {
             using var doc = JsonDocument.Parse(response);
-            var serviceIdsBySlug = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+            // List response shape: { "services": [ {id, slug, is_active, credential_source: {...}}, ... ] }
+            // The catalog response also nests under "services" (and additionally "custom_services"),
+            // so reusing EnumerateProxyServiceItems is safe — but we accept *only* rows that look
+            // like UserService instances by checking presence of `slug`.
+            var resolutionsBySlug = new Dictionary<string, ServiceResolution>(StringComparer.OrdinalIgnoreCase);
             foreach (var svc in EnumerateProxyServiceItems(doc.RootElement))
             {
                 var slug = ReadString(svc, "slug");
                 if (string.IsNullOrWhiteSpace(slug))
                     continue;
 
-                var id = ReadString(svc, "id", "service_id");
-                if (!string.IsNullOrWhiteSpace(id))
-                    serviceIdsBySlug[slug] = id;
+                var id = ReadString(svc, "id");
+                if (string.IsNullOrWhiteSpace(id))
+                    continue;
+
+                var isActive = TryReadBool(svc, "is_active") ?? true;
+                var credentialSource = svc.TryGetProperty("credential_source", out var cs) ? cs : default;
+                var sourceType = credentialSource.ValueKind == JsonValueKind.Object
+                    ? ReadString(credentialSource, "type")
+                    : null;
+                var orgAllowed = credentialSource.ValueKind == JsonValueKind.Object
+                    ? TryReadBool(credentialSource, "allowed")
+                    : null;
+
+                // Surface the *first* match per slug. If the user happens to have multiple
+                // UserService rows for the same slug (legacy migration leftovers), order is
+                // server-determined; we don't try to pick "the best" one because the caller
+                // doesn't have a preference signal here.
+                if (resolutionsBySlug.ContainsKey(slug))
+                    continue;
+
+                resolutionsBySlug[slug] = new ServiceResolution(
+                    Id: id!,
+                    IsActive: isActive,
+                    CredentialSourceType: sourceType,
+                    OrgAllowed: orgAllowed);
             }
 
-            var missingSlugs = requiredSlugs
-                .Where(slug => !serviceIdsBySlug.ContainsKey(slug))
-                .Distinct(StringComparer.OrdinalIgnoreCase)
-                .ToArray();
-            if (missingSlugs.Length > 0)
+            var ids = new List<string>(requiredSlugs.Count);
+            foreach (var slug in requiredSlugs.Distinct(StringComparer.OrdinalIgnoreCase))
             {
-                return (null,
-                    $"Missing required Nyx proxy services: {string.Join(", ", missingSlugs)}. API key creation was rejected to avoid broad proxy access.");
+                if (!resolutionsBySlug.TryGetValue(slug, out var resolution))
+                {
+                    return (null, JsonSerializer.Serialize(new
+                    {
+                        error = "service_not_connected",
+                        slug,
+                        hint = $"NyxID has no connected user-service for slug `{slug}`. Connect the provider at NyxID before creating this agent.",
+                    }));
+                }
+
+                if (!resolution.IsActive)
+                {
+                    return (null, JsonSerializer.Serialize(new
+                    {
+                        error = "service_inactive",
+                        slug,
+                        hint = $"NyxID user-service for slug `{slug}` is inactive. Re-activate it at NyxID before creating this agent.",
+                    }));
+                }
+
+                if (string.Equals(resolution.CredentialSourceType, "org", StringComparison.OrdinalIgnoreCase) &&
+                    resolution.OrgAllowed != true)
+                {
+                    return (null, JsonSerializer.Serialize(new
+                    {
+                        error = "service_org_viewer_only",
+                        slug,
+                        hint = $"NyxID user-service for slug `{slug}` is shared by your org but your role does not permit using it as a proxy target. Ask an admin to widen the org role scope, or connect a personal credential.",
+                    }));
+                }
+
+                ids.Add(resolution.Id);
             }
 
-            var ids = requiredSlugs
-                .Select(slug => serviceIdsBySlug[slug])
-                .Distinct(StringComparer.Ordinal)
-                .ToArray();
-            return (ids, null);
+            return (ids.Distinct(StringComparer.Ordinal).ToArray(), null);
         }
         catch (JsonException)
         {
-            return (null, "Could not parse Nyx proxy service discovery response.");
+            return (null, JsonSerializer.Serialize(new
+            {
+                error = "user_services_parse_failed",
+                hint = "NyxID user-services response was not valid JSON.",
+            }));
         }
     }
+
+    private readonly record struct ServiceResolution(
+        string Id,
+        bool IsActive,
+        string? CredentialSourceType,
+        bool? OrgAllowed);
 
     private async Task<string?> BuildGitHubAuthorizationResponseAsync(
         NyxIdApiClient client,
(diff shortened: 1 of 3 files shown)


eanzhao:
## Review: does this fix #417, and is it architecturally sound?

**TL;DR — fix is correct, but ship with two outstanding concerns Codex already flagged plus one additional one below.**

### ✅ Does this solve #417?

Yes. Root cause is correctly identified and the fix is targeted:

- `ListUserServicesAsync` → `GET /api/v1/user-services` is the right endpoint. The returned `id` is the per-user `UserService.id` that NyxID's enforcement (`proxy.rs:1030`) actually checks against `allowed_service_ids`.
- The new regression test `_AllowedServiceIds_AreUserServiceIds_NotCatalogIds` pins exactly the bug — stubs distinct `id` vs `catalog_service_id` and asserts the api-key payload carries the per-user `id`. This is the right shape of regression pin and would have caught the original bug pre-merge.
- Splitting the failure modes into `service_not_connected` / `service_inactive` / `service_org_viewer_only` is a meaningful upgrade. The old free-text "Missing required Nyx proxy services" message conflated three distinct causes that need three distinct user actions.
- The `BuildGitHubAuthorizationResponseAsync` upstream check is left untouched, so the "OAuth never connected" path still surfaces `oauth_required` and never falls through to `github_proxy_access_denied`.

### ⚠️ Outstanding concerns (in priority order)

**1. P2 — orphan API keys on preflight failure (Codex flagged, I concur)**

The PR drops `BestEffortRevokeApiKeyAsync` and justifies it as "the api-key is healthy". That conflates *correctly scoped* with *useful*. After this PR, the failure flow is:

1. `BuildGitHubAuthorizationResponseAsync` passes (OAuth was connected at step T₀).
2. `POST /api-keys` mints a key with the right `allowed_service_ids`.
3. `PreflightGitHubProxyAsync` calls `/rate_limit`, gets 401/403 because the user revoked the OAuth grant between T₀ and T₁.
4. We return the structured error. **The api-key stays in NyxID, with no consumer.**

Every retry of `/daily` while OAuth is revoked creates another orphan key. PR #412's reviewer (r3141699756) caught this exact concern under #411 and added the revoke. The reasoning that fixed it then is still valid now — the OAuth-revocation window is precisely the case the preflight was kept for, so it's the case we should expect to hit. Either re-introduce the best-effort revoke (limited to preflight failure, not OAuth-not-connected), or note this as an accepted trade-off and confirm NyxID has a TTL on unused keys.

**2. P1 — first-match-wins for duplicate slugs (Codex flagged, I concur)**

`if (resolutionsBySlug.ContainsKey(slug)) continue;` freezes the first row seen. Stable order is server-determined. If NyxID returns `[org-shared allowed:false, personal active:true]` for the same slug (legacy migration leftovers + a new personal connect), the user gets `service_org_viewer_only` even though a usable row exists. The doc comment acknowledges this but treats it as acceptable — I'd argue it's not, because the failure surfaces an actionable-but-wrong instruction ("ask an admin to widen org role scope") when the user has *already* connected a personal credential.

Suggest preferring rows in this order: `is_active && credential_source.type == "personal"` → `is_active && org && allowed:true` → otherwise. Cheap to do, no new test surface needed beyond a single `_PrefersActivePersonalRow_OverInactiveOrgRow` case.

**3. (mine) Helper-function contract leak: `EnumerateProxyServiceItems` reused for `/user-services`**

`EnumerateProxyServiceItems` (line 1621) walks `services`, `custom_services`, `data`. Its name and original purpose are scoped to `/proxy/services` (the catalog endpoint). The PR reuses it for `/user-services` based on the coincidence that both endpoints happen to nest under `services`. The new doc-comment acknowledges this — *"reusing EnumerateProxyServiceItems is safe — but we accept *only* rows that look like UserService instances by checking presence of `slug`"* — which is exactly the kind of "safe by coincidence" coupling that breaks silently when one of the two endpoints evolves independently at NyxID. If NyxID ever adds `custom_services` (org-shared overlay?) or `data` (paginated wrapper?) under `/user-services`, this enumerator picks them up with no contract check.

This violates CLAUDE.md's "API 字段单一语义 / 一个字段只表达一个含义" applied at the helper level: one helper doing two endpoint shapes with no protocol contract.

Cheapest fix: rename to `EnumerateServiceListItems` (drop "Proxy" from the name) and inline-document that *both* endpoints are expected to use this nesting convention. Better fix: a dedicated `EnumerateUserServiceItems` that only walks `services` (per the actual `/user-services` response shape the issue body documents) so a future schema drift fails loudly. The minor duplication is preferable to silent coupling — matches CLAUDE.md "正确架构优先".

### Minor

- `var isActive = TryReadBool(svc, "is_active") ?? true;` — fail-open default. Defensible (the worst-case fallout is creating an api-key for an inactive service, which NyxID will reject at proxy time with a different error), but worth flipping to `?? false` if NyxID's contract guarantees `is_active` is always present. Match server behavior; don't guess.
- The structured error `error` field uses string sentinels. For a tool wire-output that's fine, but if any upstream consumer branches on these values they should be a typed enum / proto field per CLAUDE.md "核心语义强类型". Not a blocker for this PR — flag for follow-up if these errors get parsed anywhere downstream.

### Architecture check: clean

- Layering respected: `AgentBuilderTool` (agents/) calls `NyxIdApiClient` (src/Aevatar.AI.ToolProviders.NyxId/). No cross-layer reverse deps introduced.
- "删除优先": `BestEffortRevokeApiKeyAsync` and dead `#411` comments removed. (See concern #1 — the deletion is correct in spirit but the orphan-key consequence needs handling separately.)
- "变更必须可验证": tests pass, includes a regression pin for the exact original bug.

### Verdict

Fix #417 is correct. I'd ask for #1 (orphan revoke restoration) before merge — that one was caught and fixed once already and is regressing. #2 and #3 are improvements rather than blockers, but #2 is small enough to land in this same PR.

agents/Aevatar.GAgents.ChannelRuntime/AgentBuilderTool.cs
```diff
@@ -998,56 +993,152 @@ private static bool SupportsManagedLifecycle(string? agentType) =>
...
+                    : null;
+                var orgAllowed = credentialSource.ValueKind == JsonValueKind.Object
+                    ? TryReadBool(credentialSource, "allowed")
+                    : null;
+
+                // Surface the *first* match per slug. If the user happens to have multiple
+                // UserService rows for the same slug (legacy migration leftovers), order is
+                // server-determined; we don't try to pick "the best" one because the caller
+                // doesn't have a preference signal here.
+                if (resolutionsBySlug.ContainsKey(slug))
```
eanzhao: Blocking: the resolver should not freeze the first row for a slug and validate only that row. `/user-services` can legitimately contain multiple rows for the same catalog slug, such as an org-shared `allowed:false` row plus a personal active row, or inactive legacy rows plus a valid current row. With the current first-match behavior, an ineligible row that appears first returns `service_org_viewer_only` / `service_inactive` even though a valid `UserService.id` exists later in the payload, so `/daily` still fails for mixed bindings. Please collect candidates per slug and choose an eligible row (`is_active == true` and non-org or `credential_source.allowed == true`), only returning the specific inactive/org-viewer error when no eligible candidate exists. Add a regression where the invalid row appears before the valid row.

agents/Aevatar.GAgents.ChannelRuntime/AgentBuilderTool.cs
```diff
@@ -244,21 +244,16 @@ private async Task<string> CreateDailyReportAgentAsync(
...
+        // at all" case; this preflight only earns its keep when the OAuth grant was revoked or
+        // had its scopes downgraded between connect-time and create-time. The api-key itself is
+        // healthy at this point (`#417` fixed `allowed_service_ids` to carry per-user
+        // `UserService.id`s), so we do NOT revoke it on preflight failure — the failure mode is
+        // upstream GitHub denying the stored credential, not a malformed proxy scope.
         var preflight = await PreflightGitHubProxyAsync(nyxClient, apiKeyValue!, providerSlug, ct);
         if (preflight is not null)
-        {
-            await BestEffortRevokeApiKeyAsync(nyxClient, token, apiKeyId!, "github_preflight_failed", ct);
             return preflight;
```
eanzhao, line 262: Blocking/resource lifecycle: this branch aborts after minting the NyxID API key but before any actor/catalog state owns that key. Even if the downstream GitHub credential is the failing party, the newly minted key is now unreachable from Aevatar; after the user re-authorizes GitHub it can become a usable orphan scoped to the user services. That violates the resource ownership/cleanup rule for create flows. Keep the corrected GitHub re-authorization hint, but restore best-effort key deletion for aborts that happen after `CreateApiKeyAsync` and before actor initialization/persistence, and flip the test to assert the DELETE is attempted.

eanzhao (commented):
Core #417 direction is right: this PR switches the API-key scope resolver from the catalog list to `/user-services` and uses per-user `UserService.id` values. I found two remaining issues that can still break the flow or leave unreachable credentials behind.

agents/Aevatar.GAgents.ChannelRuntime/AgentBuilderTool.cs
eanzhao: Fixed in 94db0546.

`bestBySlug` now keeps the most eligible row per slug instead of freezing the first one — replace logic only swaps in a candidate when (a) we have no row yet, or (b) existing is ineligible AND candidate is eligible. Eligibility is encoded as `ServiceResolution.IsEligible = is_active && !(org && allowed != true)`. Specific `service_inactive` / `service_org_viewer_only` errors are only emitted when no eligible row exists for a slug.

Regression test `_PicksEligibleRow_When_DuplicateSlugRowsExist` covers both invalid-before-valid orderings:
- `api-github`: org-viewer (`allowed:false`) before personal active
- `api-lark-bot`: inactive before active

Pinned that `allowed_service_ids` carries `svc-github-personal` + `svc-lark-active`, not the first-seen ineligible ids.

agents/Aevatar.GAgents.ChannelRuntime/AgentBuilderTool.cs
eanzhao: Fixed in 94db0546.

Restored `BestEffortRevokeApiKeyAsync` and the call in the preflight failure branch. The hint stays at the corrected GitHub re-authorization wording (no regression on what #417 fixed). Comment at the call site (AgentBuilderTool.cs:252-257) explains the new rationale: each `/daily` retry mints a new api-key before preflight runs, so without revoke the user accumulates one orphan key per failed retry until they re-authorize GitHub.

Test `_FailsClosed_When_GithubProxyDeniedForNewKey` flipped to assert the DELETE on `/api/v1/api-keys/key-403` IS issued (was: assert it is NOT issued). Stub for the DELETE handler restored.

eanzhao (commented):
Blocking after checking the local NyxID source at `~/Code/NyxID` (HEAD `3279d9c95249e1ca159c66bec5754ba35281c783`): the API-key create payload still needs `allow_all_services = false` when it sends `allowed_service_ids`.

NyxID contract from source:
- `CreateApiKeyRequest.allow_all_services` defaults to `true` (`backend/src/handlers/api_keys.rs:105-106`).
- `create_key` passes `body.allowed_service_ids` and `body.allow_all_services` straight to `key_service::create_api_key` (`backend/src/handlers/api_keys.rs:1109-1112`).
- `key_service::create_api_key` validates service IDs only when `!all_svcs` (`backend/src/services/key_service.rs:177-185`).
- proxy enforcement only checks `allowed_service_ids` when `!auth_user.allow_all_services` (`backend/src/handlers/proxy.rs:1030-1033`).

So with the current Aevatar payload, the key is broad and the corrected `UserService.id` list is ignored under the current NyxID source. That means #417 is not actually pinned by the payload, and the create flow also violates the narrow-scope boundary it is trying to enforce.

Please add `allow_all_services = false` in `BuildCreateApiKeyPayload` and update the tests that currently assert the field is omitted (for example `AgentBuilderToolTests.cs` around the `TryGetProperty("allow_all_services", out _)` assertions). `allow_all_nodes` can stay at the NyxID default unless this flow is also meant to restrict node routing.

eanzhao:
Fixed in d65adf57.

Reviewer is right — verified independently against `~/Code/NyxID` HEAD `3279d9c9`:

- `backend/src/handlers/api_keys.rs:105` — `allow_all_services` is `#[serde(default = "default_true")]`, so omission = `true`.
- `backend/src/handlers/proxy.rs:1030` — enforcement only fires when `!auth_user.allow_all_services`.
- `backend/src/services/key_service.rs:179,183` — same default in the create path, and `validate_service_ids` only runs under narrow scope.

The 403 we saw in production reproduced because the channel-relay delegation token forces children to inherit `allow_all_services=false`. That's not a contract we should depend on — any future caller minting directly under a broader parent (session token, etc.) would silently bypass the narrow scope this flow is supposed to enforce, and the resolver work in this PR would be a no-op.

Changes:
- `BuildCreateApiKeyPayload` now sets `allow_all_services = false` explicitly. Side benefit: this also turns on NyxID's `validate_service_ids` at create-time (`key_service.rs:183`), so a malformed `UserService.id` fails fast at `POST /api-keys` instead of silently passing through and 403'ing every later proxy call.
- `allow_all_nodes` left at the NyxID default per your note — this flow doesn't restrict node routing.
- Tests at `AgentBuilderToolTests.cs:220` and `:1750` flipped from `TryGetProperty(...).Should().BeFalse()` (field absent) to `GetProperty("allow_all_services").GetBoolean().Should().BeFalse()` (field present and `false`).
- Resolver doc comment updated to note the narrow scope is now first-class via the payload, not contingent on parent inheritance.

421/421 tests passing.

eanzhao (commented):
Re-reviewed latest head `d65adf5755646229a6d6330c1b6692d51c0b3643` against the local NyxID source at `~/Code/NyxID` (`3279d9c95249e1ca159c66bec5754ba35281c783`). The code-level blockers are addressed:

- `BuildCreateApiKeyPayload` now sends `allow_all_services = false`, so NyxID validates and enforces the `allowed_service_ids` list instead of defaulting to broad service access.
- `ResolveProxyServiceIdsAsync` now prefers an eligible `UserService.id` when duplicate slugs are returned, and only emits inactive/org-viewer errors when no eligible row exists.
- The GitHub preflight failure path restores best-effort key deletion after the key is minted but before any actor/catalog state owns it.

Local verification: `dotnet test test/Aevatar.GAgents.ChannelRuntime.Tests/Aevatar.GAgents.ChannelRuntime.Tests.csproj --nologo` passes `421/421`.

Non-blocking cleanup before merge: the PR description is stale. It still says `BestEffortRevokeApiKeyAsync` was retired and that the preflight test asserts the key is not revoked, and the test plan still says `420/420`; latest code restored revoke and tests are `421/421`. Please update the PR body so the merge record matches the final behavior.

eanzhao (commented):
Follow-up after checking production-style logs from the deployed fix: the new code path is active (`GET /user-services` -> `POST /api-keys` -> `GET /proxy/s/api-github/rate_limit` -> `403` -> `DELETE /api-keys/{id}`), but `/daily` can still fail before actor creation.

The remaining architectural gap is that PR #418 scopes the child key to one or more resolved `UserService.id` values, but all proxy calls still go through the slug route (`/api/v1/proxy/s/api-github/...` and later `State.OutboundConfig.NyxProviderSlug`) without pinning the same `UserService.id` via NyxID's `?_nyxid_via=<user_service_id>` override.

NyxID source has that override specifically because slug resolution is not a stable identity contract:

- `backend/src/handlers/proxy.rs:319-370` / `:531-584` accept `?_nyxid_via=<user_service_id>` and then execute with that exact `UserService.id`.
- Without `_nyxid_via`, the slug route calls `resolve_proxy_target_from_user_service(..., Some(slug), None)` and resolves again at proxy time.
- `backend/src/services/user_service_service.rs:173-182` `find_by_slug` is a `find_one` over active rows with no stable sort. If multiple active rows exist for the same slug, the proxy can choose a different `UserService.id` than the one Aevatar put into `allowed_service_ids`, and NyxID then correctly returns `api_key_scope_forbidden` at `proxy.rs:1030-1034`.

So the current fix proves the create payload uses `UserService.id`, but it still does not bind later proxy execution to the same service identity. The production log's 403 could be a real GitHub OAuth 403, but the current Aevatar code collapses any proxy 401/403 into `github_proxy_access_denied` and does not log the NyxID response body, so it cannot distinguish GitHub `Bad credentials` from NyxID `api_key_scope_forbidden` / `org_role_insufficient`.

Suggested fix before treating #417 as closed:

1. Resolve service bindings as a typed slug -> `UserService.id` map, not just a flat ID list.
2. Use that map to append `_nyxid_via=<id>` on the GitHub preflight call and on every runtime proxy call made with the scoped child key. Persist the map in typed proto fields on the relevant agent config/catalog state rather than relying on slug-only routing.
3. In `PreflightGitHubProxyAsync`, parse `proxy_body` and return/log distinct NyxID-scope errors separately from downstream GitHub OAuth errors. Otherwise the user gets told to re-authorize GitHub when the actual failure is still service-id scoping.

`/agents` silence in the same logs looks separate: `channel-relay/reply` returns 502 for the interactive card payload before any agent-builder create path runs, and the relay token is single-use so the text fallback is intentionally not retried. That should be tracked outside #418 unless this PR is expected to cover the card payload issue too.

PR #421, opened 2026-04-25
fix(nyxid-proxy): inject default User-Agent so GitHub doesn't 403

### LINKED ISSUE #411: SkillRunner daily_report GitHub proxy 403s at runtime

eanzhao:
Closes #411.

## Summary

Fixes the production `/daily` 403 — both at agent-create time (preflight) and at SkillRunner runtime (every `nyxid_proxy` GitHub call).

The smoking gun came from the diagnostic logs added in #420. GitHub returned **403 with body**:

> Request forbidden by administrative rules. Please make sure your request has a User-Agent header (https://docs.github.com/en/rest/overview/resources-in-the-rest-api#user-agent-required).

Not OAuth. Not `ApiKeyScopeForbidden`. **GitHub rejects any REST request without a `User-Agent` header.**

## Why this only manifested in production

| Caller | UA sent | GitHub response |
|---|---|---|
| `nyxid proxy request ...` (Rust `reqwest`) | `reqwest/0.x.y` (auto) | ✅ 200 |
| `curl` | `curl/8.x` (auto) | ✅ 200 |
| **Production backend (.NET `HttpClient`)** | **none — `HttpClient` doesn't send one by default** | ❌ 403 |

Every CLI reproduction I did against the same NyxID account, same UserService, same api-key scope returned 200, because `reqwest` auto-injects a UA. The .NET `HttpClient` does not. NyxID's proxy transparently forwards whatever the client sends, so the missing-at-the-.NET-layer header surfaces as `403 Request forbidden by administrative rules` from GitHub at the bottom of the stack.

This is also why all of #418's fixes were correct — `allow_all_services=false`, per-user `UserService.id`s, eligible-row picker, `BestEffortRevoke` — none of those had any effect on the GitHub-side rejection.

## Why this fix closes #411

#411 describes the runtime symptom directly: `daily_report` agent gets created, then at SkillRunner execution time **every** GitHub call through `nyxid_proxy` returns 403. The runtime path is `SkillRunnerGAgent` → LLM tool calls `nyxid_proxy` → `NyxIdProxyTool.ExecuteAsync` (`src/Aevatar.AI.ToolProviders.NyxId/Tools/NyxIdProxyTool.cs:111`) → `NyxIdApiClient.ProxyRequestAsync` → NyxID → GitHub. By injecting the default UA at the `ProxyRequestAsync` boundary, every call on that runtime path now carries a UA and stops being rejected.

Mapping #411's suggested fixes to what landed:

| #411 suggestion | Status |
|---|---|
| 1. Preflight GitHub access at agent-create time | ✅ Landed in #418 (`PreflightGitHubProxyAsync` in `AgentBuilderTool.cs`) |
| 2. Surface "all GitHub calls failed at runtime" as SkillRunner failure with `LastError` | Not addressed here — root cause removed makes this a secondary concern. Can open a follow-up if anyone wants the bookkeeping anyway |
| 3. Sanitized diagnostic logging for proxy 4xx bodies | Partial — `github_proxy_access_denied` envelope now includes the upstream `proxy_body` (carries the real GitHub error), and the temporary `Diagnostic[#417]` logs from #420 captured the production root cause that drove this PR (logs removed here, the diagnostic envelope stays) |
| 4. Typed GitHub activity tool | Not addressed; orthogonal to the 403 root cause |

## Changes

- `src/Aevatar.AI.ToolProviders.NyxId/NyxIdApiClient.cs`:
  - Add `DefaultProxyUserAgent` constant (`aevatar-agent-builder`).
  - `ProxyRequestAsync` now injects this UA on every call when the caller hasn't supplied one. Caller-supplied `User-Agent` in `extraHeaders` (case-insensitive) wins.
  - Pinned at the proxy boundary so agent-builder preflight, SkillRunner runtime `nyxid_proxy` calls, and any future agent that reaches GitHub via NyxID proxy all benefit without each call site remembering to send UA.
- `agents/Aevatar.GAgents.ChannelRuntime/AgentBuilderTool.cs`:
  - Rewrite `github_proxy_access_denied` hint. The previous wording ("OAuth grant was revoked, re-authorize") was based on the #418 misdiagnosis; the actual production failure was missing UA, not revoked OAuth. New hint covers both causes.
  - Remove the four `Diagnostic[#417]` log lines added in #420. They served their purpose and don't have a long-term place on the hot path (full payload JSON including api-key prefixes).
  - Tighten the comment block at the preflight call site to reflect the broader rationale (UA, OAuth revocation, etc.).
- `test/Aevatar.AI.Tests/NyxIdApiClientCoverageTests.cs`:
  - `_ShouldInjectDefaultUserAgent_WhenCallerOmitsIt` — pins the regression
  - `_ShouldHonorCallerSuppliedUserAgent` — pins that caller override still works
- `test/Aevatar.GAgents.ChannelRuntime.Tests/AgentBuilderToolTests.cs`:
  - Soften the hint substring assertion to a case-insensitive match (per review at 4175638754) so future hint copy edits don't require synchronized assertion flips.

## Verification

- [x] `dotnet test test/Aevatar.AI.Tests/Aevatar.AI.Tests.csproj` — 12/12 passing
- [x] `dotnet test test/Aevatar.GAgents.ChannelRuntime.Tests/...` — 421/421 passing (verified at 160c4748 after the review-driven assertion fix)
- [ ] Production smoke after merge: `/daily alice` from a Lark DM should now succeed end-to-end and the resulting agent's scheduled runs should produce a real GitHub-backed report.

## Follow-up

Filed at the proxy layer too: ChronoAIProject/NyxID#514 proposes NyxID default-inject `User-Agent` for proxy targets whose upstream requires it. The contract "every NyxID proxy caller must remember to send UA" is fragile (Python `urllib`, Java `HttpURLConnection`, .NET `HttpClient` all default to no UA); the proxy is a better place to enforce GitHub's hard requirement than every SDK.

## Related

- #411 — the runtime-failure issue this closes (root cause was layered: catalog-id-vs-UserService-id → allow_all_services default → User-Agent)
- #418 — `allow_all_services=false` + per-user `UserService.id` fix (correct, but masked by the UA issue underneath)
- #420 — diagnostic logs that surfaced this; logs removed in this PR
- #416 / #419 — `/agents` 502 from malformed Lark v2 card payload (separate issue, fixed in #419)
- ChronoAIProject/NyxID#514 — upstream proposal to default-inject UA at the proxy layer

### DIFF
--- a/agents/Aevatar.GAgents.ChannelRuntime/AgentBuilderTool.cs
+++ b/agents/Aevatar.GAgents.ChannelRuntime/AgentBuilderTool.cs
@@ -1833,7 +1833,7 @@ public sealed class AgentBuilderTool : IAgentTool
                 detail = string.IsNullOrWhiteSpace(detail) ? "GitHub proxy returned 401/403 for the new agent API key." : detail,
                 http_status = status,
                 proxy_body = string.IsNullOrWhiteSpace(body) ? null : body,
-                hint = "GitHub returned 401/403 to NyxID's stored OAuth token. The token was likely revoked at GitHub or had its scopes downgraded after the provider was connected. Re-authorize the GitHub provider at NyxID, then run /daily again.",
+                hint = "GitHub returned 401/403 through the NyxID proxy. Common causes: (a) the OAuth grant for GitHub was revoked at github.com/settings/applications or its scopes were downgraded — re-authorize the GitHub provider at NyxID; (b) the request reached GitHub without a User-Agent header (NyxIdApiClient now sends a default; if you see this, check that the deployed binary includes that fix). The agent will not produce a useful daily report until proxy access succeeds.",
                 nyx_provider_slug = nyxProviderSlug,
             });
         }
--- a/agents/Aevatar.GAgents.ChannelRuntime/AgentBuilderTool.cs
+++ b/agents/Aevatar.GAgents.ChannelRuntime/AgentBuilderTool.cs
@@ -233,20 +233,10 @@ public sealed class AgentBuilderTool : IAgentTool
             ? SkillRunnerDefaults.GenerateActorId()
             : args.Str("agent_id")!.Trim();
 
-        // Diagnostic (#417 follow-up, PR for production 403 root-cause hunt): log the resolved
-        // service-id list and the exact payload we're about to send to NyxID `POST /api-keys`.
-        // The api-key request body is otherwise opaque in production; without this we can't tell
-        // whether the deployed binary actually carries the `allow_all_services=false` field, or
-        // whether `requiredServiceIds` resolved to per-user `UserService.id`s vs catalog ids.
-        var apiKeyPayload = BuildCreateApiKeyPayload(agentId, requiredServiceIds.value!);
-        _logger?.LogInformation(
-            "Diagnostic[#417]: minting agent api-key. agentId={AgentId} requiredSlugs={Slugs} resolvedServiceIds={Ids} payload={Payload}",
-            agentId,
-            string.Join(",", templateSpec!.RequiredServiceSlugs),
-            string.Join(",", requiredServiceIds.value!),
-            apiKeyPayload);
-
-        var createKeyResponse = await nyxClient.CreateApiKeyAsync(token, apiKeyPayload, ct);
+        var createKeyResponse = await nyxClient.CreateApiKeyAsync(
+            token,
+            BuildCreateApiKeyPayload(agentId, requiredServiceIds.value!),
+            ct);
 
         if (IsErrorPayload(createKeyResponse))
             return createKeyResponse;
@@ -254,30 +244,19 @@ public sealed class AgentBuilderTool : IAgentTool
         if (!TryParseApiKeyCreateResponse(createKeyResponse, out var apiKeyId, out var apiKeyValue, out var apiKeyError))
             return JsonSerializer.Serialize(new { error = apiKeyError });
 
-        _logger?.LogInformation(
-            "Diagnostic[#417]: agent api-key created. agentId={AgentId} apiKeyId={ApiKeyId}",
-            agentId,
-            apiKeyId);
-
-        // Issue aevatarAI/aevatar#411 / #417 follow-up: catch in-flight GitHub OAuth revocation.
-        // The earlier `BuildGitHubAuthorizationResponseAsync` check covers the "no provider token
-        // at all" case; this preflight only earns its keep when the OAuth grant was revoked or
-        // had its scopes downgraded between connect-time and create-time.
+        // Issue aevatarAI/aevatar#411 / #417 follow-up: catch in-flight GitHub-side issues.
+        // The earlier `BuildGitHubAuthorizationResponseAsync` check covers the "no provider
+        // token at all" case; this preflight catches misconfigurations that only surface at
+        // request time (the original case under #421 was a missing `User-Agent` header that
+        // GitHub rejects with 403; OAuth grant revocation is the other one).
         //
-        // Codex review (PR #418 r3141846175): even though the api-key itself is correctly
-        // configured under #417, we still revoke it on preflight failure. Reason: each retry of
-        // `/daily` mints a *new* api-key before the preflight runs, so without revoke the user's
-        // NyxID account accumulates one orphan proxy-scoped key per failed retry until they
-        // re-authorize GitHub. The revoke is a best-effort cleanup, not a safety claim about the
-        // key's correctness.
+        // PR #418 review r3141846175: revoke the freshly-minted key on preflight failure so
+        // each `/daily` retry doesn't leave another orphan proxy-scoped key behind in the
+        // user's NyxID account. The revoke is best-effort cleanup, not a safety claim about
+        // the key's correctness.
         var preflight = await PreflightGitHubProxyAsync(nyxClient, apiKeyValue!, providerSlug, ct);
         if (preflight is not null)
         {
-            _logger?.LogWarning(
-                "Diagnostic[#417]: preflight failed; revoking new api-key. agentId={AgentId} apiKeyId={ApiKeyId} preflightResponse={Response}",
-                agentId,
-                apiKeyId,
-                preflight);
             await BestEffortRevokeApiKeyAsync(nyxClient, token, apiKeyId!, "github_preflight_failed", ct);
             return preflight;
         }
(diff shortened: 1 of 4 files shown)


agents/Aevatar.GAgents.ChannelRuntime/AgentBuilderTool.cs
```diff
@@ -1833,7 +1800,7 @@ private static string NormalizeScopeId(string? value) =>
                 detail = string.IsNullOrWhiteSpace(detail) ? "GitHub proxy returned 401/403 for the new agent API key." : detail,
                 http_status = status,
                 proxy_body = string.IsNullOrWhiteSpace(body) ? null : body,
-                hint = "GitHub returned 401/403 to NyxID's stored OAuth token. The token was likely revoked at GitHub or had its scopes downgraded after the provider was connected. Re-authorize the GitHub provider at NyxID, then run /daily again.",
+                hint = "GitHub returned 401/403 through the NyxID proxy. Common causes: (a) the OAuth grant for GitHub was revoked at github.com/settings/applications or its scopes were downgraded — re-authorize the GitHub provider at NyxID; (b) the request reached GitHub without a User-Agent header (NyxIdApiClient now sends a default; if you see this, check that the deployed binary includes that fix). The agent will not produce a useful daily report until proxy access succeeds.",
```
eanzhao, line 1803: This changed hint currently breaks the existing ChannelRuntime test suite: `ExecuteAsync_CreateAgent_DailyReport_FailsClosed_When_GithubProxyDeniedForNewKey` still asserts the hint contains `Re-authorize`, but the new text contains lowercase `re-authorize`. I ran `dotnet test test/Aevatar.GAgents.ChannelRuntime.Tests/Aevatar.GAgents.ChannelRuntime.Tests.csproj --nologo` at `2140fcb052f6ce9c67980abd1d16e8127c8204ce`; result was `Failed: 1, Passed: 420, Total: 421`. Please update either the assertion or the message before merge.

eanzhao (commented):
I agree the root-cause hypothesis is plausible and matches the production-only shape: .NET `HttpClient` sends no default User-Agent, NyxID forwards caller headers, and NyxID's `forward_request` allowlist includes `user-agent`, so injecting the default at `NyxIdApiClient.ProxyRequestAsync` is the right boundary.

However, this PR is not merge-ready as-is: the local test command claimed in the PR body does not pass at head `2140fcb052f6ce9c67980abd1d16e8127c8204ce`.

`dotnet test test/Aevatar.GAgents.ChannelRuntime.Tests/Aevatar.GAgents.ChannelRuntime.Tests.csproj --nologo`

Result: `Failed: 1, Passed: 420, Total: 421`.

Failing test:
`AgentBuilderToolTests.ExecuteAsync_CreateAgent_DailyReport_FailsClosed_When_GithubProxyDeniedForNewKey`

The new hint says `re-authorize` (lowercase), while the existing assertion still expects `Re-authorize` (uppercase). Please update the test assertion or the message so the stated 421/421 verification is true.

eanzhao:
Fixed in 160c4748.

You're right — I miscounted in the PR body. The new hint uses lowercase `re-authorize` while the existing assertion at `AgentBuilderToolTests.cs:437` still pinned `"Re-authorize"`, so the test was failing at `2140fcb0` (420/421).

Switched the assertion to a case-insensitive substring match (`.ToLowerInvariant().Should().Contain("re-authorize")`). The intent of the assertion is "hint still mentions re-authorization for the OAuth-revocation case", not "hint matches one specific capitalization" — case-pinning adds friction for future hint copy edits without strengthening the regression coverage.

Local verification at 160c4748:
```
dotnet test test/Aevatar.GAgents.ChannelRuntime.Tests/Aevatar.GAgents.ChannelRuntime.Tests.csproj --nologo
Passed!  - Failed: 0, Passed: 421, Skipped: 0, Total: 421
```

PR #426, opened 2026-04-27
Route Create Team to Studio member creation

### LINKED ISSUE #319: Route Create Team directly to Studio member creation
## Background

The current `/teams/new` path makes `Create Team` feel like a staged team creation flow, but it does not create a persisted team or member. It collects `teamName` / `entryName`, then sends the user to Studio. Most create-team route params are intentionally not carried into Studio.

The actual creatable object today is the first Studio team member. Completing Studio `Build -> Bind -> Invoke -> Observe` means the initial member is ready; it should not be described as "team created" until real team authority exists.

## Updated direction

The main Create Team flow should no longer route through `/teams/new`.

Short-term frontend flow:

`Teams -> Create Team -> Studio create first team member`

`/teams/new` should remain only as a compatibility / saved draft recovery page for existing links.

## Scope

- Change the Teams `Create Team` CTA to open Studio directly.
- Add a typed Studio route intent, e.g. `intent=create-member`.
- When Studio opens with that intent, open or focus the existing Create member flow.
- Keep `/teams/new` working for old links and saved draft recovery.
- Update `/teams/new` copy so it no longer presents itself as the primary create flow.
- Do not add backend contracts.
- Do not auto-create drafts just because `intent=create-member` is present.
- Do not claim that a real team has been created.

## Acceptance

- Clicking `Create Team` from Teams opens Studio member creation directly.
- `/teams/new` is no longer the primary Create Team route.
- Studio treats `intent=create-member` as a typed UI intent, not a generic query bag.
- Studio does not auto-create workflow drafts from the intent alone.
- Completing `Build -> Bind -> Invoke -> Observe` is described as `Initial member ready`, not `Team created`.
- Existing saved draft resume/delete behavior on `/teams/new` remains intact.

## Related context

- Design doc: `/Users/potter/.gstack/projects/unknown/potter-fix-2026-04-24_create-team-copy-flow-framing-design-20260424-173444.md`
- Backend member-first follow-up: #325
- Related implementation area:
  - `apps/aevatar-console-web/src/pages/teams/*`
  - `apps/aevatar-console-web/src/pages/studio/index.tsx`
  - `apps/aevatar-console-web/src/shared/studio/navigation.ts`

potter-sun:
## Summary
- Route the Teams home Create Team CTA directly into Studio with a typed `create-member` intent.
- Auto-open the existing Studio create-member flow for that intent without creating a backend draft or team fact.
- Keep `/teams/new` as a hidden legacy recovery page and update copy/tests to describe initial member draft recovery.
- Collapse the console shell navigation by default for this focused Studio entry.

## Why
The previous Create Team page implied a persisted Team was being created, while the current product flow actually starts team construction by creating the first Studio member. This makes the route and copy honest without introducing backend team semantics.

Closes #319

## Validation
- `npm --prefix apps/aevatar-console-web test -- --runInBand apps/aevatar-console-web/src/pages/teams/new.test.tsx apps/aevatar-console-web/src/pages/teams/home.test.tsx apps/aevatar-console-web/src/pages/studio/index.test.tsx apps/aevatar-console-web/src/app.layout.test.ts apps/aevatar-console-web/src/shared/studio/navigation.test.ts apps/aevatar-console-web/src/routesConfig.test.ts apps/aevatar-console-web/src/shared/navigation/navigationMenuSelection.test.ts`
- `git diff --check`

## Note
`npm --prefix apps/aevatar-console-web run biome:lint` still reports existing unrelated lint debt in the console web package.


AbigailDeng (commented):
Two actionable issues stood out in this flow: scope should be preserved when entering create-member from Teams, and the menu-collapse behavior likely needs a controlled state instead of relying on `defaultCollapsed`. Leaving both inline for easier follow-up.

apps/aevatar-console-web/src/pages/teams/home.tsx
```diff
@@ -1081,7 +1081,9 @@ const TeamsHomePage: React.FC = () => {
         <Space wrap>
           <Button
             icon={<PlusOutlined />}
-            onClick={() => history.push(buildTeamCreateHref())}
+            onClick={() =>
+              history.push(buildStudioRoute({ tab: "studio", intent: "create-member" }))
```
AbigailDeng: This CTA now deep-links straight into `/studio?intent=create-member`, but it no longer forwards the current `scopeId`. `StudioPage` falls back to `appContext/authSession` when `routeState.scopeId` is empty, so if the user is currently viewing a non-default scope here, the first member draft can be created under a different scope than the one they launched from. Please pass the current `scopeId` into `buildStudioRoute(...)` so the create flow stays scoped to the team context the user came from.

apps/aevatar-console-web/src/app.tsx
```diff
@@ -814,6 +822,10 @@ export const layout = ({
       overflow: "hidden",
       padding: 0,
     },
+    defaultCollapsed: shouldDefaultCollapseLayout(
+      window.location.pathname,
+      window.location.search,
+    ),
```
AbigailDeng: This looks like it only changes ProLayout's initial collapsed state. When the user clicks the Teams CTA and navigates to `/studio?intent=create-member` inside the SPA, the existing layout instance is typically reused, so `defaultCollapsed` is unlikely to collapse the menu on that transition. In practice the focused Studio entry may only collapse after a hard refresh or direct deep link. This probably needs a controlled collapsed value that reacts to location changes, not just a different default.

Issue #340, opened 2026-04-23
feat(console-web): Member-first Invoke flow and Studio closeout


AbigailDeng:
Part of #322

## Background
Invoke 是 member-first 生命周期的最后一个核心交互页。
同时，Studio 还需要把剩余的 Observe、Team entry、深链、Workflow polish 和全链路回归一起收尾，才能真正完成 #322 的可验收闭环。

## Goal
先把 Invoke 页面收口为“只调用当前 member”，再完成 Studio 剩余 closeout，让用户可以顺畅完成：
`Build -> Bind -> Invoke -> Observe`

## Scope
Day 1
- Invoke 页面只围绕当前 member
- 收口 endpoint、request editor、send、history、transcript
- 保证从 Bind 进入 Invoke 的上下文稳定

Day 2
- Observe 做最小 member-first 收口
- Team Detail -> Studio entry handoff 收尾
- 深链语义稳定到 `scopeId + memberId + step + focus`
- Workflow 剩余 member-first polish 收尾
- 补全主链回归测试

## Out of Scope
- 新后端 member API
- 超出 #322 范围的新功能扩张
- 独立的视觉大改版

## Main Files
- `apps/aevatar-console-web/src/pages/studio/components/StudioMemberInvokePanel.tsx`
- `apps/aevatar-console-web/src/pages/studio/index.tsx`
- `apps/aevatar-console-web/src/pages/teams/detail.tsx`
- `apps/aevatar-console-web/src/shared/studio/navigation.ts`
- `apps/aevatar-console-web/src/pages/studio/index.test.tsx`

## Checklist
- [ ] Invoke 目标明确为当前 member
- [ ] endpoint / request editor / send / history / transcript 收口
- [ ] Observe 最小 member-first 收口
- [ ] Team entry / 深链稳定
- [ ] Workflow 剩余 polish 收尾
- [ ] 主链回归测试补齐

## Acceptance
- 用户能顺畅完成 `Build -> Bind -> Invoke -> Observe`
- 从 Team 进入 Studio 后上下文稳定
- Studio 主链不再出现明显主语漂移
- member-first 核心回归测试通过

AbigailDeng:
Linking the backend blocker explicitly here: #364

Current read after frontend/codebase review:

- `#340` can still complete a meaningful frontend closeout on the current legacy runtime/binding APIs
- but `#340` cannot honestly become **100% pure member-first end-to-end** until the backend gaps in `#364` are resolved

[...]

PR #427, opened 2026-04-27
Add member-first Studio APIs

### LINKED ISSUE #364: [Blocker] Backend gaps blocking 100% pure member-first Invoke/Observe closeout for #340
## Related

- parent frontend closeout: #340
- broader backend track: #325
- umbrella issue: #322

## Why this issue exists

#325 covers the broader backend member model work.

This issue is narrower on purpose:

it isolates the backend contract gaps that specifically block `#340 feat(console-web): Member-first Invoke flow and Studio closeout` from reaching a **100% pure member-first** outcome on the current frontend deadline.

In other words:

- `#340` can still ship a meaningful frontend closeout on top of the current legacy runtime/binding APIs
- but `#340` **cannot honestly become fully member-first end-to-end** without the backend pieces below

> 本 issue 的 deadline 仅通过 GitHub Milestone due date 管理。

## Current frontend reality

Today Studio Invoke/Observe still ultimately rely on:

- `scopeId + serviceId + endpointId` for invoke
- `scopeId + serviceId + runId` for runs / observe
- scope binding state as the main source of published contract truth

That means frontend can improve framing, navigation, handoff, and lifecycle continuity, but it still cannot make `memberId` the real authority of bind / invoke / observe.

### 1. No authoritative `memberId -> publishedServiceId` mapping

Frontend currently has no backend-owned truth for:

- which published service belongs to the current member
- whether that mapping is stable across rename / rebuild / rebind

Without this mapping, frontend must keep using legacy `serviceId` semantics in places where product semantics want `memberId`.

### 2. No member-centric binding API

Missing backend APIs such as:

- `GET /api/scopes/{scopeId}/members/{memberId}/binding`
- `PUT /api/scopes/{scopeId}/members/{memberId}/binding`

Without these, Bind cannot truly mean:

> publish the current member revision to that member's stable published service

Frontend must keep reading/writing scope/service binding state instead.

### 3. No member-centric invoke API

Missing backend APIs such as:

- `POST /api/scopes/{scopeId}/members/{memberId}/invoke/{endpointId}`
- `POST /api/scopes/{scopeId}/members/{memberId}/invoke/{endpointId}:stream`

Without these, Invoke can only call runtime through legacy scope/service routes.

That means Studio can look member-first, but the real invoke target is still resolved via service-level runtime semantics.

### 4. No member-centric runs / observe query API

Missing backend APIs such as:

- `GET /api/scopes/{scopeId}/members/{memberId}/runs`
- `GET /api/scopes/{scopeId}/members/{memberId}/runs/{runId}`

Without these, Observe cannot truly answer:

> show me the runtime facts for the current member

It can only continue to interpret runs through legacy service/runtime paths.

### 5. No backend member roster / detail truth for Studio routing

Missing backend member-centric list/detail APIs mean frontend still has to infer or bridge member focus from workflow / script / published service facts.

That keeps route semantics unstable:

- route `memberId` cannot yet be the fully honest primary key
- deep link continuity across `Build -> Bind -> Invoke -> Observe` remains partly compatibility-driven

## Deadline impact on frontend #340

This is the critical part.

If the backend items above are **not available before the current #322 closeout milestone due date**, frontend `#340` can still finish a strong closeout, but only with a reduced truth level:

### Frontend can still finish

- member-first Invoke page framing
- stable `Bind -> Invoke` handoff
- lifecycle gating fixes
- Team entry / Studio deep-link cleanup
- member-first Observe framing and staged empty states
- main regression coverage

### Frontend cannot honestly claim

- route `memberId` is the true authority for invoke
- Bind is fully member-owned instead of scope/service-backed
- Observe is querying member-owned runtime history
- Studio no longer depends on legacy service-level runtime semantics

### Practical risk to the deadline

If backend contract work slips, the likely outcome is:

1. frontend ships a **compatibility-mode member-first closeout** for #340
2. the final acceptance language for #340 must explicitly avoid promising a fully pure member-first runtime chain
3. there will still be follow-up work after the current closeout milestone to remove legacy service-first runtime assumptions

This is important because otherwise frontend may appear visually done while still carrying semantic debt that blocks final closure of #322.

## Minimum backend deliverables that would unblock pure closeout

The minimum useful backend slice for `#340` is smaller than the full #325 scope:

- [ ] authoritative `memberId -> publishedServiceId` read path
- [ ] `GET /api/scopes/{scopeId}/members/{memberId}/binding`
- [ ] `PUT /api/scopes/{scopeId}/members/{memberId}/binding`
- [ ] member-centric invoke route or equivalent backend resolver that lets frontend invoke by `memberId`
- [ ] member-centric runs query path for Observe

## Acceptance for this blocker issue

- [ ] Backend contract gaps impacting pure member-first `#340` are explicitly tracked in one place
- [ ] `#340` can reference this issue to distinguish frontend closeout from backend semantic blockers
- [ ] team alignment is explicit on what can ship by the current closeout milestone due date vs. what still depends on backend

## One-line outcome

If this issue is unresolved, `#340` should be treated as a **frontend closeout on top of legacy scope/service runtime semantics**, not the final fully pure member-first backend truth.

eanzhao:
## Problem
Studio Bind / Invoke / Observe needs member-first backend routes so the frontend can address a member and let the backend resolve the published service id.

Closes #364

## Solution
- Add `IMemberPublishedServiceResolver` and a deterministic default resolver that maps normalized `memberId` to the member-owned `publishedServiceId` while rejecting service-key-unsafe member ids.
- Add member-first scope routes for published service resolution, binding upsert/status, invoke/stream invoke, run list/detail, audit, resume, signal, and stop.
- Gate member binding writes to scope-admin until an actor-owned member catalog defines authoritative membership, ownership, and cleanup semantics.
- Require member routes to pass both scope access and member access, allowing either the matching authenticated member or a scope-admin.
- Remove public member API `appId` overrides and app/namespace exposure from member resolver/response contracts.
- Keep observe/run reads on existing read-model-backed services and document the contract in `docs/2026-04-27-member-first-studio-apis.md`.

## Impact Paths
- `src/Aevatar.Hosting`
- `src/platform/Aevatar.GAgentService.Abstractions`
- `src/platform/Aevatar.GAgentService.Application`
- `src/platform/Aevatar.GAgentService.Hosting`
- `test/Aevatar.GAgentService.Tests`
- `test/Aevatar.GAgentService.Integration.Tests`
- `test/Aevatar.Studio.Tests`
- `docs/`

## Validation
- PASS: `dotnet test test/Aevatar.GAgentService.Tests/Aevatar.GAgentService.Tests.csproj --nologo --filter DefaultMemberPublishedServiceResolverTests`
- PASS: `dotnet test test/Aevatar.Studio.Tests/Aevatar.Studio.Tests.csproj --nologo --filter ScopeEndpointAccessTests`
- PASS: `dotnet test test/Aevatar.GAgentService.Integration.Tests/Aevatar.GAgentService.Integration.Tests.csproj --nologo --filter ScopeServiceEndpointsTests`
- PASS: `dotnet test test/Aevatar.GAgentService.Tests/Aevatar.GAgentService.Tests.csproj --nologo`
- PASS: `dotnet build aevatar.slnx --nologo`
- PASS: `bash tools/ci/test_stability_guards.sh`
- PASS: `bash tools/ci/query_projection_priming_guard.sh`
- PASS: `git diff --check`
- PARTIAL: `bash tools/ci/architecture_guards.sh` passes through proto/channel/projection/workflow guards, then fails in `playground_asset_drift_guard.sh` because existing CLI playground build output differs from `demos/Aevatar.Demos.Workflow.Web/wwwroot` assets. This PR does not modify those frontend assets.

### DIFF
--- a/docs/2026-04-27-member-first-studio-apis.md
+++ b/docs/2026-04-27-member-first-studio-apis.md
@@ -1,0 +1,40 @@
+# Member-First Studio APIs
+
+## Scope
+
+This slice closes the minimum backend contract gap for Studio member-first Bind / Invoke / Observe flows.
+
+The current authoritative resolver maps each normalized `memberId` to a stable published service with the same id:
+
+| Field | Meaning |
+|---|---|
+| `scopeId` | Team scope that owns the member and published service. |
+| `memberId` | Studio member address used by frontend routes. |
+| `publishedServiceId` | Stable service id used by backend service runtime for that member. |
+| `publishedServiceKey` | Internal service identity key for diagnostics and contract inspection. |
+
+The resolver is exposed through `IMemberPublishedServiceResolver`, so a later actor-owned member catalog can replace the deterministic mapping without changing HTTP routes. Until that catalog is authoritative, member binding writes are restricted to scope administrators; member reads and invokes require either the matching authenticated member claim or a scope administrator role.
+
+## Routes
+
+| Route | Purpose |
+|---|---|
+| `GET /api/scopes/{scopeId}/members/{memberId}/published-service` | Resolve the member-owned published service id. |
+| `GET /api/scopes/{scopeId}/members/{memberId}/binding` | Read current binding status for the member-owned published service. |
+| `PUT /api/scopes/{scopeId}/members/{memberId}/binding` | Publish workflow/script/GAgent implementation to the member-owned published service. |
+| `POST /api/scopes/{scopeId}/members/{memberId}/invoke/{endpointId}` | Invoke a typed endpoint by member id. |
+| `POST /api/scopes/{scopeId}/members/{memberId}/invoke/{endpointId}:stream` | Invoke an SSE endpoint by member id. |
+| `GET /api/scopes/{scopeId}/members/{memberId}/runs` | List read-model-backed runs for the member-owned published service. |
+| `GET /api/scopes/{scopeId}/members/{memberId}/runs/{runId}` | Read a run summary for the member-owned published service. |
+| `GET /api/scopes/{scopeId}/members/{memberId}/runs/{runId}/audit` | Read a run audit report for the member-owned published service. |
+| `POST /api/scopes/{scopeId}/members/{memberId}/runs/{runId}:resume` | Resume a member-owned published service run. |
+| `POST /api/scopes/{scopeId}/members/{memberId}/runs/{runId}:signal` | Signal a member-owned published service run. |
+| `POST /api/scopes/{scopeId}/members/{memberId}/runs/{runId}:stop` | Stop a member-owned published service run. |
+
+## Semantics
+
+- Member routes for Bind / Invoke / Observe-read / run lifecycle control do not require frontend callers to know or pass `serviceId`.
+- Binding and invoke still use the existing service command/runtime path after the resolver has produced `publishedServiceId`.
+- Runs and run detail still read workflow run read models; they do not query actor state or replay events.
+- Responses use `publishedServiceId` instead of overloading `serviceId` in member-centric DTOs.
+- The member-first public contract does not accept an `appId` override or expose the fixed service namespace.
--- a/src/platform/Aevatar.GAgentService.Abstractions/Ports/IMemberPublishedServiceResolver.cs
+++ b/src/platform/Aevatar.GAgentService.Abstractions/Ports/IMemberPublishedServiceResolver.cs
@@ -1,0 +1,17 @@
+namespace Aevatar.GAgentService.Abstractions.Ports;
+
+public sealed record MemberPublishedServiceResolveRequest(
+    string ScopeId,
+    string MemberId);
+
+public sealed record MemberPublishedServiceResolution(
+    string ScopeId,
+    string MemberId,
+    string PublishedServiceId);
+
+public interface IMemberPublishedServiceResolver
+{
+    Task<MemberPublishedServiceResolution> ResolveAsync(
+        MemberPublishedServiceResolveRequest request,
+        CancellationToken ct = default);
+}
--- a/src/platform/Aevatar.GAgentService.Application/Bindings/DefaultMemberPublishedServiceResolver.cs
+++ b/src/platform/Aevatar.GAgentService.Application/Bindings/DefaultMemberPublishedServiceResolver.cs
@@ -1,0 +1,34 @@
+using Aevatar.GAgentService.Abstractions.Ports;
+using Aevatar.GAgentService.Application.Workflows;
+
+namespace Aevatar.GAgentService.Application.Bindings;
+
+public sealed class DefaultMemberPublishedServiceResolver : IMemberPublishedServiceResolver
+{
+    public Task<MemberPublishedServiceResolution> ResolveAsync(
+        MemberPublishedServiceResolveRequest request,
+        CancellationToken ct = default)
+    {
+        ArgumentNullException.ThrowIfNull(request);
+        ct.ThrowIfCancellationRequested();
+
+        var scopeId = ScopeWorkflowCapabilityOptions.NormalizeRequired(request.ScopeId, nameof(request.ScopeId));
+        var memberId = NormalizeMemberId(request.MemberId);
+
+        // TODO: replace this deterministic development resolver with an actor-owned member
+        // catalog that defines authoritative membership, ownership, and cleanup semantics.
+        return Task.FromResult(new MemberPublishedServiceResolution(
+            scopeId,
+            memberId,
+            memberId));
+    }
+
+    private static string NormalizeMemberId(string memberId)
+    {
+        var normalized = ScopeWorkflowCapabilityOptions.NormalizeRequired(memberId, nameof(MemberPublishedServiceResolveRequest.MemberId));
+        if (normalized.IndexOfAny([':', '/', '\\', '?', '#']) >= 0)
+            throw new InvalidOperationException("memberId must not contain ':', '/', '\\', '?' or '#'.");
+
+        return normalized;
+    }
+}
--- a/src/platform/Aevatar.GAgentService.Hosting/Endpoints/ScopeServiceEndpoints.cs
+++ b/src/platform/Aevatar.GAgentService.Hosting/Endpoints/ScopeServiceEndpoints.cs
@@ -53,14 +53,25 @@ public static class ScopeServiceEndpoints
         group.MapPost("/{scopeId}/workflow/draft-run", HandleDraftRunAsync);
         group.MapPut("/{scopeId}/binding", HandleUpsertBindingAsync);
         group.MapGet("/{scopeId}/binding", HandleGetBindingAsync);
+        group.MapGet("/{scopeId}/members/{memberId}/published-service", HandleGetMemberPublishedServiceAsync);
+        group.MapPut("/{scopeId}/members/{memberId}/binding", HandleUpsertMemberBindingAsync);
+        group.MapGet("/{scopeId}/members/{memberId}/binding", HandleGetMemberBindingAsync);
         group.MapPost("/{scopeId}/binding/revisions/{revisionId}:activate", HandleActivateBindingRevisionAsync);
         group.MapGet("/{scopeId}/revisions", HandleGetDefaultServiceRevisionsAsync);
         group.MapGet("/{scopeId}/revisions/{revisionId}", HandleGetDefaultServiceRevisionAsync);
         group.MapPost("/{scopeId}/binding/revisions/{revisionId}:retire", HandleRetireBindingRevisionAsync);
         group.MapPost("/{scopeId}/invoke/chat:stream", HandleInvokeDefaultChatStreamAsync);
         group.MapPost("/{scopeId}/invoke/{endpointId}", HandleInvokeDefaultAsync);
+        group.MapPost("/{scopeId}/members/{memberId}/invoke/{endpointId}:stream", HandleInvokeMemberStreamAsync);
+        group.MapPost("/{scopeId}/members/{memberId}/invoke/{endpointId}", HandleInvokeMemberAsync);
         group.MapGet("/{scopeId}/runs", HandleListDefaultRunsAsync);
         group.MapGet("/{scopeId}/runs/{runId}", HandleGetDefaultRunAsync);
+        group.MapGet("/{scopeId}/members/{memberId}/runs", HandleListMemberRunsAsync);
+        group.MapGet("/{scopeId}/members/{memberId}/runs/{runId}", HandleGetMemberRunAsync);
+        group.MapGet("/{scopeId}/members/{memberId}/runs/{runId}/audit", HandleGetMemberRunAuditAsync);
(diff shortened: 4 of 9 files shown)


src/platform/Aevatar.GAgentService.Application/Bindings/DefaultMemberPublishedServiceResolver.cs
```diff
@@ -0,0 +1,38 @@
...
+        var memberId = ScopeWorkflowCapabilityOptions.NormalizeRequired(request.MemberId, nameof(request.MemberId));
+        var appId = ScopeWorkflowCapabilityConventions.ResolveAppId(_options, request.AppId);
+        var @namespace = ScopeWorkflowCapabilityOptions.NormalizeRequired(
+            _options.ServiceNamespace,
+            nameof(_options.ServiceNamespace));
+
+        return Task.FromResult(new MemberPublishedServiceResolution(
+            scopeId,
+            memberId,
+            memberId,
```
eanzhao: **Default resource semantics gap (CLAUDE.md "默认路径须定义资源语义").** This resolver maps *any* trimmed `memberId` to a `publishedServiceId` 1:1, with no membership truth, no scope-side member catalog, and no creation/cleanup contract. Combined with `PUT /api/scopes/{scopeId}/members/{memberId}/binding`, this means: any caller authorized for the scope can pass an arbitrary `memberId` and the system silently materialises a brand-new published-service identity for it — with no owner, no reuse rule, and no cleanup responsibility.

The doc explicitly leaves this for a future actor-owned member catalog (`IMemberPublishedServiceResolver` is a placeholder), which is a reasonable design choice for the slice — but the HTTP route is now *live* and write-capable, so the gap is exposed before the actor exists. Per the rule "任何'缺失即创建'策略须同时定义归属、复用规则和清理责任", the minimum is one of:

1. Reject memberIds that don't already exist in a scope-side member roster (read-side pre-check), so PUT cannot conjure new services; or
2. Add a clearly-scoped TODO in the resolver pointing to the follow-up actor catalog issue, and gate `PUT` behind a scope-admin claim until the catalog lands; or
3. Track ownership/cleanup as a hard blocker before this contract is shipped to the frontend as authoritative.

As written, the resolver is exactly the "通用接口允许绕过 actor 边界 / 权威源" anti-pattern called out in the design completeness rules.

src/platform/Aevatar.GAgentService.Application/Bindings/DefaultMemberPublishedServiceResolver.cs
```diff
@@ -0,0 +1,38 @@
...
+
+    public Task<MemberPublishedServiceResolution> ResolveAsync(
+        MemberPublishedServiceResolveRequest request,
+        CancellationToken ct = default)
+    {
+        ArgumentNullException.ThrowIfNull(request);
+        ct.ThrowIfCancellationRequested();
+
+        var scopeId = ScopeWorkflowCapabilityOptions.NormalizeRequired(request.ScopeId, nameof(request.ScopeId));
+        var memberId = ScopeWorkflowCapabilityOptions.NormalizeRequired(request.MemberId, nameof(request.MemberId));
```
eanzhao: `memberId` is normalized via `NormalizeRequired` (only trim+empty check), then returned verbatim as `PublishedServiceId`. That serviceId is concatenated into `publishedServiceKey` via `ServiceKeys.Build` which joins with `:`. If a caller passes `memberId = "foo:bar"`, the resulting key becomes `scope-a:default:default:foo:bar` — same shape as a 5-segment key, ambiguous to anything that splits the key for diagnostics/inspection (which is what the doc says `publishedServiceKey` is for).

The existing workflowId path uses `ScopeWorkflowCapabilityConventions.NormalizeWorkflowId`, which explicitly rejects `:`. The member resolver should apply the same rejection (and probably reject other characters that cause issues in identity/path segments). Worth a unit test for the rejection too — `DefaultMemberPublishedServiceResolverTests` currently only covers blank rejection.

src/platform/Aevatar.GAgentService.Hosting/Endpoints/ScopeServiceEndpoints.cs
```diff
@@ -249,6 +256,142 @@ private static async Task<IResult> HandleGetBindingAsync(
...
+        string scopeId,
+        string memberId,
+        UpsertScopeBindingHttpRequest request,
+        [FromServices] IMemberPublishedServiceResolver memberPublishedServiceResolver,
+        [FromServices] IScopeBindingCommandPort commandPort,
+        CancellationToken ct)
+    {
+        try
+        {
+            if (AevatarScopeAccessGuard.TryCreateScopeAccessDeniedResult(http, scopeId, out var denied))
```
eanzhao, line 309: **Authorization is scope-only; no per-member ownership check.** `AevatarScopeAccessGuard` validates only that the caller's `scope_id` claim matches the route `scopeId`. Combined with the deterministic `memberId == publishedServiceId` mapping, this means user A authenticated for scope X can:

- `PUT /api/scopes/X/members/<userB>/binding` — overwrite user B's published service definition
- `POST /api/scopes/X/members/<userB>/invoke/<endpoint>` — invoke user B's endpoint as themselves
- `GET /api/scopes/X/members/<userB>/runs` — read user B's run history

If the product semantics treat `memberId` as a per-user address (which the issue #364 framing strongly implies — "member-first" replaces "service-first" precisely so `memberId` is the user-facing primary key), this is a horizontal privilege escalation across members of the same scope. The current `IMemberPublishedServiceResolver` placeholder has no hook for caller-vs-member checks.

Minimum: an `IMemberAccessGuard` (or extending the resolver request to carry the caller's identity and assert ownership) should land alongside the actor-owned catalog. Until then, write/invoke routes should probably be gated to scope-admin or carry a clear `// TODO: per-member auth pending #...` so this isn't accidentally treated as the long-lived contract.

docs/2026-04-27-member-first-studio-apis.md
```diff
@@ -0,0 +1,35 @@
...
+| `GET /api/scopes/{scopeId}/members/{memberId}/binding` | Read current binding status for the member-owned published service. |
+| `PUT /api/scopes/{scopeId}/members/{memberId}/binding` | Publish workflow/script/GAgent implementation to the member-owned published service. |
+| `POST /api/scopes/{scopeId}/members/{memberId}/invoke/{endpointId}` | Invoke a typed endpoint by member id. |
+| `POST /api/scopes/{scopeId}/members/{memberId}/invoke/{endpointId}:stream` | Invoke an SSE endpoint by member id. |
+| `GET /api/scopes/{scopeId}/members/{memberId}/runs` | List read-model-backed runs for the member-owned published service. |
+| `GET /api/scopes/{scopeId}/members/{memberId}/runs/{runId}` | Read a run summary for the member-owned published service. |
+
+## Semantics
+
+- Member routes do not require frontend callers to know or pass `serviceId`.
```
eanzhao: > Member routes do not require frontend callers to know or pass `serviceId`.

This overstates the contract. The PR exposes member-level routes for `binding`, `invoke`, `invoke:stream`, `runs`, `runs/{runId}` — but not `runs/{runId}:resume`, `runs/{runId}:signal`, `runs/{runId}:stop`, or `runs/{runId}/audit`. For those, the frontend must still hit `/services/{serviceId}/runs/{runId}:...`, which means deep-linking from a member-first run detail page back into a control op falls back to legacy serviceId semantics — exactly the "compatibility-mode member-first" outcome #364 wanted to avoid.

Either:
- add the missing run lifecycle ops at the member level in this slice (they're trivial wrappers via the same resolver), or
- soften the doc claim to scope it to Bind / Invoke / Observe-read explicitly, and track the lifecycle ops as a follow-up so #340 can plan accordingly.

src/platform/Aevatar.GAgentService.Abstractions/Ports/IMemberPublishedServiceResolver.cs
```diff
@@ -0,0 +1,20 @@
...
+    string ScopeId,
+    string MemberId,
+    string? AppId = null);
+
+public sealed record MemberPublishedServiceResolution(
+    string ScopeId,
+    string MemberId,
+    string PublishedServiceId,
+    string AppId,
+    string Namespace);
```
eanzhao: `MemberPublishedServiceResolveRequest.AppId` and `MemberPublishedServiceResolution.Namespace` (plus the `appId` query param on every member endpoint) are surfaced to callers, but `ScopeWorkflowCapabilityOptions.ServiceAppId` and `ServiceNamespace` are *hard-pinned* to `"default"` (their setters are no-ops). So:

- `Namespace` in the response is always `"default"` — leaks an internal constant.
- `AppId` is overrideable via query (`ResolveAppId` returns the override if non-empty), but the underlying default service flow is pinned. Passing `?appId=foo` produces a `ServiceIdentity` with `AppId=foo` while every other lookup path resolves to `default` — the resulting key/identity may not match anything in the catalog.

For the member-first contract specifically, both fields look like they should be removed from the public surface (resolver request + response DTO + query parameters) since the doc table doesn't mention them and the system has no story for non-default values. Otherwise they're foot-gun surface — "字段单一语义" is violated when a field is documented as one thing but silently ignored or coerced.

eanzhao (commented):
Reviewed against CLAUDE.md architecture rules and the linked #364 minimum slice. Tests + build pass and the slice is correctly scoped to unblock #340, but the new write surface introduces ownership/auth/validation concerns that should at minimum be tracked as known follow-ups before this becomes the long-lived contract — see inline comments.

eanzhao (commented):
All five comments addressed cleanly:

1. **Resource semantics gap** — `// TODO` added in resolver pointing to actor-owned catalog; PUT binding now requires scope-admin role via `AevatarMemberAccessGuard.TryCreateScopeAdminRequiredResult`. ✅
2. **memberId validation** — `NormalizeMemberId` rejects `:`, `/`, `\`, `?`, `#`; new test `ResolveAsync_ShouldRejectMemberIdThatBreaksServiceKeySegments` covers it. ✅
3. **Authorization** — New `AevatarMemberAccessGuard` enforces matching member claim or scope-admin role for reads/invoke, and scope-admin only for PUT binding; integration tests cover allow/deny paths via `X-Test-Member-Id` / `X-Test-Role` headers. ✅
4. **Run lifecycle ops** — Added `runs/{runId}/audit`, `:resume`, `:signal`, `:stop` member endpoints with full integration tests; doc claim softened and table updated. ✅
5. **AppId / Namespace leakage** — Removed `AppId` from request DTO, `AppId`/`Namespace` from response DTO, `appId` query param from all member endpoints; new `UpsertMemberScopeBindingHttpRequest` has clean public surface; doc adds explicit semantic. ✅

Resolver unit tests pass locally (3/3).

One minor observation, not blocking — feel free to fold into a follow-up: `AevatarMemberAccessGuard.HasScopeAdminRole` checks for any admin-typed role on the principal without validating it's bound to the requested scope. Today this is fine because `AevatarScopeAccessGuard` already enforces single-scope identity (rejects ambiguous scope claims), so admin-of-scope-A can't reach the member guard for scope-B. But if the system later supports multi-scope users, the role claim would need a scope binding (e.g., `{type: "scope.role:scope-a", value: "admin"}`) to avoid cross-scope admin escalation. Cheap to fix later when the actor-owned catalog lands.

PR #428, opened 2026-04-27
feat(studio): member-first authority + bind orchestration (#325)

### LINKED ISSUE #325: [Feature] Backend: member-first Studio APIs with stable published service per member
## Summary

Introduce a real `member` backend model for Studio so each member has a stable, backend-generated `publishedServiceId`, and Studio can bind `workflow / script / gagent` implementations to that member-owned published service instead of falling back to the scope default service.

## Why

Studio is already trying to present a `member-first` lifecycle (`Build / Bind / Invoke / Observe`), but the current backend contract is still mostly `scope-first`:

- `PUT /api/scopes/{scopeId}/binding` defaults to the scope default service when `ServiceId` is omitted.
- Studio currently does not have a real backend `member` entity to anchor identity.
- `serviceId` is still leaking into places where the product model really wants `memberId`.

That prevents the correct product behavior:

- `Create member` should support `workflow / script / gagent`.
- Every member should own one stable published service.
- `Bind` should mean: publish the current member revision to that member's service.
- `serviceId` should be a published contract, not a substitute for `memberId`.

### 1. Introduce a real member model

Add a backend-owned `TeamMember` / `StudioMember` model (naming can follow the domain conventions) with at least:

- `memberId`
- `scopeId`
- `displayName`
- `description?`
- `implementationKind`
- `implementationRef`
- `publishedServiceId`
- `lifecycleStage`
- `createdAt`
- `updatedAt`

Notes:

- `memberId` is the stable identity.
- `publishedServiceId` is the stable published contract surface.
- `workflowId / scriptId / gagentTypeName` are implementation refs, not the member identity.

### 2. Backend-generated stable `publishedServiceId`

When a member is created, backend should generate and persist a stable `publishedServiceId`.

Requirements:

- generated by backend, not frontend
- generated once at member creation time
- never recomputed from mutable display name on read
- rename-safe
- ordinary users do not need to type or manage it

Recommended rule:

- derive from immutable `memberId` (optionally with a readable slug prefix)
- store the generated value directly on the member record

### 3. Member-centric Studio APIs

Add member-centric APIs for Studio:

- `POST /api/scopes/{scopeId}/members`
- `GET /api/scopes/{scopeId}/members`
- `GET /api/scopes/{scopeId}/members/{memberId}`
- `PUT /api/scopes/{scopeId}/members/{memberId}/binding`
- `GET /api/scopes/{scopeId}/members/{memberId}/binding`

`POST /members` should support creating members with:

- `implementationKind = workflow`
- `implementationKind = script`
- `implementationKind = gagent`

and return at least:

- `memberId`
- `displayName`
- `implementationKind`
- `publishedServiceId`
- `lifecycleStage`
- initial implementation ref when available

### 4. Member binding behavior

`PUT /members/{memberId}/binding` should:

1. load the member
2. resolve that member's stable `publishedServiceId`
3. publish the current implementation revision to that service
4. return binding/contract metadata for Studio

This should support all three implementation kinds:

- workflow
- script
- gagent

## Reuse existing backend capabilities

The new member-centric path should reuse existing capabilities instead of rebuilding service binding logic from scratch.

Already reusable:

- current scope binding command already supports explicit `ServiceId`
- current scope binding command already supports:
  - workflow
  - script
  - gagent
- existing service-level invoke / revisions / runs / bindings endpoints can remain the runtime/governance surface behind the member layer

So the main missing backend work is:

- stable member identity
- stable member-owned published service mapping
- member-centric API surface
- binding orchestration through that mapping

## Acceptance criteria

- [ ] Backend has a first-class member model with persistent `memberId` and `publishedServiceId`
- [ ] Creating a member supports `workflow / script / gagent`
- [ ] `publishedServiceId` is backend-generated, stable, and rename-safe
- [ ] Studio can list members from backend without inferring them from workflow files or published services
- [ ] Member binding publishes to the member's own stable published service, not the scope default service
- [ ] Binding supports workflow, script, and gagent members
- [ ] `serviceId` is no longer required as a user-facing input for ordinary Studio member binding
- [ ] Legacy scope-first binding endpoints can remain available for other surfaces, but Studio is no longer blocked on them as its primary contract

## Non-goals

- frontend UI redesign in this issue
- full team router redesign in this issue
- removing all legacy scope-level runtime endpoints immediately

## Migration / compatibility notes

Suggested compatibility approach:

- keep existing scope-first endpoints for legacy/runtime surfaces
- introduce member-first endpoints in parallel
- let Studio migrate to the new member-first APIs
- provide a compatibility/bootstrap path for existing members/scope-bound assets where needed

## Related docs

- `docs/design/2026-04-22-team-member-first-prd.md`
- `docs/design/2026-04-22-studio-member-lifecycle-spec.md`
- `docs/design/2026-04-21-studio-workflow-member-lifecycle-prd.md`

## One-line outcome

After this issue, Studio should be able to treat `member` as the only primary object, with `workflow / script / gagent` as implementation types and `published service` as the member's stable publish surface.

eanzhao:
Closes #325.

## Summary
- Introduces a real `member` backend authority (`StudioMemberGAgent`) so Studio binds workflows / scripts / gagents to a member's own stable `publishedServiceId` instead of falling back to the scope-default service.
- `publishedServiceId` is generated once at create time from the immutable `memberId` and persisted on the actor — never recomputed on read, never mutated by rename.
- New member-first HTTP surface under `/api/scopes/{scopeId}/members[...]`. ServiceId is no longer a user-facing input.
- Bind orchestration delegates to the existing `IScopeBindingCommandPort` with `ServiceId = publishedServiceId`, then records the resulting revision back on the member authority — no second binding system.

## Acceptance criteria from #325
- [x] First-class member model with persistent `memberId` and `publishedServiceId`.
- [x] Creating a member supports `workflow / script / gagent`.
- [x] `publishedServiceId` is backend-generated, stable, and rename-safe (locked in by `StudioMemberGAgentStateTests` + `StudioMemberConventionsTests`).
- [x] Studio can list members from backend without inferring them from workflow files or services.
- [x] Member binding publishes to the member's own stable published service, not the scope default service (locked in by `StudioMemberServiceBindingTests`).
- [x] Binding supports workflow, script, and gagent members.
- [x] `serviceId` is not a user-facing input on member endpoints.
- [x] Legacy scope-first endpoints under `Aevatar.GAgentService.Hosting` remain untouched and continue to work.

### `agents/Aevatar.GAgents.StudioMember/` (new)
- `studio_member_messages.proto` — typed state, events, implementation refs (no generic bag).
- `StudioMemberGAgent.cs` — actor that owns the canonical authority state.
- `StudioMemberConventions.cs` — actor id (`studio-member:{scopeId}:{memberId}`) + rename-safe `publishedServiceId` (`member-{memberId}`).

### `src/Aevatar.Studio.Projection`
- `ReadModels/studio_projection_readmodels.proto` — adds `StudioMemberCurrentStateDocument` with denormalized roster fields.
- `Projectors/StudioMemberCurrentStateProjector.cs` — materializes committed events into the read model.
- `QueryPorts/ProjectionStudioMemberQueryPort.cs` — pure-read query port; roster scans constrained to `scope_id`.
- `CommandServices/ActorDispatchStudioMemberCommandService.cs` — write-side command port that dispatches via `IStudioActorBootstrap` + `IActorDispatchPort`.

### `src/Aevatar.Studio.Application`
- `Studio/Abstractions/IStudioMemberCommandPort.cs`, `IStudioMemberQueryPort.cs`, `IStudioMemberService.cs`.
- `Studio/Contracts/MemberContracts.cs` + `MemberImplementationKindMapper.cs` — wire ↔ proto mapping.
- `Studio/Services/StudioMemberService.cs` — bind orchestration (resolves member → builds `ScopeBindingUpsertRequest` with `ServiceId = publishedServiceId` → delegates to `IScopeBindingCommandPort` → records revision back on member authority).

### `src/Aevatar.Studio.Hosting`
- `Endpoints/StudioMemberEndpoints.cs` — `POST/GET/PUT` under `/api/scopes/{scopeId}/members[...]`.
- DI registrations + `StudioMemberCurrentStateDocument` document store provider hooks (Elasticsearch + InMemory).

### `test/Aevatar.Studio.Tests` (+19 new tests)
- `StudioMemberConventionsTests` — actor-id layout, rename-safety of `publishedServiceId`.
- `StudioMemberImplementationKindMapperTests` — wire ↔ proto mapping.
- `StudioMemberGAgentStateTests` — state transitions including rename-safety invariant.
- `StudioMemberServiceBindingTests` — bind orchestration: ServiceId is always the member's `publishedServiceId`, revision recorded back, all three impl kinds supported.

## Test plan
- [x] `dotnet build aevatar.slnx --nologo` — 0 errors.
- [x] `dotnet test test/Aevatar.Studio.Tests/Aevatar.Studio.Tests.csproj` — 235 passed (no regressions).
- [x] `bash tools/ci/projection_state_version_guard.sh`
- [x] `bash tools/ci/projection_state_mirror_current_state_guard.sh`
- [x] `bash tools/ci/projection_route_mapping_guard.sh`
- [x] `bash tools/ci/query_projection_priming_guard.sh`
- [x] `bash tools/ci/workflow_binding_boundary_guard.sh`
- [x] `bash tools/ci/proto_lint_guard.sh`
- [x] `bash tools/ci/committed_state_projection_guard.sh`
- [x] `bash tools/ci/cqrs_eventsourcing_boundary_guard.sh`

## Non-goals (matches issue)
- No frontend work in this PR.
- Existing scope-first endpoints (`PUT /api/scopes/{scopeId}/binding` etc.) stay untouched for legacy / runtime callers.
- No team-router redesign.


src/Aevatar.Studio.Projection/CommandServices/ActorDispatchStudioMemberCommandService.cs
```diff
@@ -0,0 +1,196 @@
...
+    private readonly IStudioMemberQueryPort _queryPort;
+
+    public ActorDispatchStudioMemberCommandService(
+        IStudioActorBootstrap bootstrap,
+        IActorDispatchPort dispatchPort,
+        IStudioMemberQueryPort queryPort)
+    {
+        _bootstrap = bootstrap ?? throw new ArgumentNullException(nameof(bootstrap));
+        _dispatchPort = dispatchPort ?? throw new ArgumentNullException(nameof(dispatchPort));
+        _queryPort = queryPort ?? throw new ArgumentNullException(nameof(queryPort));
```
eanzhao: Dead injection. `_queryPort` is assigned but never read in this class. Either drop the dep + ctor parameter, or actually use it (e.g. for the duplicate-create check before dispatching `StudioMemberCreatedEvent`).

src/Aevatar.Studio.Application/Studio/Services/StudioMemberService.cs
```diff
@@ -0,0 +1,216 @@
...
+        var implementationKindWire = detail.Summary.ImplementationKind;
+        var bindingRequest = BuildScopeBindingRequest(
+            scopeId,
+            memberId,
+            publishedServiceId,
+            implementationKindWire,
+            detail.Summary.DisplayName,
+            request);
+
+        var bindingResult = await _scopeBindingCommandPort.UpsertAsync(bindingRequest, ct);
```
eanzhao, line 116: Bind flow gap vs issue #325.

The issue says `bind` should *publish the current member revision* to the member’s service — implying the implementation revision is already known to the member. Here, `BindAsync` reads the implementation spec out of the request body and forwards it directly to `_scopeBindingCommandPort.UpsertAsync` without ever calling `_memberCommandPort.UpdateImplementationAsync` or otherwise touching `implementation_ref` on the actor.

Net effect: `StudioMemberState.ImplementationRef` stays empty after bind, the `BuildReady` lifecycle stage is unreachable (since `ApplyBound` jumps straight to `BindReady`), and the read model can never answer “what is this member implemented by.” That contradicts the proto, the projector denormalization plan, and issue #325’s required `implementationRef` field.

Suggested fix: before delegating to the scope binding port, call `_memberCommandPort.UpdateImplementationAsync` (or fold the impl-ref persistence into the bind event). Then the member authority is the source of truth for `implementation_ref` and the lifecycle progression `Created → BuildReady → BindReady` becomes real.

src/Aevatar.Studio.Application/Studio/Services/StudioMemberService.cs
```diff
@@ -0,0 +1,216 @@
...
+
+        await _memberCommandPort.RecordBindingAsync(
+            scopeId,
+            memberId,
+            bindingResult.ServiceId,
+            bindingResult.RevisionId,
+            implementationKindWire,
+            ct);
+
```
eanzhao, line 133: Two-phase write with no compensation. If `RecordBindingAsync` throws after `_scopeBindingCommandPort.UpsertAsync` has succeeded, the platform-side has a real revision but the StudioMember actor doesn’t know about it. Next bind will create another upstream revision and only that one is recorded. Acceptable as “last write wins,” but worth either a comment or — preferably — making the read of `last_binding` lazy off the platform read model so the member doesn’t have to maintain its own copy.

src/Aevatar.Studio.Application/Studio/Abstractions/IStudioMemberCommandPort.cs
```diff
@@ -0,0 +1,47 @@
...
+    /// <summary>
+    /// Updates the implementation reference for an existing member (e.g. once
+    /// Studio has produced a workflow yaml, script revision, or wired up a
+    /// gagent type for the member). Idempotent on identical inputs.
+    /// </summary>
+    Task UpdateImplementationAsync(
+        string scopeId,
+        string memberId,
+        StudioMemberImplementationRefResponse implementation,
+        CancellationToken ct = default);
```
eanzhao, line 33: `UpdateImplementationAsync` has no caller in this PR. Either wire it into the bind flow (see comment on `StudioMemberService.BindAsync`), expose it via a `PUT /members/{memberId}/implementation` endpoint, or drop it. CLAUDE.md: *删除优先：空转发、重复抽象、无业务价值代码直接删除，不保留兼容空壳*.

agents/Aevatar.GAgents.StudioMember/StudioMemberGAgent.cs
```diff
@@ -0,0 +1,162 @@
...
+    [EventHandler(EndpointName = "renameMember")]
+    public async Task HandleRenamed(StudioMemberRenamedEvent evt)
+    {
+        if (string.IsNullOrEmpty(State.MemberId))
+        {
+            throw new InvalidOperationException("member not yet created.");
+        }
+
+        await PersistDomainEventAsync(evt);
+    }
```
eanzhao, line 67: `HandleRenamed` (and `StudioMemberRenamedEvent`) has no producer in this PR — there is no service / endpoint that emits a rename event. The rename-safe-`publishedServiceId` invariant is unit-tested at the state machine level (good), but the rename surface itself is unreachable end-to-end. Either expose `PUT /members/{memberId}` to drive it, or remove until the next PR needs it.

eanzhao (commented):
## Review summary

The shape of the PR is solid — typed proto contracts, dedicated actor, dedicated projector, denormalized roster fields, scope-bound query. The unit tests around `publishedServiceId` rename-safety and bind orchestration are clean and on-target.

However I see one **dead injection**, one **lifecycle-model gap that contradicts issue #325**, and several smaller concerns. Detailed inline comments below; the most important ones grouped here:

### 1. `implementation_ref` is modeled but never populated, and `bind_ready` is unreachable through `build_ready`

Issue #325 explicitly lists `implementationRef` as a required field on the member model, and frames the lifecycle as `build → bind`:

> `Bind` should mean: publish the current member revision to that member's service.

In this PR:
- `StudioMemberState.implementation_ref` is on the proto.
- `StudioMemberImplementationUpdatedEvent` and `HandleImplementationUpdated` exist.
- `IStudioMemberCommandPort.UpdateImplementationAsync` and `ActorDispatchStudioMemberCommandService.UpdateImplementationAsync` exist.

But **no service / HTTP path ever calls `UpdateImplementationAsync`**. `StudioMemberService.BindAsync` accepts the workflow yaml / script id / actor type *in the bind request body* and forwards it directly to `IScopeBindingCommandPort` — without first persisting it on the member authority via `UpdateImplementationAsync`.

Consequences:
- `member.implementation_ref` is always empty. The denormalized `state_root` inside `StudioMemberCurrentStateDocument` reflects this — the read model has no idea what the member is implemented by.
- `LifecycleStage` skips `BuildReady` entirely: `ApplyBound` sets `BindReady` regardless of current stage, so members go `Created → BindReady` directly. The three-stage lifecycle in the proto and in the spec is collapsed to two stages by the actual flow.
- The bind flow violates the issue’s mental model: it conflates *Build* (declare the implementation revision) and *Bind* (publish it to my service). The correct model is that bind reads the current revision off the member, not from the request body.

This is a scope/architecture gap, not a small polish issue. Two ways to close it:
- Add a separate `PUT /members/{memberId}/implementation` (or fold it into `POST /members` + later updates) that calls `UpdateImplementationAsync`, and have `BindAsync` *use* the persisted `implementation_ref` instead of the request body. This matches issue #325.
- Or, keep the current single-shot bind but include a `RecordImplementationAsync` step inside `BindAsync` so the member state is consistent with the published revision after bind. Less clean, but at least the actor stops lying.

Either way, today the actor’s `implementation_ref` field and the `BuildReady` stage are dead surface that the read model still reports.

### 2. Dead `IStudioMemberQueryPort` dependency in the command service

`ActorDispatchStudioMemberCommandService` injects `IStudioMemberQueryPort _queryPort` (assigned in the ctor) but never reads it. Per CLAUDE.md *“删除优先：空转发、重复抽象、无业务价值代码直接删除，不保留兼容空壳”* — drop the field + ctor parameter, or actually use it (e.g. for the duplicate-create check).

### 3. `UpdateImplementationAsync` and `StudioMemberRenamedEvent` are unused public surface

Same “删除优先” point. `UpdateImplementationAsync` is on the command port + implementation but no caller. `StudioMemberRenamedEvent` is handled by the actor and projector but no service / endpoint emits it. If they’re meant for the next PR, fine — but right now they are uncovered surface that drift if not exercised. Either wire them in this PR (preferred — see point 1) or drop them and add later when needed.

### 4. `state_root` Any-pack of full actor state in the read-model document

`StudioMemberCurrentStateDocument.state_root: Any` packs `StudioMemberState` directly, and `ProjectionStudioMemberQueryPort.GetAsync` unpacks it to read `implementation_ref` and `last_binding`. CLAUDE.md is explicit:

> 状态镜像契约面向查询：state mirror payload 作为 readmodel 输入时须是面向读侧的稳定强类型契约，非 actor 内部 state 的原样 dump.

`UserConfigCurrentStateDocument` (also Studio) is the right pattern — fully denormalized, no `state_root`. The new doc here should denormalize `implementation_ref` (e.g. `implementation_workflow_id`, `implementation_script_id`, `implementation_actor_type_name` + `implementation_workflow_revision`/`script_revision`) and `last_binding` (`last_bound_at`, the kind, etc.) — the partial denormalization (`last_bound_revision_id` only) is the worst of both worlds because it forces the query port to peek at the actor’s internal state. I get that other Studio docs (`GAgentRegistryCurrentStateDocument`, `ConnectorCatalogCurrentStateDocument`) already use the `Any state_root` pattern; this PR is a chance to not double down on it for a brand-new doc.

### 5. Two-phase bind has no compensation

`StudioMemberService.BindAsync`:
1. `await _scopeBindingCommandPort.UpsertAsync(...)` — creates the revision upstream
2. `await _memberCommandPort.RecordBindingAsync(...)` — records on member

If step 2 throws, the platform-side service has a fresh revision but the member authority never observes the binding. Next bind will create yet another upstream revision and only that one will be recorded. Acceptable as “last write wins,” but at least add a comment, or compensate by having `BindAsync` derive `last_binding` lazily from the platform-side read model on read. Worth thinking about before this is the load-bearing path.

### 6. Smaller items

- `MemberImplementationKindMapper.ToWireName(StudioMemberLifecycleStage.Unspecified)` returns `"created"`. Silently lying about the type — should return empty or throw.
- `ProjectionStudioMemberQueryPort.ListAsync` hardcodes `Take = DefaultRosterPageSize (200)` with no paging. Silently truncates large rosters; either expose paging or document the limit.
- `StudioMemberConventions.NormalizeMemberId` doesn’t reject `:`. The comment in `ActorDispatchStudioMemberCommandService.GenerateMemberId` claims the format is “free of separators that StudioMemberConventions builds with (':')”, but a user-supplied memberId via `CreateStudioMemberRequest.MemberId` can contain `:` — making `studio-member:scope:m:nested:bad` ambiguous. Either reject `:` (and any other unsafe chars) in `NormalizeMemberId`, or stop accepting user-supplied ids and always generate them.
- `ApplyRenamed` overwrites `Description` with `evt.Description` unconditionally — empty string in the rename event wipes existing description. Likely fine (rename payload always carries the full description), but worth a comment or a typed “did the caller intend to clear?” field. Low priority since rename has no caller yet (see point 3).
- `HandleCreated` idempotency: same `MemberId` returns silently even if `displayName` / `implementationKind` / `description` differ from the existing ones. That’s probably the right call (don’t let a duplicate POST mutate stable identity), but worth surfacing as “first-write wins” — currently it’s implicit.

src/Aevatar.Studio.Projection/ReadModels/studio_projection_readmodels.proto
```diff
@@ -100,3 +100,30 @@ message ChatConversationCurrentStateDocument {
...
+// used by query ports to list members per scope; published_service_id is the
+// stable per-member service id used by member binding orchestration.
+
+message StudioMemberCurrentStateDocument {
+  string id = 1;
+  string actor_id = 2;
+  int64 state_version = 3;
+  string last_event_id = 4;
+  google.protobuf.Timestamp updated_at = 5;
+  google.protobuf.Any state_root = 10;
```
eanzhao: `state_root: Any` is the *actor’s internal state dumped as-is*, and `ProjectionStudioMemberQueryPort.GetAsync` unpacks it to read `implementation_ref` and `last_binding`. CLAUDE.md *状态镜像契约面向查询*: the read-model payload should be a query-shaped contract, not the actor’s internal state.

`UserConfigCurrentStateDocument` in this same proto file is the right pattern — fully denormalized, no `state_root`. Suggest doing the same here: drop `state_root`, denormalize `last_bound_at`, `last_bound_implementation_kind`, and the three implementation-ref shapes (`workflow_id` + `workflow_revision` / `script_id` + `script_revision` / `actor_type_name`). The current partial denormalization (only `last_bound_revision_id`) is the worst of both worlds because it forces the query port to crack open the internal state shape.

src/Aevatar.Studio.Application/Studio/Contracts/MemberImplementationKindMapper.cs
```diff
@@ -0,0 +1,45 @@
...
+        StudioMemberImplementationKind.Gagent => MemberImplementationKindNames.GAgent,
+        _ => string.Empty,
+    };
+
+    public static string ToWireName(StudioMemberLifecycleStage stage) => stage switch
+    {
+        StudioMemberLifecycleStage.Created => MemberLifecycleStageNames.Created,
+        StudioMemberLifecycleStage.BuildReady => MemberLifecycleStageNames.BuildReady,
+        StudioMemberLifecycleStage.BindReady => MemberLifecycleStageNames.BindReady,
+        _ => MemberLifecycleStageNames.Created,
```
eanzhao: Silent type lie: `ToWireName(StudioMemberLifecycleStage.Unspecified)` returns `"created"`. If `Unspecified` ever ends up in the document, callers will see a member in `created` stage when in fact the projection is malformed. Should either return an empty string or throw. Same caution for the kind mapper if a value drifts.

src/Aevatar.Studio.Projection/QueryPorts/ProjectionStudioMemberQueryPort.cs
```diff
@@ -0,0 +1,157 @@
...
+            [
+                new ProjectionDocumentFilter
+                {
+                    FieldPath = "scope_id",
+                    Operator = ProjectionDocumentFilterOperator.Eq,
+                    Value = ProjectionDocumentValue.FromString(normalizedScopeId),
+                },
+            ],
+            Take = DefaultRosterPageSize,
+        };
```
eanzhao, line 56: Hardcoded `Take = DefaultRosterPageSize (200)` with no paging. Silently truncates if a scope ever has > 200 members — Studio frontend would just stop seeing some members with no signal. At minimum surface a `nextPageToken` or document the limit on the contract; preferably accept a paging cursor on `ListAsync`.

src/Aevatar.Studio.Projection/QueryPorts/ProjectionStudioMemberQueryPort.cs
```diff
@@ -0,0 +1,157 @@
...
+
+    private static StudioMemberDetailResponse ToDetail(StudioMemberCurrentStateDocument document)
+    {
+        var summary = ToSummary(document);
+        StudioMemberImplementationRefResponse? implementationRef = null;
+        StudioMemberBindingContractResponse? lastBinding = null;
+
+        if (document.StateRoot != null && document.StateRoot.Is(StudioMemberState.Descriptor))
+        {
+            var state = document.StateRoot.Unpack<StudioMemberState>();
```
eanzhao: Reading internal actor state from the read-model document. Tied to the `state_root` Any-pack issue on the proto: the query port should not need `Unpack<StudioMemberState>()`. Once `implementation_ref` and `last_binding` are denormalized into the document, this whole branch goes away.

agents/Aevatar.GAgents.StudioMember/StudioMemberConventions.cs
```diff
@@ -0,0 +1,54 @@
...
+        return trimmed;
+    }
+
+    public static string NormalizeMemberId(string? memberId)
+    {
+        var trimmed = memberId?.Trim();
+        if (string.IsNullOrEmpty(trimmed))
+            throw new ArgumentException("memberId is required.", nameof(memberId));
+        return trimmed;
+    }
```
eanzhao, line 59: `NormalizeMemberId` doesn’t reject `:`. `ActorDispatchStudioMemberCommandService.GenerateMemberId` says the generated format is *“URL-safe and free of separators that StudioMemberConventions builds with (':')”*, but `CreateStudioMemberRequest.MemberId` lets a caller supply an arbitrary id, which goes through this normalize and back into `BuildActorId`. A memberId of `m:nested` produces actor id `studio-member:scope-1:m:nested` — ambiguous if anything ever parses it, and bypasses the immutable-id intent. Either reject `:` (and other unsafe chars) here, or stop accepting user-supplied ids and always generate them in the command service.

agents/Aevatar.GAgents.StudioMember/StudioMemberGAgent.cs
```diff
@@ -0,0 +1,162 @@
...
+            // Idempotent re-create with same identity is a no-op; otherwise reject
+            // so a stray duplicate cannot overwrite an existing member's
+            // publishedServiceId or scope binding.
+            if (!string.Equals(State.MemberId, evt.MemberId, StringComparison.Ordinal))
+            {
+                throw new InvalidOperationException(
+                    $"member already initialized with id '{State.MemberId}'.");
+            }
+
+            return;
```
eanzhao, line 52: Idempotent re-create: same `member_id` returns silently even if `display_name` / `implementation_kind` / `description` in the new event differ from the persisted ones. Probably the right call (don’t let a duplicate POST mutate stable identity), but it’s a *first-write-wins* policy and right now it’s implicit. Worth either a comment, or making the mismatch on the *non-identity* fields throw the same way `member_id` mismatch does — otherwise the silent ignore can mask a real bug client-side.

eanzhao (commented):
## Follow-up review — additional issues

Further pass after the first review surfaced more concerns. Inline comments below; brief grouping here:

### A. New layering coupling: `Aevatar.Studio.Application` now project-refs `Aevatar.GAgents.StudioMember`

This is the first agent package that `Aevatar.Studio.Application` takes a `<ProjectReference>` on. Compare to `UserConfig` / `RoleCatalog` / `ConnectorCatalog` — those agent packages are referenced only by `Aevatar.Studio.Projection`, not by Application. The Application layer there receives Application-shaped DTOs and the Projection layer maps them onto the proto event types.

In this PR, `Aevatar.Studio.Application/Studio/Contracts/MemberImplementationKindMapper.cs` imports `Aevatar.GAgents.StudioMember` and uses `StudioMemberImplementationKind` / `StudioMemberLifecycleStage` (proto-generated enums) directly. CLAUDE.md *上层依赖抽象，禁止…对具体实现的直接耦合* — Application should be agnostic of the proto representation; that representation is an Agent / boundary concern.

Suggested shape:
- Define `MemberImplementationKind` / `MemberLifecycleStage` as Application-owned enums (or keep them as the existing wire-string constants).
- Move the proto-enum mapper to the Projection layer (next to `ActorDispatchStudioMemberCommandService` and `ProjectionStudioMemberQueryPort`).
- Drop the agent ProjectReference from `Aevatar.Studio.Application.csproj`.

This is a structural item, not a one-line fix — but worth resolving now while the surface is small. Once anything else lands on top of this, ripping it out gets harder.

### B. HTTP status semantics: "member not found" surfaces as 400, not 404

`StudioMemberService.BindAsync` throws `InvalidOperationException` when the member doesn’t exist, and `StudioMemberEndpoints.HandleBindAsync` catches `InvalidOperationException` → 400 BadRequest with `INVALID_STUDIO_MEMBER_BINDING`. This conflates *missing resource* with *bad input*. Studio frontend can’t reliably distinguish “user typo” from “real 4xx”. Same shape for `GetBindingAsync` and (less obviously) the validation throws. Consider:
- A typed `MemberNotFoundException` (or `KeyNotFoundException`) mapped to 404 in the endpoint.
- Reserve 400 / `INVALID_STUDIO_MEMBER_BINDING` for actual bad-input cases (missing yaml, wrong kind etc.).

### C. Defensive-actor: `evt.PublishedServiceId` is trusted blind

`StudioMemberGAgent.HandleCreated` persists whatever `evt.PublishedServiceId` came in. The dispatcher in `ActorDispatchStudioMemberCommandService.CreateAsync` derives it via `StudioMemberConventions.BuildPublishedServiceId(memberId)`, but the actor itself doesn’t enforce that invariant. A future caller dispatching the event directly (or replaying an old event with a stale derivation rule) could persist a `published_service_id` that no longer matches `BuildPublishedServiceId(memberId)`. Cheap fix: in `ApplyCreated`, rebuild from `memberId` and discard `evt.PublishedServiceId`, so the convention stays a single source of truth.

### D. Unchecked enum casts in the query port

`ProjectionStudioMemberQueryPort.ToSummary` casts the int columns straight to enum:

```csharp
var implementationKind = (StudioMemberImplementationKind)document.ImplementationKind;
var lifecycleStage = (StudioMemberLifecycleStage)document.LifecycleStage;
```

If the document was projected by a future build that knows new enum values, or by a corrupt write, the cast silently produces an out-of-range enum and downstream `ToWireName` returns `""` (which is at least no longer the silent `"created"` lie after fix #2). Worth either an `Enum.IsDefined` guard with a logged warning, or storing the wire string directly in the document so the projection contract is the single source of truth.

### E. `StateRoot.Is(StudioMemberState.Descriptor)` failure is silent

When `state_root` is null or carries a different descriptor, `ToDetail` silently drops `implementation_ref` and `last_binding` from the response — without distinguishing “doesn’t exist yet” from “projection is malformed.” Combined with the still-open *state-mirror dump* concern from review #1, this is the second-order cost of leaning on `Any state_root` for derived fields. Until the document is fully denormalized, at least surface a structured warning when the descriptor mismatches; a silent partial response makes downstream debugging painful.

### F. Unused `using` in StudioMemberService.cs

`Aevatar.GAgents.StudioMember` is imported but no symbol from it is referenced in the file. Drop it (and once point A is addressed it stops being importable from this layer at all).

### G. No length / character bounds on `displayName`, `description`, user-supplied `memberId`

`CreateStudioMemberRequest` accepts arbitrary content. A 10MB displayName persists end-to-end (actor state → projection → ES doc → API response). Doesn’t take much to be a soft attack surface or a noisy-neighbor bug. Add minimal length caps + character whitelist on the create path. (Memberid `:` rejection is now in place after fix #3 — extending it to other unsafe chars and a max length is the next step.)

src/Aevatar.Studio.Application/Aevatar.Studio.Application.csproj
```diff
@@ -18,6 +18,7 @@
     <ProjectReference Include="..\Aevatar.Scripting.Infrastructure\Aevatar.Scripting.Infrastructure.csproj" />
     <ProjectReference Include="..\workflow\Aevatar.Workflow.Application.Abstractions\Aevatar.Workflow.Application.Abstractions.csproj" />
     <ProjectReference Include="..\workflow\Aevatar.Workflow.Core\Aevatar.Workflow.Core.csproj" />
+    <ProjectReference Include="..\..\agents\Aevatar.GAgents.StudioMember\Aevatar.GAgents.StudioMember.csproj" />
```
eanzhao: **New layering coupling**: this is the first time `Aevatar.Studio.Application` takes a `<ProjectReference>` on an agent package. UserConfig, RoleCatalog, ConnectorCatalog all live behind `Aevatar.Studio.Projection`. Pulling `Aevatar.GAgents.StudioMember` into Application means the Application contract now depends on the proto-generated enums (`StudioMemberImplementationKind`, `StudioMemberLifecycleStage`) — see `MemberImplementationKindMapper.cs`.

CLAUDE.md *严格分层 / 上层依赖抽象，禁止…对具体实现的直接耦合*. Suggest: move the proto-enum mapper to the Projection layer (next to `ActorDispatchStudioMemberCommandService`), have Application stay on the wire-string constants in `MemberImplementationKindNames`, and drop this ProjectReference.

src/Aevatar.Studio.Projection/Mapping/MemberImplementationKindMapper.cs
```diff
@@ -0,0 +1,45 @@
+using Aevatar.GAgents.StudioMember;
```
eanzhao, line 1: Tied to the layering comment on the csproj — this mapper is the only consumer of `Aevatar.GAgents.StudioMember.*` in Application. Easiest fix: move the file to `Aevatar.Studio.Projection/Adapters/`, leave the wire-string constants (`MemberImplementationKindNames`) in Application, and have the Projection-side `ActorDispatchStudioMemberCommandService` and `ProjectionStudioMemberQueryPort` use the moved mapper.

src/Aevatar.Studio.Application/Studio/Services/StudioMemberService.cs
```diff
@@ -0,0 +1,216 @@
+using Aevatar.GAgentService.Abstractions;
+using Aevatar.GAgentService.Abstractions.Ports;
+using Aevatar.GAgents.StudioMember;
```
eanzhao: Unused `using` — no symbol from `Aevatar.GAgents.StudioMember` is referenced in this file. Remove.

src/Aevatar.Studio.Application/Studio/Services/StudioMemberService.cs
```diff
@@ -0,0 +1,216 @@
...
+    public async Task<StudioMemberBindingResponse> BindAsync(
+        string scopeId,
+        string memberId,
+        UpdateStudioMemberBindingRequest request,
+        CancellationToken ct = default)
+    {
+        ArgumentNullException.ThrowIfNull(request);
+
+        var detail = await _memberQueryPort.GetAsync(scopeId, memberId, ct)
+            ?? throw new InvalidOperationException(
```
eanzhao: Wrong HTTP semantics. `member not found` is a 404, not a 400. The endpoint catches `InvalidOperationException` → 400 with code `INVALID_STUDIO_MEMBER_BINDING`, which conflates *missing resource* with *bad input*. Throw a typed `MemberNotFoundException` (or `KeyNotFoundException`) and map it to 404 in `StudioMemberEndpoints.HandleBindAsync` / `HandleGetBindingAsync`.

agents/Aevatar.GAgents.StudioMember/StudioMemberGAgent.cs
```diff
@@ -0,0 +1,162 @@
...
+    {
+        return new StudioMemberState
+        {
+            MemberId = evt.MemberId,
+            ScopeId = evt.ScopeId,
+            DisplayName = evt.DisplayName,
+            Description = evt.Description,
+            ImplementationKind = evt.ImplementationKind,
+            ImplementationRef = null,
+            PublishedServiceId = evt.PublishedServiceId,
```
eanzhao: Defensive-actor invariant: `evt.PublishedServiceId` is trusted blind here. The dispatcher derives it via `StudioMemberConventions.BuildPublishedServiceId(memberId)`, but the actor doesn’t re-derive or assert. A future caller (or a replayed historical event from a different derivation rule) could persist a `published_service_id` that no longer matches the convention.

Cheap fix: rebuild it inside `ApplyCreated` and ignore `evt.PublishedServiceId`, so the convention stays single-source. The event field becomes documentation rather than load-bearing.

src/Aevatar.Studio.Projection/QueryPorts/ProjectionStudioMemberQueryPort.cs
```diff
@@ -0,0 +1,157 @@
...
+        if (!string.Equals(document.ScopeId, normalizedScopeId, StringComparison.Ordinal))
+            return null;
+
+        return ToDetail(document);
+    }
+
+    private static StudioMemberSummaryResponse ToSummary(StudioMemberCurrentStateDocument document)
+    {
+        var implementationKind = (StudioMemberImplementationKind)document.ImplementationKind;
+        var lifecycleStage = (StudioMemberLifecycleStage)document.LifecycleStage;
```
eanzhao: Unchecked int → enum casts. If a stale projection wrote a value outside the enum (corrupt index, future-version writer, etc.), the cast silently produces an out-of-range enum and `ToWireName` returns `""`. Either guard with `Enum.IsDefined` and log, or — cleaner — store the wire string in the document and drop the int columns entirely.

src/Aevatar.Studio.Projection/QueryPorts/ProjectionStudioMemberQueryPort.cs
```diff
@@ -0,0 +1,157 @@
...
+            UpdatedAt: document.UpdatedAt?.ToDateTimeOffset() ?? DateTimeOffset.MinValue);
+    }
+
+    private static StudioMemberDetailResponse ToDetail(StudioMemberCurrentStateDocument document)
+    {
+        var summary = ToSummary(document);
+        StudioMemberImplementationRefResponse? implementationRef = null;
+        StudioMemberBindingContractResponse? lastBinding = null;
+
+        if (document.StateRoot != null && document.StateRoot.Is(StudioMemberState.Descriptor))
```
eanzhao: Silent fall-through: if `StateRoot` is null or carries a non-`StudioMemberState` descriptor, the response loses `implementation_ref` + `last_binding` with no signal. 

Combined with the still-open state-mirror-dump concern from review #1, this is the failure mode of leaning on `Any state_root` for derived fields. Until denormalization happens, at least log a structured warning when the descriptor mismatches — a silent partial response makes downstream debugging painful.

src/Aevatar.Studio.Hosting/Endpoints/StudioMemberEndpoints.cs
```diff
@@ -0,0 +1,158 @@
...
+        {
+            return Results.Ok(await memberService.BindAsync(scopeId, memberId, request, ct));
+        }
+        catch (InvalidOperationException ex)
+        {
+            return Results.BadRequest(new
+            {
+                code = "INVALID_STUDIO_MEMBER_BINDING",
+                message = ex.Message,
+            });
```
eanzhao: Catch-all for `InvalidOperationException` → 400. Combined with the “not found” throw on the service side, this returns 400 `INVALID_STUDIO_MEMBER_BINDING` for both *missing member* and *bad bind payload*. Studio frontend can’t reliably distinguish them. Differentiate by exception type and map missing-member to 404.

src/Aevatar.Studio.Application/Studio/Contracts/MemberContracts.cs
```diff
@@ -0,0 +1,105 @@
...
+
+public sealed record StudioMemberRosterResponse(
+    string ScopeId,
+    IReadOnlyList<StudioMemberSummaryResponse> Members);
+
+public sealed record CreateStudioMemberRequest(
+    string DisplayName,
+    string ImplementationKind,
+    string? Description = null,
+    string? MemberId = null);
```
eanzhao, line 86: No bounds on `DisplayName`, `Description`, or `MemberId`. A 10MB displayName will persist all the way through (actor state → projection → ES doc → API response). Add length caps + character validation at the create path. With the `:` rejection in `NormalizeMemberId` now in place, extend to a stricter slug whitelist for `MemberId` (e.g. `^[A-Za-z0-9_-]{1,64}$`) and a sane max length on `DisplayName` / `Description`.

src/Aevatar.Studio.Application/Studio/Abstractions/StudioMemberNotFoundException.cs
```diff
@@ -0,0 +1,20 @@
+namespace Aevatar.Studio.Application.Studio.Abstractions;
+
+/// <summary>
+/// Thrown when a member-centric operation targets an id that has no
+/// corresponding member document in the requested scope. Endpoints map this
+/// to HTTP 404 — distinct from validation errors that map to 400.
+/// </summary>
+public sealed class StudioMemberNotFoundException : InvalidOperationException
```
eanzhao: Subclassing `InvalidOperationException` makes the 404 / 400 distinction depend on catch ordering in every endpoint. A reorder turns 404 into 400 silently. Plus, `BindAsync_ShouldFail_WhenMemberDoesNotExist` asserts `ThrowAsync<InvalidOperationException>()` and so passes regardless of whether the typed `StudioMemberNotFoundException` is preserved.

Suggest base on `Exception` (or `KeyNotFoundException`) and update the binding test to assert on `StudioMemberNotFoundException` directly.

test/Aevatar.Studio.Tests/StudioMemberServiceBindingTests.cs
```diff
@@ -0,0 +1,290 @@
...
+
+        // The orchestrator records the resulting revision back on the
+        // member authority so /members/.../binding can read it from the
+        // read model later.
+        commandPort.RecordedBindings.Should().ContainSingle()
+            .Which.PublishedServiceId.Should().Be(PublishedServiceId);
+        response.PublishedServiceId.Should().Be(PublishedServiceId);
+        response.RevisionId.Should().Be(bindingPort.IssuedRevisionId);
+        response.ImplementationKind.Should().Be(MemberImplementationKindNames.Workflow);
+    }
```
eanzhao, line 70: Coverage gap: the new bind-lifecycle wiring (`UpdateImplementationAsync` → `RecordBindingAsync`) is the headline fix, but it’s not asserted here. Dropping the `UpdateImplementationAsync` call from `BindAsync` would still leave every test in this file green.

`RecordedImplementationUpdates` is captured on the stub (line 232) — assert on it: exactly one entry, the right kind + ref shape, and recorded **before** the `RecordedBindings` entry so the `BuildReady → BindReady` order is locked in. Apply the same to the script / gagent tests.

agents/Aevatar.GAgents.StudioMember/StudioMemberGAgent.cs
```diff
@@ -0,0 +1,202 @@
...
+        next.Description = evt.Description;
+        next.UpdatedAtUtc = evt.UpdatedAtUtc;
+        return next;
+    }
+
+    private static StudioMemberState ApplyImplementationUpdated(
+        StudioMemberState state, StudioMemberImplementationUpdatedEvent evt)
+    {
+        var next = state.Clone();
+        next.ImplementationKind = evt.ImplementationKind;
```
eanzhao: `ImplementationKind` drift post-create.

`HandleCreated` now rejects re-create with a different kind (good), but `ApplyImplementationUpdated` overwrites `State.ImplementationKind` with `evt.ImplementationKind` unconditionally — so the kind invariant the create event was supposed to lock can still be silently violated by dispatching an `StudioMemberImplementationUpdatedEvent { ImplementationKind = …different… }`.

Either drop `ImplementationKind` from the implementation-updated event (use `State.ImplementationKind`), or have `HandleImplementationUpdated` reject when `evt.ImplementationKind != State.ImplementationKind`. The bind orchestration doesn’t mutate kind today, but the actor protocol allows it.

src/Aevatar.Studio.Hosting/Endpoints/StudioMemberEndpoints.cs
```diff
@@ -0,0 +1,173 @@
...
+        IStudioMemberService memberService,
+        CancellationToken ct)
+    {
+        if (AevatarScopeAccessGuard.TryCreateScopeAccessDeniedResult(http, scopeId, out var denied))
+            return denied;
+
+        try
+        {
+            var binding = await memberService.GetBindingAsync(scopeId, memberId, ct);
+            return binding == null ? Results.NotFound() : Results.Ok(binding);
```
eanzhao: Two different 404 shapes for the same endpoint. After the `StudioMemberNotFoundException` change, the `binding == null` branch only fires when the member exists but has never been bound. So callers see:
- `404 { code: STUDIO_MEMBER_NOT_FOUND, … }` for unknown member
- bare `404` (empty body) for member-exists-no-binding

Frontend can’t distinguish them from status code alone, and “exists but never bound” is arguably not a missing-resource case at all. Suggest 200 with `lastBinding: null`, or a typed `STUDIO_MEMBER_NOT_BOUND_YET` body — either way, distinguishable.

src/Aevatar.Studio.Hosting/Endpoints/StudioMemberEndpoints.cs
```diff
@@ -0,0 +1,173 @@
...
+    {
+        if (AevatarScopeAccessGuard.TryCreateScopeAccessDeniedResult(http, scopeId, out var denied))
+            return denied;
+
+        try
+        {
+            var summary = await memberService.CreateAsync(scopeId, request, ct);
+            return Results.Created($"/api/scopes/{scopeId}/members/{summary.MemberId}", summary);
+        }
+        catch (StudioMemberNotFoundException ex)
```
eanzhao: Dead catch — `CreateAsync` cannot legitimately throw `StudioMemberNotFoundException` (there is no member to be missing yet). Remove. The corresponding catches on the bind / get-binding handlers are load-bearing and should stay.

src/Aevatar.Studio.Application/Studio/Services/StudioMemberService.cs
```diff
@@ -0,0 +1,283 @@
...
+            ?? throw new ArgumentNullException(nameof(scopeBindingCommandPort));
+    }
+
+    public Task<StudioMemberSummaryResponse> CreateAsync(
+        string scopeId,
+        CreateStudioMemberRequest request,
+        CancellationToken ct = default)
+    {
+        ArgumentNullException.ThrowIfNull(request);
+        return _memberCommandPort.CreateAsync(scopeId, request, ct);
```
eanzhao, line 57: Validation lives in the wrong layer.

Length caps + slug pattern live in `ActorDispatchStudioMemberCommandService.ValidateUserSuppliedMemberId` and inline in the Projection-layer command service. If anyone swaps the command port — alternate impl, in-memory test variant, migration tool — the bounds disappear silently.

`StudioMemberInputLimits` is already in Application (`MemberContracts.cs`). Lift the actual `displayName.Length > MaxDisplayNameLength` / `description.Length > MaxDescriptionLength` / slug-regex checks here in `CreateAsync`, and let the command port trust its input.

src/Aevatar.Studio.Application/Studio/Contracts/MemberContracts.cs
```diff
@@ -0,0 +1,127 @@
...
+    DateTimeOffset BoundAt);
+
+public sealed record StudioMemberRosterResponse(
+    string ScopeId,
+    IReadOnlyList<StudioMemberSummaryResponse> Members,
+    string? NextPageToken = null);
+
+public sealed record StudioMemberRosterPageRequest(
+    int? PageSize = null,
+    string? Cursor = null);
```
eanzhao: Naming inconsistency: HTTP query param is `pageToken`, response field is `NextPageToken`, internal DTO field is `Cursor`. Pick one — `PageToken` for the DTO is probably the lower-impact rename.

eanzhao (commented):
## Follow-up review on the fix commit

Looked through `03a51d7a` end-to-end. The fix is solid:
- Lifecycle gap closed (`BindAsync` now does `UpdateImplementation → RecordBinding`, projector / actor reach `BuildReady` and `BindReady` legitimately).
- `state_root` `Any` is gone, document is fully denormalized; query port no longer peeks inside actor state.
- Convention single-source: `ApplyCreated` re-derives `publishedServiceId` from `BuildPublishedServiceId(memberId)`.
- Layering restored: `Aevatar.Studio.Application.csproj` no longer ProjectRefs the agent package; `MemberImplementationKindMapper` lives in `Aevatar.Studio.Projection.Mapping`.
- Input bounds + slug pattern + 404 semantics all in place; pagination wired through `pageSize` / `pageToken`.
- `HandleCreated` now rejects mismatched non-identity fields (first-write-wins on identity).
- 287 Studio tests pass.

A handful of residual items below — none of them are blockers, but worth landing or at least filing.

### 1. `StudioMemberNotFoundException : InvalidOperationException` is brittle

The new exception subclasses `InvalidOperationException`. That works *only because* every endpoint puts the `StudioMemberNotFoundException` catch ahead of the `InvalidOperationException` catch:

```csharp
catch (StudioMemberNotFoundException ex) { return NotFound(ex); }
catch (InvalidOperationException ex) { return BadRequest(...); }
```

Reorder those by accident in a future refactor and the 404 silently downgrades to a 400 with `INVALID_STUDIO_MEMBER_BINDING`. Same brittleness shows up in tests: `BindAsync_ShouldFail_WhenMemberDoesNotExist` asserts `ThrowAsync<InvalidOperationException>()` (line 134), so it passes regardless of whether the throw is the typed `StudioMemberNotFoundException` or any random IOEx (e.g. the `'has no publishedServiceId'` invariant violation). The test wouldn’t catch a regression that swapped the typed exception for a plain throw.

Suggest one of:
- Subclass `Exception` (or `KeyNotFoundException`) instead of `InvalidOperationException`, and update the test to `ThrowAsync<StudioMemberNotFoundException>()`.
- Or keep the inheritance but tighten the test.

### 2. The bind-lifecycle wiring is uncovered by `StudioMemberServiceBindingTests`

This is the main fix on the PR — `BindAsync` calls `UpdateImplementationAsync` between the scope upsert and the `RecordBinding` so the actor walks `Created → BuildReady → BindReady`. The recording stub captures it (`RecordedImplementationUpdates`, line 232) but **no test asserts on it**. The state-machine half of the fix is locked in by `StudioMemberPRReviewFixesTests`, but the orchestration half can be silently undone — drop the `UpdateImplementationAsync` call from `BindAsync` and every existing test in `StudioMemberServiceBindingTests` still passes.

Fix: in each of `BindAsync_Workflow…` / `…_Script…` / `…_GAgent…`, assert `commandPort.RecordedImplementationUpdates` contains exactly one entry with the right kind + ref shape, and assert it’s recorded **before** `RecordedBindings` (so the lifecycle order is locked in too).

### 3. `ApplyImplementationUpdated` lets `ImplementationKind` drift post-create

```csharp
next.ImplementationKind = evt.ImplementationKind;
next.ImplementationRef = evt.ImplementationRef?.Clone();
```

`HandleCreated` now rejects mismatched-kind re-creates (good), but `HandleImplementationUpdated` happily overwrites `State.ImplementationKind` with whatever the event says. So a script-kind member can be silently mutated into a workflow-kind member by dispatching `StudioMemberImplementationUpdatedEvent { ImplementationKind = Workflow, … }`. The bind orchestration today never does this, but the actor protocol allows it.

Either:
- Drop `ImplementationKind` from `StudioMemberImplementationUpdatedEvent` (use `State.ImplementationKind`), or
- Have `HandleImplementationUpdated` reject when `evt.ImplementationKind != State.ImplementationKind`.

The second is safer if there’s any chance the event field has external consumers; the first is cleaner.

### 4. `HandleGetBindingAsync` returns 404 with two different shapes

```csharp
var binding = await memberService.GetBindingAsync(scopeId, memberId, ct);
return binding == null ? Results.NotFound() : Results.Ok(binding);
```

Now that `GetBindingAsync` throws `StudioMemberNotFoundException` for missing members, the **bare** `Results.NotFound()` only fires for *member exists, never bound*. So callers see two 404 shapes: `{ code: STUDIO_MEMBER_NOT_FOUND, … }` for unknown member vs. empty 404 body for unbound member. Frontend can’t distinguish them from status code alone.

“Member exists but never bound” isn’t a missing resource; consider returning `200 { lastBinding: null }` (the request *is* satisfied) or a typed `STUDIO_MEMBER_NOT_BOUND_YET` 404 body. Either way, distinguishable.

### 5. `StudioMemberService.CreateAsync` does no validation; it pass-throughs

Length caps + slug pattern live inside `ActorDispatchStudioMemberCommandService.ValidateUserSuppliedMemberId` (Projection layer). If anyone swaps the command port — for tests, for an in-memory variant, for a future migration — those bounds disappear silently. Application is the natural home for input shape validation; today it’s only wiring.

Minimal fix: lift `ValidateUserSuppliedMemberId`, the displayName cap, and the description cap to `StudioMemberInputLimits` (already in Application) and have `StudioMemberService.CreateAsync` call them before delegating. The command port can then trust the input it receives.

### 6. `HandleCreateAsync` has a dead `catch (StudioMemberNotFoundException)`

`CreateAsync` cannot legitimately throw `StudioMemberNotFoundException` — there’s no member to be missing yet. Delete the catch in `HandleCreateAsync` (line 55–58). Same comment isn’t a problem on the bind / get-binding handlers, where the catch is load-bearing.

### 7. Naming: `StudioMemberRosterPageRequest.Cursor` vs HTTP `pageToken`

Minor — the public HTTP contract is `pageToken` (and the response field is `NextPageToken`), the internal request DTO is `Cursor`. Either the contract field renames to `Cursor` or the DTO renames to `PageToken`. Pick one to keep search-grepability sane.

eanzhao (commented):
Latest head fixes several earlier items. I left two additional comments on actor-level invariants that are still enforceable through direct command/event paths.

agents/Aevatar.GAgents.StudioMember/StudioMemberGAgent.cs
```diff
@@ -0,0 +1,217 @@
...
+        if (string.IsNullOrEmpty(State.MemberId))
+        {
+            throw new InvalidOperationException("member not yet created.");
+        }
+
+        // ImplementationKind is locked at create. Reject mismatched kinds so
+        // a Script member can't be silently mutated into a Workflow member by
+        // dispatching an UpdatedEvent with a different kind. Unspecified is
+        // accepted as "carry the existing kind" (defensive default).
+        if (evt.ImplementationKind != StudioMemberImplementationKind.Unspecified
```
eanzhao, line 81: `ImplementationKind` is now locked, but the payload shape is still unchecked. Because `StudioMemberImplementationRef` is not a `oneof`, a caller can send `ImplementationKind = Unspecified` (or even the matching kind) with the wrong branch populated, e.g. a Script member carrying `Workflow`. `ApplyImplementationUpdated` will clone that ref and the projector can later publish `implementation_kind = "script"` with workflow fields set. Please validate that exactly the branch matching the existing `State.ImplementationKind` is populated before persisting the event.

agents/Aevatar.GAgents.StudioMember/StudioMemberGAgent.cs
```diff
@@ -0,0 +1,217 @@
...
+
+    [EventHandler(EndpointName = "recordBinding")]
+    public async Task HandleBound(StudioMemberBoundEvent evt)
+    {
+        if (string.IsNullOrEmpty(State.MemberId))
+        {
+            throw new InvalidOperationException("member not yet created.");
+        }
+
+        await PersistDomainEventAsync(evt);
```
eanzhao, line 101: `recordBinding` can still be dispatched directly after create and will make the actor `BindReady` with no resolved `ImplementationRef`; `StudioMemberGAgentStateTests.Bound_ShouldCaptureLastBindingAndAdvanceLifecycle` currently locks in that path. The service now sends `UpdateImplementationAsync` before `RecordBindingAsync`, but the actor authority should enforce the invariant too: reject binding unless a matching implementation ref is already present, and also verify `evt.PublishedServiceId == State.PublishedServiceId` and the event kind matches the locked member kind. Otherwise a malformed command can publish a read model that says the member is bind-ready while it has no implementation, or records another service id as this member's binding.

src/Aevatar.Studio.Hosting/Endpoints/StudioMemberEndpoints.cs
```diff
@@ -0,0 +1,169 @@
...
+        IStudioMemberService memberService,
+        CancellationToken ct)
+    {
+        if (AevatarScopeAccessGuard.TryCreateScopeAccessDeniedResult(http, scopeId, out var denied))
+            return denied;
+
+        try
+        {
+            var binding = await memberService.GetBindingAsync(scopeId, memberId, ct);
+            return binding == null ? Results.NotFound() : Results.Ok(binding);
```
eanzhao: Still open from review #3.

Two different 404 shapes for one endpoint:
- `binding == null` → bare `404`, fires only when *member exists but was never bound*
- `catch (StudioMemberNotFoundException)` → typed `404 { code: STUDIO_MEMBER_NOT_FOUND, ... }`, fires when member is missing

Frontend can’t distinguish them from status code alone, and the "exists but never bound" case isn’t really a missing resource. Suggest 200 with `lastBinding: null` (preferred), 204, or a typed `STUDIO_MEMBER_NOT_BOUND_YET` 404 body.

src/Aevatar.Studio.Application/Studio/Services/StudioMemberService.cs
eanzhao, line 57: Still open from review #3.

`StudioMemberInputLimits` lives in Application (`MemberContracts.cs`), but the only callers that *enforce* the bounds live in Projection (`ActorDispatchStudioMemberCommandService.CreateAsync` lines 50/57/210/215). `StudioMemberService.CreateAsync` is a one-line pass-through with no validation. Swap the command port (alt impl / in-memory test / migration tool) and every length cap + slug pattern silently disappears.

Lift the length checks + slug regex up into this method (the constants are already in this layer; it’s a no-import refactor) and let the command port trust the input shape it receives.

src/Aevatar.Studio.Hosting/Endpoints/StudioMemberEndpoints.cs
```diff
@@ -0,0 +1,169 @@
...
+        IStudioMemberService memberService,
+        CancellationToken ct)
+    {
+        if (AevatarScopeAccessGuard.TryCreateScopeAccessDeniedResult(http, scopeId, out var denied))
+            return denied;
+
+        try
+        {
+            var detail = await memberService.GetAsync(scopeId, memberId, ct);
+            return detail == null ? Results.NotFound() : Results.Ok(detail);
```
eanzhao: New finding — inconsistency from the partial fix to review #3 item 4.

`HandleBindAsync` and `HandleGetBindingAsync` both return the typed `STUDIO_MEMBER_NOT_FOUND` body for missing members; `HandleGetAsync` returns a bare `Results.NotFound()` for the same condition. Three endpoints now have three different 404 shapes (this one bare, get-binding has two, bind has one typed). Frontend gets different bodies for the same logical "member not found" depending on which endpoint hit it.

Fix: have `IStudioMemberService.GetAsync` throw `StudioMemberNotFoundException` for missing members (matches `GetBindingAsync` already does this) and add the same typed `catch + NotFound(ex)` here. The `detail == null` branch becomes unreachable and can go.

eanzhao (commented):
## Follow-up review on `c934ec97`

Verified: kind-lock + apply-immutability, `StudioMemberNotFoundException : KeyNotFoundException`, dead catch in `HandleCreateAsync` removed, `Cursor → PageToken`, `OperationsInOrder` ordering assertions on all three bind tests, typed `BindAsync_ShouldFail_WhenMemberDoesNotExist` assertion, plus the new `ApplyImplementationUpdated_ShouldNotMutateImplementationKind`. 288 Studio tests pass.

Two items from review #3 are still not addressed, and the same shape mismatch shows up in a third endpoint that wasn’t flagged before. Inline below.

### 1. (still open) `HandleGetBindingAsync` returns two different 404 shapes

```csharp
var binding = await memberService.GetBindingAsync(scopeId, memberId, ct);
return binding == null ? Results.NotFound() : Results.Ok(binding);   // ← bare 404
...
catch (StudioMemberNotFoundException ex) { return NotFound(ex); }    // ← typed 404
```

After the `StudioMemberNotFoundException` change, the `binding == null` branch only fires when *the member exists but has never been bound*. So callers see:
- `404 { code: STUDIO_MEMBER_NOT_FOUND, scopeId, memberId, message }` for unknown member
- bare `404` (empty body) for member-exists-no-binding

Frontend can’t distinguish them from status code alone, and "exists but never bound" arguably isn’t a missing-resource case at all (the GET succeeded, the member is there, it just doesn’t have a binding yet). Suggestions, in order of preference:
1. `200 OK { lastBinding: null }` — request was satisfied, no binding to return.
2. `204 No Content`.
3. Keep 404 but emit a typed `STUDIO_MEMBER_NOT_BOUND_YET` body so frontends can branch on the code field.

### 2. (still open) Input validation lives in the wrong layer

`StudioMemberInputLimits.MaxDisplayNameLength` / `MaxDescriptionLength` / `MaxMemberIdLength` / `MemberIdPattern` are defined in `Aevatar.Studio.Application/Studio/Contracts/MemberContracts.cs`, but the **only** enforcement lives in `Aevatar.Studio.Projection/CommandServices/ActorDispatchStudioMemberCommandService.cs` (lines 50, 57, 210, 215). `StudioMemberService.CreateAsync` (Application) is a single-line pass-through:

```csharp
public Task<StudioMemberSummaryResponse> CreateAsync(
    string scopeId, CreateStudioMemberRequest request, CancellationToken ct = default)
{
    ArgumentNullException.ThrowIfNull(request);
    return _memberCommandPort.CreateAsync(scopeId, request, ct);
}
```

If anyone swaps the command port — alternate impl, in-memory variant, migration tool — every input bound disappears with no signal. Move the length / slug checks into `StudioMemberService.CreateAsync` (the constants are already in Application, that’s a no-import lift), and let the command port trust its input. Belongs to the layer that owns the Application contract.

### 3. (new) `HandleGetAsync` returns a bare 404 for missing member, inconsistent with bind / get-binding

```csharp
var detail = await memberService.GetAsync(scopeId, memberId, ct);
return detail == null ? Results.NotFound() : Results.Ok(detail);
```

`HandleBindAsync` and `HandleGetBindingAsync` return the typed `STUDIO_MEMBER_NOT_FOUND` body for missing members; `HandleGetAsync` returns a bare 404 for the same condition. Callers get different shapes for the same logical "member not found."

Easiest fix: have `IStudioMemberService.GetAsync` throw `StudioMemberNotFoundException` instead of returning null (matches `GetBindingAsync` already), let the endpoint catch it and call `NotFound(ex)` for a typed body. The `detail == null` branch then becomes unreachable and can be removed.

This also frees up `HandleGetBindingAsync` to use 200/204 for the never-bound case (item 1) without the comparison being lopsided across endpoints.

PR #434, opened 2026-04-27
feat(studio): optimistic concurrency for role/connector catalog drafts

### LINKED ISSUE #432: [Refactor] Studio role/connector catalog drafts need optimistic concurrency control
## Background

Discovered during #251 Q-04 review and confirmed by 2026-04-27 code audit.

## Problem

Role and connector draft persistence is **last-writer-wins** within a scope:

- `ActorBackedRoleCatalogStore.SaveRoleDraftAsync` (`src/Aevatar.Studio.Infrastructure/ActorBacked/ActorBackedRoleCatalogStore.cs:124-145`)
- `ActorBackedConnectorCatalogStore.SaveConnectorDraftAsync` (`src/Aevatar.Studio.Infrastructure/ActorBacked/ActorBackedConnectorCatalogStore.cs:135-157`)

Neither path carries an expected version. The actor state (`RoleCatalogState`, `ConnectorCatalogState` proto) has no version / ETag field. The apply path (`ApplyDraftSaved` in `RoleCatalogGAgent.cs:67-77` / `ConnectorCatalogGAgent.cs:68-78`) overwrites unconditionally. HTTP controllers (`RolesController`, `ConnectorsController`) accept no `If-Match` header and return no `ETag`.

Scope isolation (per-scope actors) prevents cross-scope interference, but two concurrent users in the **same scope** silently overwrite each other:

1. User A reads draft (v?)
2. User B reads draft (v?)
3. User A saves → applied unconditionally
4. User B saves → silently overwrites A's changes

## Acceptance criteria

- Add `version` (or equivalent monotonic counter) field to `RoleCatalogState` and `ConnectorCatalogState` proto; increment on each `Apply*` step
- Surface version on read (`ETag` response header or response body field)
- Accept expected version on write (`If-Match` header or request body field); return **409 Conflict** on mismatch
- Integration tests for concurrent draft writers in the same scope (parallel save → second writer sees 409)

## Out of scope

- Workflow draft concurrency (different store, separate question)
- Cross-scope concurrency (already handled by per-scope actor isolation)

## Related

- #251 Q-04 — the resolver-waterfall part of Q-04 was addressed by PR #277, but this concurrency residue was never tracked separately
- #250 — test plan should add concurrent-draft assertions once this lands

eanzhao:
## Summary

Closes #432. Adds optimistic concurrency control to role and connector catalog drafts so that two concurrent writers in the same scope no longer silently overwrite each other.

- HTTP: GET returns `ETag: \"<version>\"`; PUT/DELETE accept `If-Match` header (or `expectedVersion` request body field) and return **409** with `code: VERSION_CONFLICT` on mismatch.
- Service / store: write methods now accept `long? expectedVersion`; stored DTOs and responses surface `Version`.
- Actor (RoleCatalogGAgent, ConnectorCatalogGAgent): each Apply step bumps `LastAppliedEventVersion`; each handler validates `expected_version` when set and raises `EventStoreOptimisticConcurrencyException` on mismatch.
- Backward compatible: callers that omit `expected_version` / `If-Match` continue to work as last-writer-wins.

## Proto changes

- `RoleCatalogState` / `ConnectorCatalogState`: add `int64 last_applied_event_version`.
- `*CatalogSavedEvent` / `*DraftSavedEvent` / `*DraftDeletedEvent`: add `optional int64 expected_version` (proto3 optional; unset = skip check).

## Files

- Actors: `agents/Aevatar.GAgents.RoleCatalog/RoleCatalogGAgent.cs`, `agents/Aevatar.GAgents.ConnectorCatalog/ConnectorCatalogGAgent.cs`
- Store interfaces & impls: `IRoleCatalogStore` / `ActorBackedRoleCatalogStore`, `IConnectorCatalogStore` / `ActorBackedConnectorCatalogStore`
- Stored DTOs: `IStudioWorkspaceStore.cs` (`StoredRoleCatalog/Draft`, `StoredConnectorCatalog/Draft` get `Version`)
- Service contracts: `RoleContracts.cs`, `ConnectorContracts.cs` (`Version` on responses, `ExpectedVersion?` on requests)
- Services: `RoleCatalogService.cs`, `ConnectorService.cs`
- Controllers: `RolesController.cs`, `ConnectorsController.cs` (ETag/If-Match plumbing + 409)
- New: `ETagSupport.cs` (shared header helpers)

## Test plan

- [x] \`dotnet build aevatar.slnx\` — clean (only pre-existing warnings)
- [x] \`dotnet test test/Aevatar.Tools.Cli.Tests\` — 399 / 399 pass
- [x] \`dotnet test test/Aevatar.GAgentService.Integration.Tests\` — 277 / 277 pass
- [x] \`bash tools/ci/test_stability_guards.sh\` — passed

### New test coverage

In \`ActorBackedGAgentStateTransitionTests.cs\`:
- \`RoleCatalog_Apply_IncrementsLastAppliedEventVersion\`
- \`ConnectorCatalog_Apply_IncrementsLastAppliedEventVersion\`
- Existing apply helpers updated to mirror the actor (bump version on each Apply).

In \`ActorBackedStoreAdapterTests.cs\`:
- \`RoleCatalogStore_SaveDraft_PlumbsExpectedVersionToEvent\`
- \`RoleCatalogStore_SaveDraft_WithoutExpectedVersion_LeavesEventUnsetForOptimisticBypass\`
- \`RoleCatalogStore_DeleteDraft_PlumbsExpectedVersionToEvent\`
- \`RoleCatalogStore_GetDraft_ReturnsVersionFromProjectionState\`
- \`ConnectorCatalogStore_SaveDraft_PlumbsExpectedVersionToEvent\`
- \`ConnectorCatalogStore_GetDraft_ReturnsVersionFromProjectionState\`

## Out of scope (deferred)

- End-to-end actor harness for the conflict-throw path. Foundation's \`EventStoreOptimisticConcurrencyException\` semantics are exercised separately; full e2e coverage to follow.
- Frontend updates to send \`If-Match\` headers (next-step task once backend is live).


src/Aevatar.Studio.Infrastructure/ActorBacked/ActorBackedRoleCatalogStore.cs
```diff
@@ -58,23 +58,29 @@ public async Task<StoredRoleCatalog> GetRoleCatalogAsync(CancellationToken cance
...
         await ActorCommandDispatcher.SendAsync(_dispatchPort, actor, evt, cancellationToken);
 
+        var refreshed = await ReadProjectedStateAsync(cancellationToken);
         return new StoredRoleCatalog(
             HomeDirectory: ActorHomeDirectory,
             FilePath: ActorFilePath,
             FileExists: true,
-            Roles: catalog.Roles);
+            Roles: catalog.Roles,
+            Version: refreshed?.LastAppliedEventVersion ?? ((expectedVersion ?? catalog.Version) + 1));
```
eanzhao: This can return a stale version/ETag after a successful write. `ReadProjectedStateAsync` reads the CQRS projection immediately after dispatch, but that projection is eventually consistent; if it still contains the pre-write document, the response uses the old `LastAppliedEventVersion`. A client can then `GET v5 -> PUT If-Match v5` successfully, receive `ETag: "5"` again, and hit a false `VERSION_CONFLICT` on the next guarded write. The same pattern exists in the connector catalog and draft save paths. The write response should derive the next version from the committed actor/write-side result, or avoid emitting a stronger post-write ETag than the system can actually guarantee.

src/Aevatar.Studio.Hosting/Controllers/ETagSupport.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        if (!request.Headers.TryGetValue(IfMatchHeader, out var values) || values.Count == 0)
+            return null;
+
+        var raw = values.ToString().Trim();
+        if (string.IsNullOrEmpty(raw))
+            return null;
+
+        var unquoted = raw.Trim('"');
+        return long.TryParse(unquoted, out var version) ? version : null;
```
eanzhao: Malformed or unsupported `If-Match` values currently disable concurrency protection because this returns `null` for both “header absent” and “header present but invalid”. That means a caller intending a guarded write with `If-Match: W/"5"`, `If-Match: "5", "6"`, or a typo silently falls back to last-writer-wins. Please distinguish absence from parse failure and reject invalid preconditions instead of treating them as unguarded writes.

eanzhao (commented):
I found two correctness issues in the optimistic concurrency plumbing. The focused test slice passes locally, but these should be addressed before relying on the new ETag flow.

src/Aevatar.Studio.Infrastructure/ActorBacked/ActorBackedRoleCatalogStore.cs
eanzhao: Fixed in 8b1d922. Dropped the post-write projection re-read entirely — `SaveRoleCatalogAsync` / `SaveRoleDraftAsync` / connector equivalents now return `Version = expectedVersion + 1` deterministically (the actor enforces the match and Apply increments by exactly one). When the caller doesn't supply `expected_version`, the response carries `Version = 0` and the controller skips the `ETag` header — clients re-GET for an authoritative value instead of being lied to. Helper `NextDeterministicVersion` documents the invariant. Controller-level test `SaveDraft_WithValidIfMatch_PassesExpectedVersionToStore_AndEmitsDeterministicETag` locks in `If-Match: "5" → ETag: "6"` for the happy path.

src/Aevatar.Studio.Hosting/Controllers/ETagSupport.cs
eanzhao: Fixed in 8b1d922. `TryParseIfMatch` is replaced by tristate `ParseIfMatch(HttpRequest, out long) → IfMatchStatus { Absent | Valid | Invalid }`. Controllers reject `Invalid` with `400 MALFORMED_IF_MATCH`; only `Valid` values bind `ExpectedVersion`; `Absent` passes through unchanged. The new theory `ParseIfMatch_WhenMalformed_ReturnsInvalid` covers `W/"5"`, `"5","6"`, `*`, non-numeric, empty quoted, and negative; controller-level `SaveDraft_WithMalformedIfMatch_Returns400_AndDoesNotInvokeStore` and the DELETE counterpart prove the store is never called when a malformed precondition arrives.

src/Aevatar.Studio.Hosting/Controllers/RolesController.cs
```diff
@@ -151,4 +181,58 @@ public async Task<IActionResult> DeleteDraft(CancellationToken cancellationToken
...
+    {
+        var status = ETagSupport.ParseIfMatch(Request, out var version);
+        if (status == IfMatchStatus.Invalid)
+        {
+            effective = request;
+            malformed = MalformedIfMatch();
+            return true;
+        }
+
+        effective = status == IfMatchStatus.Valid && request.ExpectedVersion is null
```
eanzhao: This still lets the request body override an explicit `If-Match` precondition. The XML comment says the header takes precedence, but when both are present this keeps `request.ExpectedVersion`; more importantly, an HTTP `If-Match: "5"` can be bypassed by sending `expectedVersion: 6` in the body, allowing the write to succeed when the header precondition should fail. If `If-Match` is present and valid, use that value as the expected version, or reject the request when the body value disagrees. The connector controller has the same pattern.

eanzhao (commented):
Re-reviewed the fix. The stale post-write ETag issue and malformed If-Match fallback are addressed, and the focused ETag/controller/store tests pass locally. I left one remaining inline concern around mixed header/body preconditions.

src/Aevatar.Studio.Hosting/Controllers/RolesController.cs
eanzhao: Fixed in 819b6ab. When If-Match parses cleanly the header is authoritative; if the body explicitly disagrees the request is rejected with `400 IF_MATCH_BODY_MISMATCH`. Body-only (header absent) and header-only (body null) paths are unchanged. Symmetric fix in both `RolesController` and `ConnectorsController`. New tests `Save_WhenIfMatchHeaderDisagreesWithBodyExpectedVersion_Returns400`, `SaveDraft_WhenIfMatchHeaderDisagreesWithBodyExpectedVersion_Returns400`, and `SaveDraft_WhenIfMatchHeaderAgreesWithBodyExpectedVersion_HeaderWinsAndStoreReceivesIt` cover both controllers.

eanzhao (commented):
Re-reviewed the latest fix at `819b6ab`. The remaining If-Match/body mismatch issue is addressed: valid `If-Match` is now authoritative, mismatched body `expectedVersion` returns 400, and both roles/connectors have coverage. No remaining findings from my side on the optimistic concurrency/ETag flow.

Local verification in a detached PR worktree:
- `dotnet test test/Aevatar.Tools.Cli.Tests/Aevatar.Tools.Cli.Tests.csproj --no-restore --filter "FullyQualifiedName~ETagSupportTests|FullyQualifiedName~RolesControllerETagTests|FullyQualifiedName~ConnectorsControllerETagTests|FullyQualifiedName~ActorBackedStoreAdapterTests" --nologo` — 104 passed
- `bash tools/ci/test_stability_guards.sh` — passed

PR #438, opened 2026-04-27
fix(daily): scope GitHub username binding per Lark user (#436)

### LINKED ISSUE #436: bug(daily): GitHub username binding shared across all Lark users of one bot — last writer wins
> Tracks the same bug as #437 (user-facing report: `/daily` binding causes cross-user data leakage). This issue captures the engineering analysis and fix direction; #437 is the QA-side reproduction.

## Symptom

I shared my Lark bot with a colleague. After they ran `/daily` and bound their GitHub username (`Yuezh0127`), my own `/daily` started reporting on `Yuezh0127` instead of my GitHub account. Re-binding mine just flips it again — last writer wins. The GitHub username binding is behaving as a **single global value per bot**, not per Lark user.

Reproduction:

1. User A (bot owner) runs `/daily` in Lark, binds GitHub username `alice`.
2. User B (added to the same Lark bot) runs `/daily`, binds GitHub username `bob`.
3. User A runs `/daily` again → daily report scheduled for `bob`. A's binding is gone.

(Screenshot in original report: my `/daily` shows `Daily report scheduled for Yuezh0127` — Yuezh0127 is the colleague's GitHub account, not mine.)

Also reproduces in 1:1 private chats with the bot, not only in group chats (see #437).

## Root cause

`UserConfigGAgent` is keyed by **bot registration scope**, not by Lark user identity. All Lark users sharing one bot share one `UserConfigGAgent` actor (`user-config-{registrationScopeId}`), so any user's `SaveGithubUsernameAsync` overwrites the binding for every other user of that bot.

The relevant chain:

1. `agents/Aevatar.GAgents.ChannelRuntime/ChannelConversationTurnRunner.cs:787` — when an inbound Lark message is converted to `ChannelInboundEvent`, `RegistrationScopeId` is set to `registration.ScopeId` (a per-bot value). The per-Lark-user `SenderId` is also populated (line 779) but is not propagated downstream as the user-config scope.

2. `agents/Aevatar.GAgents.ChannelRuntime/AgentBuilderCardFlow.cs:79-81` — when prefilling the daily-report form, the saved username is read using `evt.RegistrationScopeId`:
   ```csharp
   preferredGithubUsername = (await userConfigQueryPort.GetAsync(
       NormalizeScopeId(evt.RegistrationScopeId),
       ct)).GithubUsername;
   ```

3. `agents/Aevatar.GAgents.ChannelRuntime/AgentBuilderTool.cs:186-187` — when creating the daily-report agent, the same bot-level scope is used:
   ```csharp
   var rawScopeId = NormalizeOptional(AgentToolRequestContext.TryGet(\"scope_id\"));
   var configScopeId = NormalizeScopeId(rawScopeId);
   ```
   And on save (line 1579):
   ```csharp
   await commandService.SaveGithubUsernameAsync(scopeId, githubUsername, ct);
   ```

4. `src/Aevatar.Studio.Projection/CommandServices/ActorDispatchUserConfigCommandService.cs:77` — the actor ID is derived purely from the scope:
   ```csharp
   var actorId = ActorIdPrefix + NormalizeScopeId(scopeId);
   // -> \"user-config-{bot-registration-scope}\"
   ```

5. `src/Aevatar.Studio.Projection/QueryPorts/ProjectionUserConfigQueryPort.cs:41` — read uses the same key, so all Lark users in the bot read the same record.

The proto already carries the per-Lark-user identity (`ChannelInboundEvent.sender_id`, field 2 in `agents/Aevatar.GAgents.ChannelRuntime/channel_runtime_messages.proto`); it's just not threaded into the user-config scope.

## How this slipped in

Issue #327 (closed) added `GithubUsername` to `UserConfig` under the assumption \"Day One 的现实是'一个 user 关心一个 github'，跨对话共享是 feature 不是 bug.\" That was correct for a single-owner CLI/Tools UI scenario, where the scope resolver returns a per-NyxID-user scope. But on the Lark adapter the scope is the bot's registration scope (one bot = one NyxID identity = one scope), and the bot is shared across many Lark users. The CLI semantics (one identity per scope) and the Lark semantics (many Lark users per scope) collide on the same actor.

## Suggested fix direction (not prescriptive)

The fix has to give each Lark user their own `UserConfigGAgent`. Two shapes worth considering:

- **Composite scope**: thread `ChannelInboundEvent.SenderId` through `AgentToolRequestContext` and use a composite key like `{registrationScopeId}:lark:{senderId}` for the user-config actor on Lark inbound paths. CLI/Tools paths keep their existing scope. This is the smallest blast radius.
- **Channel-user-scoped actor**: introduce a per-platform-user identity that the Lark adapter resolves (probably tying back to `ChannelUserBindingGAgent`, mentioned in #327's binding decomposition) and use that as the user-config scope. Cleaner long-term, but bigger surface change.

Either way:

- The `sender_id` (or a normalized channel-user-id) needs to flow through `AgentToolRequestContext` → `AgentBuilderTool` → `IUserConfigCommandService` and through `AgentBuilderCardFlow` → `IUserConfigQueryPort`.
- Existing data keyed at the bot-registration scope is now ambiguous. Probably acceptable to let each user re-bind on first `/daily` (the form already prompts when no username is prefilled), and clean up the orphaned shared record out-of-band.

## Acceptance criteria

- [ ] Two Lark users in the same bot can each bind a different GitHub username via `/daily`, and each user's `/daily` schedules the report for their own GitHub account.
- [ ] Same isolation holds in 1:1 private chats with the bot (per #437).
- [ ] Tests cover: (a) two distinct `sender_id`s in the same `RegistrationScopeId` round-trip independent `GithubUsername` values; (b) prefill returns each user's own value, not the most recent writer's.
- [ ] CLI / Tools UI path (`UserConfigEditor.tsx`) is unaffected — single-user-per-scope semantics there should not regress.
- [ ] Existing `SkillRunnerGAgent.State.OutboundConfig.GithubUsername` snapshot semantics from #327 stay intact (running agents are not retroactively repointed when a user re-binds).

## Affected files (for triage)

- \`agents/Aevatar.GAgents.ChannelRuntime/ChannelConversationTurnRunner.cs\` (line 787)
- \`agents/Aevatar.GAgents.ChannelRuntime/AgentBuilderCardFlow.cs\` (lines 67-95)
- \`agents/Aevatar.GAgents.ChannelRuntime/AgentBuilderTool.cs\` (lines 186-209, 1564-1590)
- \`src/Aevatar.Studio.Projection/CommandServices/ActorDispatchUserConfigCommandService.cs\` (lines 75-89)
- \`src/Aevatar.Studio.Projection/QueryPorts/ProjectionUserConfigQueryPort.cs\` (lines 39-63)
- \`agents/Aevatar.GAgents.ChannelRuntime/channel_runtime_messages.proto\` (\`ChannelInboundEvent.sender_id\` already present, currently unused for scope)

Related: #437 (user-facing duplicate of this bug), #327 (introduced the binding), #254 (multi-channel adapter RFC — same multi-tenant concern surfaces here).

eanzhao:
## Summary

Fixes #436. The bot's `RegistrationScopeId` is per-NyxID-account (one bot = one scope), so multiple Lark users sharing one bot all hit the same `UserConfigGAgent` actor (`user-config-{registrationScopeId}`) and each `/daily` save overwrites everyone else's saved `github_username`. Last writer wins — colleague binds their username, bot owner's daily report starts targeting colleague's GitHub account.

This patch composes `{registrationScopeId}:{platform}:{senderId}` for the user-config read (form prefill in `AgentBuilderCardFlow.TryResolveAsync`) and write (preference save in `AgentBuilderTool.CreateDailyReportAgentAsync`). Each Lark user now gets their own `UserConfigGAgent`.

`SkillRunnerGAgent.ScopeId` stays **bot-scoped** on purpose — downstream tools that read `scope_id` as a NyxID tenant key (binding store, service invocation, etc.) need the bot's scope, not a per-end-user composite. The CLI / Tools UI goes through `IAppScopeResolver` and is unaffected.

## What changed

- `agents/Aevatar.GAgents.ChannelRuntime/ChannelUserConfigScope.cs` (new) — composes the per-end-user scope from `ChannelInboundEvent` or from `AgentToolRequestContext.CurrentMetadata`. Falls back to bare `registrationScopeId` when no sender is present (programmatic / system paths).
- `AgentBuilderCardFlow.TryResolveAsync` — reads via `ChannelUserConfigScope.FromInboundEvent(evt)` instead of `NormalizeScopeId(evt.RegistrationScopeId)`.
- `AgentBuilderTool.CreateDailyReportAgentAsync` — separates `configScopeId` (bot scope, still used for `InitializeSkillRunnerCommand.ScopeId`) from `userConfigScopeId` (per-end-user composite, used for both `ResolveDailyReportGithubUsernameAsync` read and `SaveGithubUsernamePreferenceIfRequestedAsync` write).

## Test plan

- [x] `dotnet test test/Aevatar.GAgents.ChannelRuntime.Tests` — 473 passed, 0 failed.
- [x] `dotnet test test/Aevatar.Tools.Cli.Tests --filter UserConfig` — 26 passed, 0 failed (CLI / projection paths unaffected).
- [x] `bash tools/ci/architecture_guards.sh` — exit 0.
- [x] New `ChannelUserConfigScopeTests` — 7 tests pinning the composition rules: per-user differentiation, fallback when SenderId missing, default scope when RegistrationScopeId missing, channel-literal fallback when Platform missing, casing collapse, FromInboundEvent ↔ FromMetadata parity.
- [x] Updated `AgentBuilderCardFlowTests.TryResolveAsync_DailyReportLaunch_PrefillsSavedGithubUsername` — pins prefill reads from per-user scope.
- [x] New `AgentBuilderCardFlowTests.TryResolveAsync_DailyReportLaunch_TwoLarkUsersInSameBot_SeeIndependentSavedUsernames` — pins the bug-fix scenario directly.
- [x] Updated `AgentBuilderToolTests.ExecuteAsync_CreateAgent_DailyReport_SavesGithubUsernamePreference_WhenRequested` — asserts `SaveGithubUsernameAsync("scope-1:lark:ou_alice", ...)`, not `("scope-1", ...)`.

## Migration note

Existing user-config records keyed at the bot's `RegistrationScopeId` are now ambiguous (no per-user provenance). Each Lark user will simply re-bind on first `/daily` after this lands — the form already prompts when no username is prefilled. The orphan shared record can be cleaned up out-of-band.

### DIFF
--- a/agents/Aevatar.GAgents.ChannelRuntime/AgentBuilderTool.cs
+++ b/agents/Aevatar.GAgents.ChannelRuntime/AgentBuilderTool.cs
@@ -185,11 +185,17 @@ public sealed class AgentBuilderTool : IAgentTool
     {
         var rawScopeId = NormalizeOptional(AgentToolRequestContext.TryGet("scope_id"));
         var configScopeId = NormalizeScopeId(rawScopeId);
+        // Bot's RegistrationScopeId is per-NyxID-account (one bot = one scope), so multiple
+        // Lark users sharing one bot would otherwise share a single UserConfigGAgent and
+        // overwrite each other's saved github_username (issue #436). Compose a per-end-user
+        // scope from the channel sender for personal-preference reads/writes only;
+        // SkillRunner.ScopeId stays bot-scoped for downstream NyxID-tenant tools.
+        var userConfigScopeId = ChannelUserConfigScope.FromMetadata(AgentToolRequestContext.CurrentMetadata);
         var githubUsernameResolution = await ResolveDailyReportGithubUsernameAsync(
             args,
             nyxClient,
             token,
-            configScopeId,
+            userConfigScopeId,
             ct);
         if (githubUsernameResolution.ErrorResponse is not null)
             return githubUsernameResolution.ErrorResponse;
@@ -318,7 +324,7 @@ public sealed class AgentBuilderTool : IAgentTool
 
         var savePreferenceRequested = args.Bool("save_github_username_preference") == true;
         var preferenceSaved = await SaveGithubUsernamePreferenceIfRequestedAsync(
-            configScopeId,
+            userConfigScopeId,
             githubUsernameResolution.GithubUsername ?? string.Empty,
             savePreferenceRequested,
             ct);
--- a/agents/Aevatar.GAgents.ChannelRuntime/AgentBuilderCardFlow.cs
+++ b/agents/Aevatar.GAgents.ChannelRuntime/AgentBuilderCardFlow.cs
@@ -77,7 +77,7 @@ internal static class AgentBuilderCardFlow
             try
             {
                 preferredGithubUsername = (await userConfigQueryPort.GetAsync(
-                    NormalizeScopeId(evt.RegistrationScopeId),
+                    ChannelUserConfigScope.FromInboundEvent(evt),
                     ct)).GithubUsername;
             }
             catch (OperationCanceledException)
@@ -594,9 +594,6 @@ internal static class AgentBuilderCardFlow
         var normalized = (value ?? string.Empty).Trim();
         return normalized.Length == 0 ? null : normalized;
     }
-
-    private static string NormalizeScopeId(string? scopeId) =>
-        string.IsNullOrWhiteSpace(scopeId) ? "default" : scopeId.Trim();
 
     private static string FormatCreateSocialMediaResult(JsonElement root)
     {
--- a/agents/Aevatar.GAgents.ChannelRuntime/ChannelUserConfigScope.cs
+++ b/agents/Aevatar.GAgents.ChannelRuntime/ChannelUserConfigScope.cs
@@ -1,0 +1,49 @@
+namespace Aevatar.GAgents.ChannelRuntime;
+
+/// <summary>
+/// Composes the per-end-user scope id used by <c>UserConfigGAgent</c> for
+/// channel-bound preferences such as the saved <c>github_username</c>.
+///
+/// The bot's <c>RegistrationScopeId</c> alone is per-NyxID-account (one bot =
+/// one scope), so multiple Lark users sharing the same bot would otherwise
+/// share a single user-config record and overwrite each other's preferences
+/// (issue #436). The composite <c>{registrationScopeId}:{platform}:{senderId}</c>
+/// gives every channel sender their own actor while leaving the bot scope
+/// intact for downstream tools that legitimately need NyxID tenant scope
+/// (binding store, service invocation, etc.).
+/// </summary>
+internal static class ChannelUserConfigScope
+{
+    private const string DefaultScope = "default";
+    private const string DefaultPlatform = "channel";
+
+    public static string FromInboundEvent(ChannelInboundEvent evt)
+    {
+        ArgumentNullException.ThrowIfNull(evt);
+        return Compose(evt.RegistrationScopeId, evt.Platform, evt.SenderId);
+    }
+
+    public static string FromMetadata(IReadOnlyDictionary<string, string>? metadata)
+    {
+        if (metadata is null)
+            return DefaultScope;
+
+        metadata.TryGetValue("scope_id", out var scope);
+        metadata.TryGetValue(ChannelMetadataKeys.Platform, out var platform);
+        metadata.TryGetValue(ChannelMetadataKeys.SenderId, out var senderId);
+        return Compose(scope, platform, senderId);
+    }
+
+    private static string Compose(string? scopeId, string? platform, string? senderId)
+    {
+        var normalizedScope = string.IsNullOrWhiteSpace(scopeId) ? DefaultScope : scopeId.Trim();
+        var normalizedSender = senderId?.Trim();
+        if (string.IsNullOrEmpty(normalizedSender))
+            return normalizedScope;
+
+        var normalizedPlatform = string.IsNullOrWhiteSpace(platform)
+            ? DefaultPlatform
+            : platform.Trim().ToLowerInvariant();
+        return $"{normalizedScope}:{normalizedPlatform}:{normalizedSender}";
+    }
+}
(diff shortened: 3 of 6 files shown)


agents/Aevatar.GAgents.ChannelRuntime/AgentBuilderTool.cs
```diff
@@ -185,11 +185,17 @@ private async Task<string> CreateDailyReportAgentAsync(
     {
         var rawScopeId = NormalizeOptional(AgentToolRequestContext.TryGet("scope_id"));
         var configScopeId = NormalizeScopeId(rawScopeId);
+        // Bot's RegistrationScopeId is per-NyxID-account (one bot = one scope), so multiple
+        // Lark users sharing one bot would otherwise share a single UserConfigGAgent and
+        // overwrite each other's saved github_username (issue #436). Compose a per-end-user
+        // scope from the channel sender for personal-preference reads/writes only;
+        // SkillRunner.ScopeId stays bot-scoped for downstream NyxID-tenant tools.
+        var userConfigScopeId = ChannelUserConfigScope.FromMetadata(AgentToolRequestContext.CurrentMetadata);
```
eanzhao, line 193: The implementation here looks right: both the saved-preference read and write now use the per-channel-user scope while `configScopeId` stays bot-scoped for the runner. The remaining gap is test coverage for this exact read path. The real relay `/daily` no-username flow reaches `ResolveDailyReportGithubUsernameAsync` through `AgentBuilderTool`; today `ExecuteAsync_CreateAgent_DailyReport_UsesSavedGithubUsernamePreference_WhenArgumentMissing` still omits `channel.platform` / `channel.sender_id` and stubs `GetAsync("scope-1", ...)`, so it would not fail if this argument accidentally drifted back to `configScopeId`. Please extend that test (or add a sibling) to set `lark` + `ou_alice` and assert the query port is called with `scope-1:lark:ou_alice`.

eanzhao (commented):
I found one coverage gap while reviewing #436. The implementation looks correct, but the real no-argument `/daily` relay path depends on the AgentBuilderTool saved-preference read fallback, so that path should be pinned too.

agents/Aevatar.GAgents.ChannelRuntime/AgentBuilderTool.cs
eanzhao: Addressed in 5ac841a9. The existing test now sets `channel.platform=lark` + `channel.sender_id=ou_alice`, stubs the per-user scope (`scope-1:lark:ou_alice`) with "saved-user" AND stubs the bot-only scope (`scope-1`) with a sentinel "WRONG-bot-scope-leak", and asserts directly that `GetAsync("scope-1:lark:ou_alice")` fired and `GetAsync("scope-1")` did not. The sentinel makes the prompt-content assertion fail loudly on regression instead of accidentally still matching "saved-user".

PR #443, opened 2026-04-27
fix: clarify Studio member creation flow

### LINKED ISSUE #441: Fix Create member Script and GAgent builder entry UX
## Problem
Create member currently presents Script and GAgent like real create flows even though their member create APIs are not available yet. This makes the Member name input feel meaningful for Script/GAgent, then blocks the user.

## Scope
- Keep Workflow, Script, and GAgent as selectable member types.
- Workflow keeps the existing Member name and Workflow directory create flow.
- Script hides Member name and uses an Open Script builder action.
- GAgent hides Member name and uses an Open GAgent builder action.
- Do not touch Script binding, invoke, runtime, or backend behavior in this issue.

## Acceptance Criteria
- Selecting Workflow shows Member name and creates a blank workflow draft as before.
- Selecting Script does not show Member name and opens Build > Script from the modal action.
- Selecting GAgent does not show Member name and opens Build > GAgent from the modal action.
- Existing workflow create tests still pass.
- New Studio page tests cover Script and GAgent modal behavior.

### LINKED ISSUE #442: Complete Studio Script member lifecycle gating and invoke contract handling
## Problem
Studio can edit and draft-run Script implementations, but the member lifecycle still behaves as workflow-first. Script cannot reliably move through save observation, bind, invoke, and observe as a first-class member without risking stale source or incorrect endpoint assumptions.

## Scope
- Add Script Build state handoff from StudioScriptBuildPanel to the Studio parent.
- Only expose a Script pending bind candidate after the saved revision is observed as applied in the script catalog.
- Add a Script branch in the parent bind handler that calls studioApi.bindScopeScript.
- Refetch/select the bound service after binding.
- Make Invoke use the bound service endpoint contract instead of assuming every Script supports chat stream.
- Keep draft-run inside Build and do not treat draft-run as proof that the member is bound or invokable.

## Acceptance Criteria
- Dirty Script source does not create a bind candidate.
- Save accepted but not catalog-applied Script source does not create a bind candidate.
- Catalog-applied Script revision creates a Script pending bind candidate.
- Binding that candidate calls studioApi.bindScopeScript, not bindScopeWorkflow.
- Invoke defaults to chat only when the bound Script service exposes a chat-compatible endpoint.
- Non-chat Script endpoints show an honest contract/blocked state instead of pretending chat works.
- Frontend tests cover save-observed gating, Script bind, and non-chat invoke behavior.

potter-sun:
## Summary

- Clarify the Studio Create member modal so only Workflow uses the immediate member creation form.
- Keep Script and GAgent visible as member type choices, but route them to their Build workspaces instead of asking for a fake member name.
- Complete the Studio Script member lifecycle gating so only catalog-applied Script revisions can bind and invoke through their published endpoint contract.
- Add the Studio Script member flow implementation plan under `docs/design/`.

## Issue Links

Closes #441
Closes #442

## Impact

Workflow creation remains unchanged. Script and GAgent now behave as honest builder entry points until their direct member creation APIs land. Script save observation, bind, invoke, and observe now use the catalog-applied revision and bound service endpoint contract instead of draft-run assumptions.

## Validation

```bash
npm --prefix apps/aevatar-console-web test -- --runInBand
npm --prefix apps/aevatar-console-web run test:ui -- --runTestsByPath src/pages/studio/components/StudioBuildPanels.test.tsx src/pages/studio/index.test.tsx --runInBand
npm --prefix apps/aevatar-console-web run tsc
```


apps/aevatar-console-web/src/pages/studio/components/StudioBuildPanels.tsx
```diff
@@ -1877,11 +1975,38 @@ export const StudioScriptBuildPanel: React.FC<StudioScriptBuildPanelProps> = ({
...
+          proposalId: accepted.acceptedScript.proposalId,
+          expectedBaseRevision: accepted.acceptedScript.expectedBaseRevision,
+          acceptedAt: accepted.acceptedScript.acceptedAt,
+        },
       );
+      await onRefreshScripts?.();
+      if (observation.status === 'applied') {
+        setSaveStatus('applied');
+        setSaveNotice(
+          `Save applied for ${accepted.acceptedScript.scriptId} · revision ${accepted.acceptedScript.revisionId}.`,
```
eanzhao: **Alert severity ends up inverted for the new save outcomes (see L2130 below).**

The pre-existing `<Alert>` keys its `type` off `saveNotice.startsWith('Save accepted') ? 'success' : 'warning'`. The new messages introduced in this hunk are:
- `Save applied for ... · revision ...` (terminal **success**) — does **not** start with `Save accepted` → renders as `warning` (yellow)
- `Save accepted for ... Waiting for catalog.` (intermediate pending) → renders as `success` (green)
- `Save rejected for ...` (terminal failure) → `warning` (correct)

So on the actual success path the user sees a yellow alert, while the in-flight `Waiting for catalog` shows green. Drive the alert `type` from `saveStatus` (`applied`→success, `accepted`→info, `failed`/`rejected`→warning, errors→error) instead of substring-matching the message text.

apps/aevatar-console-web/src/pages/studio/components/StudioBuildPanels.tsx
```diff
@@ -1833,6 +1850,84 @@ export const StudioScriptBuildPanel: React.FC<StudioScriptBuildPanelProps> = ({
...
+  const primaryActionLabel =
+    !activeScript?.script?.scriptId
+      ? 'Select Script'
+      : isDirty || validationStatus === 'unknown' || validationStatus === 'invalid'
+        ? 'Validate'
+        : effectiveSaveStatus === 'accepted'
+          ? 'Waiting for catalog'
+          : effectiveSaveStatus === 'applied'
+            ? 'Continue to Bind'
+            : 'Save revision';
```
eanzhao, line 2173: **Primary action label contradicts `scriptReadyToBind` for catalog-applied scripts.**

When the panel auto-selects a script that is already catalog-applied (`activeRevision` set, source matches), the state is:
- `isDirty = false`
- `validationStatus = 'unknown'` (no local validation has run yet)
- `effectiveSaveStatus = 'applied'` (via the `activeRevision` branch above)
- `scriptReadyToBind = true`

But this ternary short-circuits to `'Validate'` whenever `validationStatus === 'unknown'`, so the green tag at L2095 reads **"Validate"** while the bottom CTA at L2240 simultaneously reads **"Continue to Bind"** and is enabled. The design doc state table says `Catalog applied → Continue to Bind` regardless of whether the user has clicked Validate locally.

Reorder so the `applied + clean` branch wins over `validationStatus === 'unknown'` (or only force `'Validate'` after the user actually edited the source / validation explicitly failed).

apps/aevatar-console-web/src/pages/studio/components/StudioBuildPanels.tsx
```diff
@@ -1877,11 +1975,38 @@ export const StudioScriptBuildPanel: React.FC<StudioScriptBuildPanelProps> = ({
...
+        activeScript.script.scriptId,
+        {
+          revisionId: accepted.acceptedScript.revisionId,
+          definitionActorId: accepted.acceptedScript.definitionActorId,
+          sourceHash: accepted.acceptedScript.sourceHash,
+          proposalId: accepted.acceptedScript.proposalId,
+          expectedBaseRevision: accepted.acceptedScript.expectedBaseRevision,
+          acceptedAt: accepted.acceptedScript.acceptedAt,
+        },
       );
```
eanzhao, line 2533: **Race: the `observeSaveScript` resolution can land against a different script.**

If the user switches `selectedScriptId` (or edits the source, which resets the panel via the L1808 effect) while this `await` is in flight, the resolution still calls `setSaveStatus(...)` and `setSaveNotice(...)` against whatever script is now active. The reset effect on script change clears `saveStatus` to `'idle'`, but this resolver runs after that and stomps on it — leaving e.g. a stale `'applied'` on a freshly-selected unrelated script.

Capture `activeScript.script.scriptId` in a local before the await and bail out (or use a `useRef` request token / `AbortController`) if it no longer matches when the observation resolves. Same applies to `setSavePending(false)` in `finally`.

apps/aevatar-console-web/src/pages/studio/components/StudioBuildPanels.tsx
```diff
@@ -1877,11 +1975,38 @@ export const StudioScriptBuildPanel: React.FC<StudioScriptBuildPanelProps> = ({
...
+        );
+      } else if (observation.status === 'rejected') {
+        setSaveStatus('failed');
+        setSaveNotice(
+          observation.message ||
+            `Save rejected for ${accepted.acceptedScript.scriptId} · revision ${accepted.acceptedScript.revisionId}.`,
+        );
+      } else {
+        setSaveNotice(
+          `Save accepted for ${accepted.acceptedScript.scriptId} · revision ${accepted.acceptedScript.revisionId}. Waiting for catalog.`,
```
eanzhao: **`observation.status === 'pending'` strands the UI in 'Waiting for catalog' indefinitely.**

If the backend returns `pending` (and `isTerminal: false`), `saveStatus` stays `'accepted'` — there is no follow-up poll, no timer, no re-observe affordance. The user has to manually save again or refresh to escape. Combined with the alert-color issue above, the user actually sees a green "Save accepted ... Waiting for catalog" alert that never moves.

Either (a) poll/re-observe with backoff until terminal, or (b) surface a clear timeout + "Re-check" button when the response is non-terminal.

apps/aevatar-console-web/src/pages/studio/index.tsx
```diff
@@ -5984,16 +6100,22 @@ const StudioPage: React.FC = () => {
...
+              }
               okButtonProps={{
                 disabled:
                   inventoryBusyAction === 'create' ||
-                  !trimOptional(createMemberName) ||
-                  createMemberKind !== 'workflow' ||
-                  !trimOptional(
-                    trimOptional(createMemberDirectoryId) || inventoryDirectoryId,
-                  ),
+                  (createMemberKind === 'workflow' &&
```
eanzhao, line 7578: **OK button stays enabled when the Script feature is disabled.**

With `createMemberKind === 'script'` and `appContextQuery.data?.features.scripts === false`, this `disabled` predicate only checks the Workflow branch, so the `Open Script builder` button is clickable. `handleCreateMember` then surfaces a `message.warning` and bails. Cleaner to disable the button preemptively so the disabled state matches reality:

```ts
disabled:
  inventoryBusyAction === 'create' ||
  (createMemberKind === 'workflow' && (...)) ||
  (createMemberKind === 'script' && !appContextQuery.data?.features.scripts),
```

apps/aevatar-console-web/src/pages/studio/index.tsx
```diff
@@ -2699,49 +2702,109 @@ const StudioPage: React.FC = () => {
...
+    if (shouldBuildScriptCandidate) {
+      const selectedId = trimOptional(selectedScriptId);
+      const catalogScriptState =
+        selectedId
+          ? availableScopeScripts.find(
+              (detail) => trimOptional(detail.script?.scriptId) === selectedId,
+            ) ?? null
+          : null;
+      const effectiveScriptState =
+        scriptBuildState?.scriptId ? scriptBuildState : catalogScriptState
```
eanzhao: **Stale `scriptBuildState` after the Script panel unmounts.**

`StudioScriptBuildPanel` only emits via `onScriptBuildStateChange` while it is mounted; when the user navigates to Bind/Invoke/Observe the panel unmounts but `scriptBuildState` in the parent retains its last value. The `(!selectedId || scriptId === selectedId)` guard a few lines down mostly compensates, but there's a window: returning to `Build > Script` with no selection, before the auto-select effect fires, briefly produces a Script bind candidate based on the previous script's state because `selectedId` is empty so `!selectedId` is true.

Either (a) reset `scriptBuildState` to `null` on Script panel unmount via a cleanup callback, or (b) tighten the guard to require `selectedId && scriptId === selectedId` in both the build and bind branches.

Adjacent nit: `displayName: trimOptional(catalogScriptState.script?.scriptId)` here, and `displayName: activeScript.script.scriptId` over in `StudioBuildPanels.tsx:1895`, both duplicate the script id into `displayName`. Either drop the field or wire it to a real human label so it carries actual information.

apps/aevatar-console-web/src/pages/studio/index.test.tsx
```diff
@@ -4669,6 +4759,90 @@ describe("StudioPage", () => {
...
+      expect(studioApi.bindScopeScript).toHaveBeenCalledWith(
+        expect.objectContaining({
+          scopeId: "scope-1",
+          displayName: "script-alpha",
+          scriptId: "script-alpha",
+          scriptRevision: "rev-1",
+          revisionId: "rev-1",
+        })
+      );
+    });
```
eanzhao, line 5340: **Missing test coverage for the non-happy paths added in this PR.**

The new lifecycle gates have no tests for:
- dirty Script source produces no bind candidate (asserted by the design doc, not covered),
- `observeSaveScript` returning `'pending'` produces no bind candidate (and currently strands the UI — see the related comment in `StudioBuildPanels.tsx`),
- `observeSaveScript` returning `'rejected'` surfaces an error and disables Continue to Bind,
- Script feature disabled keeps the modal in place instead of navigating.

This test only covers the happy path. Worth adding the dirty / pending / rejected cases so the lifecycle gates stay correct under future refactors.

eanzhao (commented):
Reviewed the Script lifecycle and Create-member modal changes. Direction matches the design doc, but there are a few correctness issues worth fixing before merge — flagged inline. Highest priority: the `saveNotice` alert color is inverted because the new `Save applied` / `Save rejected` messages no longer match the legacy `startsWith('Save accepted')` check.

Issue #372, opened 2026-04-24
Harden NyxId scoped chat endpoints with scope guard and scope-owned actor validation


louis4li:
## Context
The scoped NyxId chat endpoints under `/api/scopes/{scopeId}/nyxid-chat/...` currently rely on route shape alone and do not consistently enforce the shared authenticated scope guard before operating inside a requested scope.

This is not a new regression from the current branch. The same route shape and bare-`actorId` resolution already exist on `dev`, but PR #452 / issue #348 replace the old store-shaped contract with explicit registry command/query/admission ports, so the fix should use the new ports rather than `IGAgentActorStore`.

## Problem
- scoped NyxId chat endpoints are mounted under `/api/scopes/...` but create/list/delete do not apply `AevatarScopeAccessGuard`
- `stream` / `approve` currently require a bearer token and registry admission, but still need caller-scope validation before target admission
- callers who know a valid `actorId` can potentially target a conversation outside the authenticated caller scope if the route scope is not checked first
- registry admission proves target ownership for `(scopeId, resource, operation)`; it does not by itself prove the caller is authorized to use that `scopeId`

## Desired fix
Apply the shared scope guard to NyxID scoped chat endpoints and require conversation ownership validation through the scope resource admission port before target operations.

## Acceptance criteria
- scoped NyxID chat endpoints reject mismatched authenticated scopes via `AevatarScopeAccessGuard`
- create/list/delete validate caller scope before performing registry command/query or side effects
- `stream` / `approve` validate caller scope before `IScopeResourceAdmissionPort.AuthorizeTargetAsync`
- `stream` / `approve` / `delete` validate `(scopeId, ScopeResourceKind.GAgentActor, NyxIdChatServiceDefaults.GAgentTypeName, actorId, operation)` through `IScopeResourceAdmissionPort`
- bare `actorId` is no longer sufficient to access a conversation across scopes
- targeted endpoint coverage exists for scope mismatch and unregistered conversation paths

Issue #348, opened 2026-04-23
Refine GAgent actor registry into explicit command/query/ownership ports


louis4li:
## Background

`IGAgentActorStore` no longer behaves like a legacy side-store.

Today it is already actor-backed:
- writes dispatch `ActorRegisteredEvent` / `ActorUnregisteredEvent` to `GAgentRegistryGAgent`
- reads come from `GAgentRegistryCurrentStateDocument`

So the current abstraction is functionally a registry facade, not a generic store.

Recent scoped endpoint hardening exposed a sharper version of the same architectural problem:
- PR #384 originally tried to validate NyxID chat conversation ownership by reading `IGAgentActorStore.GetAsync(scopeId)` before `stream` / `approve` / `delete`.
- PR #390 tried the same pattern for StreamingProxy rooms before `chat` / `message stream` / `post message` / `join` / `delete` / `list participants`.
- In both cases, `GetAsync(scopeId)` reads the registry current-state projection, while creation/register writes are command-dispatched to `GAgentRegistryGAgent`.
- That means create-then-immediately-use flows can see a valid actor on the write side but a stale read model on the query side, causing false `404 ..._NOT_FOUND` during normal projection lag.

This is not just a naming issue. It shows that the current interface mixes three different semantics:
- command-side registration/unregistration
- query-side listing from a read model
- command-admission ownership checks for a target actor

Those three semantics need different contracts and consistency expectations.

## Problem

The current name and shape still suggest an old storage-style abstraction:
- `Store` naming hides that the write-side owner is `GAgentRegistryGAgent`.
- Query and command responsibilities are mixed into one interface.
- Callers can miss that reads depend on a projection read model and are eventually consistent.
- Endpoint authors can accidentally use read-model queries as synchronous authorization/admission checks.
- Tactical workarounds such as embedding scope hashes into actor IDs push ownership semantics into ID formatting, which conflicts with the architecture rule that `actorId` should remain an opaque address.

That makes the current architecture harder to read and easy to misuse, especially when reviewing scoped actor lifecycle, draft-run preparation/rollback, NyxID chat, StreamingProxy, or future scope-owned GAgent capabilities.

## Proposed direction

Refine the abstraction into explicit registry- and ownership-oriented ports.

### 1. Registry command port

Example shape:

```csharp
public interface IGAgentActorRegistryCommandPort
{
    Task RegisterActorAsync(
        string scopeId,
        string gagentType,
        string actorId,
        CancellationToken ct = default);

[...]

PR #452, opened 2026-04-27
Refactor GAgent registry ownership ports

louis4li:
## Summary

Refactors the legacy `IGAgentActorStore` surface into explicit GAgent registry command, query, and command-admission ports for issue #348.

## What Changed

- Adds `IGAgentActorRegistryCommandPort`, `IGAgentActorRegistryQueryPort`, and `IScopeResourceAdmissionPort` contracts.
- Replaces the actor-backed store implementation with `ActorBackedGAgentRegistryPorts` backed by `GAgentRegistryGAgent`.
- Adds typed registry admission messages and `GAgentRegistryOperation` enum.
- Moves list paths to eventually consistent registry read-model snapshots with `StateVersion`, `UpdatedAt`, and `ObservedAt` freshness fields.
- Keeps CLI and console-web list clients compatible with both the legacy array response and the new `{ groups, stateVersion, ... }` response shape; NyxID conversation list decoding also accepts both legacy array and wrapped `{ conversations, ... }` shapes.
- Moves NyxID chat, StreamingProxy, and draft-run target operations to registry admission instead of projection/list-based ownership checks.
- Enforces create-before-register semantics for immediately operable resources and adds rollback for registration failure/cancellation paths.
- Tightens rollback bookkeeping so registration cleanup is only attempted after registry registration returns successfully; if registration fails before acknowledgement, rollback destroys the newly-created actor directly.
- Disables direct generic `POST /{scopeId}/gagent-actors` registration with a 405 response so target creation stays in capability command endpoints.
- Adds `docs/canon/gagent-registry-ownership.md` and updates related lifecycle docs.
- Adds architecture guard coverage to prevent production references to the retired `IGAgentActorStore`/`ActorBackedGAgentActorStore` abstractions.

## Impact

This separates registry lifecycle writes, list queries, and command-path admission into distinct semantics. Command admission no longer depends on projection freshness, while list responses remain explicitly eventually consistent.

## Scope Boundaries

This PR is scoped to the GAgent registry ownership port split and the direct call-site fixes needed for issue #348: registry command/query/admission semantics, list response contract compatibility, create/register rollback behavior, and delete/unregister honesty for the affected registry-owned resources.

The following adjacent findings are intentionally tracked outside this PR:

- #372 tracks NyxID scoped endpoint caller-scope validation. Registry admission proves target ownership inside a scope; endpoint-level authorization that the caller may use that scope is a separate access-control fix.
- #462 tracks remaining ownership/admission paths that still depend on read models or actor ID shape: StudioMember bind admission, relay scope fallback by API key, and relay conversation actor IDs encoding scope ownership.
- #463 tracks moving StreamingProxy room lifecycle orchestration out of the HTTP endpoint and behind an application command surface. This PR fixes rollback correctness in the current endpoint path, but does not complete the Host/API composition-only refactor.

## Validation

Passed:

- `dotnet test test/Aevatar.AI.Tests/Aevatar.AI.Tests.csproj --no-restore --filter "FullyQualifiedName~NyxIdChatEndpointsCoverageTests|FullyQualifiedName~StreamingProxyEndpointsCoverageTests" --nologo`
- `dotnet test test/Aevatar.GAgentService.Tests/Aevatar.GAgentService.Tests.csproj --no-restore --filter "FullyQualifiedName~GAgentDraftRunActorPreparationServiceTests" --nologo`
- `npm test -- --run src/api.scope-services.test.ts` in `tools/Aevatar.Tools.Cli/Frontend`
- `npm test -- --runTestsByPath src/shared/api/runtimeGAgentApi.test.ts` in `apps/aevatar-console-web`
- `npm run build` in `tools/Aevatar.Tools.Cli/Frontend` (Vite chunk-size warning only)
- `npm run tsc` in `apps/aevatar-console-web`
- `bash tools/ci/playground_asset_drift_guard.sh`
- `git diff --check -- . ':(exclude)tools/Aevatar.Tools.Cli/wwwroot/app.js' ':(exclude)demos/Aevatar.Demos.Workflow.Web/wwwroot/app.js'`

Note: full `git diff --check` reports trailing whitespace inside generated bundled `app.js` output; source/non-bundle diff check is clean.


louis4li:
补充说明一下这轮 review 的处理边界：

- 前三项属于仍然存在、但不在这次 registry ports/dev diff 修复范围内的 ownership/admission 架构债务，已单独开 issue 跟踪：#462。
  - `StudioMemberService.BindAsync` 仍依赖 StudioMember projection readmodel 做 bind admission。
  - relay scope fallback 仍通过 ChannelBot registration projection 按 API key 反查 scope。
  - scoped relay conversation actorId 仍编码 scope ownership 信息。

- 后两项已在本 PR 内修复：
  - registry register receipt 不再把普通 dispatch ACK 直接声明为 `AdmissionVisible`；现在会显式验证 admission visibility，验证不到时降级为 `AcceptedForDispatch`。
  - unregister receipt 不再返回 `AdmissionVisible`，新增并返回 `AdmissionRemoved`，`IsAdmissionVisible` 对该阶段为 `false`。

同时补了 NyxID create conversation 的回滚逻辑：当注册没有达到 admission-visible 时，会 unregister registry entry 并 destroy actor，避免返回成功但留下半成品 actor/ownership 状态。

agents/Aevatar.GAgents.NyxidChat/NyxIdChatEndpoints.cs
```diff
@@ -81,54 +82,154 @@ public static IEndpointRouteBuilder MapNyxIdChatEndpoints(this IEndpointRouteBui
...
         var actorId = NyxIdChatServiceDefaults.GenerateActorId();
-        await actorStore.AddActorAsync(scopeId, NyxIdChatServiceDefaults.GAgentTypeName, actorId, ct);
+        await actorRuntime.CreateAsync<NyxIdChatGAgent>(actorId, ct);
+        var registrationAttempted = false;
+        try
+        {
+            var receipt = await registryCommandPort.RegisterActorAsync(
+                new GAgentActorRegistration(scopeId, NyxIdChatServiceDefaults.GAgentTypeName, actorId),
+                ct);
+            registrationAttempted = true;
```
eanzhao: **[P2] Rollback can leak an orphan registry entry.**

`registrationAttempted` is set *after* `RegisterActorAsync` returns. But `ActorBackedGAgentRegistryPorts.RegisterActorAsync` dispatches the `ActorRegisteredEvent` first and *then* runs `VerifyAdmissionVisibleAsync`. If `VerifyAdmissionVisibleAsync` throws `OperationCanceledException` (the only exception it doesn't swallow), `RegisterActorAsync` propagates it — at that point the registration **is committed** but `registrationAttempted == false`, so the catch on line 117 only destroys the actor and skips the unregister. The registry now holds an entry pointing at an actor that no longer exists.

Safer fix: drop the flag entirely and always call `UnregisterActorAsync` in the rollback. The actor's `HandleActorUnregistered` is already idempotent (`if (group is null || !group.ActorIds.Contains(...)) return;`), so an extra unregister when nothing was registered is a no-op. Same pattern repeats in `StreamingProxyEndpoints.cs:87` and `GAgentDraftRunActorPreparationService.cs:73`.

agents/Aevatar.GAgents.NyxidChat/NyxIdChatEndpoints.cs
```diff
@@ -81,54 +82,154 @@ public static IEndpointRouteBuilder MapNyxIdChatEndpoints(this IEndpointRouteBui
...
+            return Results.Ok(new
+            {
+                snapshot.ScopeId,
+                snapshot.StateVersion,
+                snapshot.UpdatedAt,
+                snapshot.ObservedAt,
+                Conversations = actorIds.Select(actorId => new { actorId }),
+            });
         }
         catch (InvalidOperationException)
```
eanzhao: **[P2] Swallowing `InvalidOperationException` as "empty list" hides real failures and is inconsistent with sibling endpoints.**

The sibling `ScopeGAgentEndpoints.HandleListActorsAsync` catches the same exception type and returns **400 `GAGENT_ACTOR_REGISTRY_ERROR`** with the original message. This handler returns **200 OK with `Conversations: []`**.

Net effect: a real registry-read failure looks to the user like "all your conversations are gone." If projection-not-yet-initialized is the case being defended against, model it explicitly in `ListActorsAsync` (return an empty snapshot from the implementation when the document isn't ready), and let unexpected exceptions bubble up. Catching `InvalidOperationException` by name is also brittle — it's not declared anywhere in the port contract.

src/Aevatar.Studio.Infrastructure/ActorBacked/ActorBackedGAgentRegistryPorts.cs
```diff
@@ -0,0 +1,246 @@
...
+                registration.ScopeId,
+                registration.ActorId);
+            return false;
+        }
+    }
+
+    private string NormalizeScopeId(string? scopeId) =>
+        string.IsNullOrWhiteSpace(scopeId)
+            ? _scopeResolver.ResolveScopeIdOrDefault()
+            : scopeId.Trim();
```
eanzhao: **[P3] Empty-`scopeId` fallback weakens the security boundary on an authoritative ownership write.**

`NormalizeScopeId("")` silently rewrites to `_scopeResolver.ResolveScopeIdOrDefault()`. All current call sites pass `scopeId` from the route, so this is dead defensively today — but the registry is the authoritative ownership source, and a future caller that forgets to pass scope will silently write/read against the resolver's default. Compare `NormalizeRequired` (line 220) which throws on empty for `GAgentType`/`ActorId`.

The canon doc itself flags this shape as forbidden: *"actor activation alone is not ownership evidence."* Equivalent for scope: implicit-resolution alone shouldn't be ownership evidence either. Suggest throwing `ArgumentException` for empty `scopeId`, matching `NormalizeRequired`.

eanzhao (commented):
## Review summary

Port split + admission contract is well-shaped and matches the new `docs/canon/gagent-registry-ownership.md` rules. Found one user-facing regression and a few correctness/observability concerns that are worth fixing before this lands.

### P1 — frontend regression (caller files not in this diff)

Direct `POST /api/scopes/{scopeId}/gagent-actors` is now `405 DIRECT_GAGENT_ACTOR_REGISTRATION_UNSUPPORTED` (deliberate, integration-tested), but the console-web caller wasn't updated:

- `apps/aevatar-console-web/src/pages/gagents/index.tsx:1791` — the actor-registry drawer's **Save** button still calls `runtimeGAgentApi.addActor(...)` → 405 → user sees an error toast on every save.
- `apps/aevatar-console-web/src/shared/api/runtimeGAgentApi.test.ts:121` — the test for `addActor` mocks `fetch` to return `ok:true, status:200` and asserts `resolves.toBeUndefined()`. It still passes, so CI doesn't catch the regression. Keeping a passing test for a deliberately-broken endpoint is worse than no test.

Fix path is one of:
1. delete `runtimeGAgentApi.addActor`, the **Save** button at `index.tsx:1791`, the drawer save flow, and this test; or
2. route the UI through a capability endpoint that creates the target (since direct registration is intentionally out of contract).

### P2 — correctness / observability

See inline comments on:
- rollback ordering in `NyxIdChatEndpoints.cs` (same pattern lives in `StreamingProxyEndpoints.cs:87` and `GAgentDraftRunActorPreparationService.cs:73`)
- `InvalidOperationException` swallow that hides registry errors as "empty conversation list"
- defense-in-depth: empty-string scope falling back to resolver default

### Things that look correct

- New canon doc (`docs/canon/gagent-registry-ownership.md`) is precise about admission-vs-list-vs-command separation, and the port split implements it.
- Architecture guard in `tools/ci/architecture_guards.sh +76` blocks future production refs to `IGAgentActorStore`/`ActorBackedGAgentActorStore`. Verified zero residual refs in `src/` or `agents/`.
- Create-then-register-then-verify-admission flow with explicit `AdmissionVisible` vs `AcceptedForDispatch` receipt is a meaningful honesty improvement over the previous fire-and-forget. `Unregister` correctly returns the new `AdmissionRemoved` stage and `IsAdmissionVisible` is `false` for it.
- 405 on direct `POST /{scopeId}/gagent-actors` is enforced by integration tests.

### Untested in this PR (flag, not blocking)

The admission protocol relies on `GAgentRegistryAdmissionNotFoundException` propagating raw through `IActorDispatchPort.DispatchAsync`. `ActorBackedStoreAdapterTests` uses fake dispatch ports that throw synchronously; nothing exercises the real Studio actor runtime end-to-end. If the production dispatch pipeline ever wraps exceptions, `AuthorizeTargetAsync` will silently downgrade every NotFound to `Unavailable` (503 instead of 404). One integration test against the real runtime asserting `Status == NotFound` for an unregistered target would close this gap.

### Style / informational

Both list decoders (`tools/Aevatar.Tools.Cli/Frontend/src/api.ts:809-833` and `apps/aevatar-console-web/src/shared/api/runtimeGAgentApi.ts:69-87`) accept the new `{ groups, stateVersion, updatedAt, observedAt }` snapshot shape but strip the freshness fields. The whole point of returning the snapshot version is to let the UI render staleness. Plumbing them through later is fine, but the value of the new contract is invisible to users until then.

agents/Aevatar.GAgents.NyxidChat/NyxIdChatEndpoints.cs
louis4li: Fixed in `8037a8e5`.

Dropped the `registrationAttempted` flag and made rollback attempt `UnregisterActorAsync` first in all three affected paths: NyxID chat, StreamingProxy room creation, and draft-run actor preparation. If unregister fails, rollback now stops before destroying the actor, so we avoid creating a registry entry that points at a destroyed actor. Added coverage for the post-commit throw/cancel case and the unregister-failure ordering.

agents/Aevatar.GAgents.NyxidChat/NyxIdChatEndpoints.cs
louis4li: Fixed in `8037a8e5`.

Removed the `InvalidOperationException` catch from the NyxID conversation list handler, so registry read failures now bubble instead of being rendered as `200 OK` with an empty list. Added a regression test asserting the failure is not swallowed.

src/Aevatar.Studio.Infrastructure/ActorBacked/ActorBackedGAgentRegistryPorts.cs
louis4li: Fixed in `8037a8e5`.

`ActorBackedGAgentRegistryPorts` now treats `scopeId` like the other authoritative ownership fields and throws on empty input instead of falling back to the ambient resolver. Added a test that empty scope fails and does not create a registry actor.

louis4li:
Addressed the actionable review feedback in `8037a8e5` and pushed to `refactor/2026-04-27_registry-ownership-ports`.

What changed:
- Removed direct frontend actor registration (`runtimeGAgentApi.addActor`, console-web Save actor flow, CLI frontend auto-persist call). The registry drawer now only lists/removes existing registry entries; actor creation stays behind capability/run paths.
- Made rollback ownership-first in NyxID chat, StreamingProxy, and draft-run actor preparation: always attempt unregister first; do not destroy the actor if unregister cannot be confirmed.
- Stopped swallowing NyxID registry list `InvalidOperationException` as an empty conversation list.
- Removed empty `scopeId` ambient fallback from `ActorBackedGAgentRegistryPorts`; empty scope now throws like other required ownership fields.

Verification run locally:
- `npm test -- --runTestsByPath src/pages/gagents/index.test.tsx src/shared/api/runtimeGAgentApi.test.ts`
- `npm run tsc` in `apps/aevatar-console-web`
- `npm test -- --run src/api.scope-services.test.ts` in `tools/Aevatar.Tools.Cli/Frontend`
- targeted .NET tests for NyxID/StreamingProxy, draft-run actor preparation, and actor-backed registry adapter
- `bash tools/ci/playground_asset_drift_guard.sh`
- `git diff --check`

Left as non-blocking follow-up per the review text: real runtime NotFound integration coverage and UI plumbing for snapshot freshness fields.

Issue #450, opened 2026-04-27
refactor(daily-prompt): versioned daily report prompt templates (prerequisite for #423 enrichment)


eanzhao:
> Architectural follow-up surfaced in [docs/audit-scorecard/2026-04-27-daily-pipeline-architecture-review.md](../blob/docs/2026-04-27_daily-pipeline-test-reference/docs/audit-scorecard/2026-04-27-daily-pipeline-architecture-review.md) §C3. Prerequisite for #423.

## Symptom

The daily report prompt template lives in code as a static `StringBuilder` ([AgentBuilderTemplates.cs:48-64](../blob/dev/agents/Aevatar.GAgents.ChannelRuntime/AgentBuilderTemplates.cs)). At agent creation time the rendered prompt string is frozen into `SkillRunnerState.SkillContent`. Implications:

- Updating the prompt requires a deploy.
- After deploy, **existing agents still use the old prompt** (frozen in their state).
- There is no way to roll out a new prompt to existing agents without `/delete-agent` + recreate.
- There is no concept of \"prompt version\" anywhere — `/agent-status` cannot tell you which prompt your daily is running on.

## Why now

issue #423 plans to substantially enrich the daily report (more sections, repo-aware queries, structured envelope). That work is gated on this: shipping a new prompt today silently leaves all existing daily-report agents on the old prompt indefinitely. The richer prompt also won't fit \"fix the string in code, redeploy\" cleanly.

## Architectural violations

- CLAUDE.md \"渐进演进: 开发期可用本地/内存实现，但生产语义必须能无缝迁移到分布式与持久化\". The prompt is a fact (versioned, queryable, owned), not a code constant.
- CLAUDE.md \"事实源唯一\". The same prompt content is currently in code AND mirrored into N agent states.

## Proposed direction

Introduce a `DailyReportTemplateCatalog` — actor-owned or document-store-backed:

- Each template version has `template_id`, `version`, `title`, `system_prompt`, `execution_prompt_template`, `created_at`, `description`.
- Subscription actor (post-#refactor-split-skill-runner) stores `template_id + template_version` instead of the rendered `skill_content` string.
- Run actor (post-split) resolves `template_id + template_version` → fetches current rendered prompt at run start.
- Two run modes:
  - **Pinned**: subscription explicitly pinned to a version; new versions don't affect it.
  - **Floating**: subscription tracks \"latest\" version of a template id; new template version applies on next run.
[...]

Issue #439, opened 2026-04-27
bug(daily): SkillRunner masks GitHub tool failures as silent "no activity" successful runs


eanzhao:
## Symptom

After `/daily eanzhao` (binding my own GitHub username), the very first daily run came back with:

```
GitHub Daily Update — eanzhao

- No meaningful public GitHub activity found for eanzhao in the last 24 hours.
- No recent authored commits surfaced in the checked window.
- No recently updated authored issues or PRs surfaced in the checked window.
- No recent issue or PR comments surfaced in the checked window.

No blockers.
```

But `eanzhao` actually had **52 commits + 114 authored issues/PRs + 32 comments** during the same 24h window. Verified by running the exact three search queries from the prompt directly against `api.github.com`:

| query | result |
| --- | --- |
| `/search/commits?q=author:eanzhao+author-date:>=2026-04-26` | 52 hits (PR #289 merge, "Address review" series, Telegram coverage, etc.) |
| `/search/issues?q=author:eanzhao+updated:>=2026-04-26` | 114 hits (PRs #427/#428/#289, issue #436 itself, charon work, …) |
| `/search/issues?q=commenter:eanzhao+updated:>=2026-04-26` | 32 hits |

`/agent-status skill-runner-de2c162c9d454cc8a4b64b190673d722` showed `Status: running`, `error_count: 0`, `last_error: ""` — i.e. the runner believed the execution had succeeded.

Re-running via `/run-agent` (Run Now button) ~15 minutes later returned the correct full report immediately, so the underlying GitHub access path was healthy. The first run was a transient infrastructure failure that surfaced as a **fake-success** in the agent's output, in `/agent-status`, and in the registry. Operator and end-user have no signal to investigate.

NyxID approval gate ruled out (no approval push received during the window).

## Root cause

Three collaborating layers all default to "swallow":

### 1. Prompt — `agents/Aevatar.GAgents.ChannelRuntime/AgentBuilderTemplates.cs:48-64`

```csharp
.AppendLine("Suggested GitHub proxy calls:")
.AppendLine("- GET /search/commits?q=author:{username}+author-date:>={iso_date}")
.AppendLine("- GET /search/issues?q=author:{username}+updated:>={iso_date}")
.AppendLine("- GET /search/issues?q=commenter:{username}+updated:>={iso_date}")
.AppendLine("If there is no meaningful activity, say so plainly instead of inventing progress.")
```

[...]

Issue #423, opened 2026-04-25
enhance(daily): richer report content + progressive delivery (streaming-edit or batched)


eanzhao:
## Summary

Day-one enhancement plan for `/daily`. The current report is functional after #421 closed the GitHub-403 root cause, but the content is thin and the delivery is one-shot — the user gets a single Lark message after the agent finishes, with no progress indication and no per-source breakdown. Two related areas to fix:

1. **Content depth** — the skill prompt at `agents/Aevatar.GAgents.ChannelRuntime/AgentBuilderTemplates.cs:48-64` only suggests three GitHub queries (commits authored, issues authored, issues commented) and asks for "3-6 concise bullet points + one blocker line". This produces something like a TL;DR, not a daily.
2. **Delivery UX** — `SkillRunnerGAgent.SendOutputAsync` (`agents/Aevatar.GAgents.ChannelRuntime/SkillRunnerGAgent.cs:254-300`) sends the entire LLM output in one `POST /open-apis/im/v1/messages` call. There is no streaming-edit equivalent of normal chat (`ChannelLlmReplyInboxRuntime` → `TurnStreamingReplySink` → `/channel-relay/reply/update`). The user sees nothing for ~30s, then a wall of text.

## Why current streaming doesn't transfer to SkillRunner

`TurnStreamingReplySink` works against the **reply token** baked into the inbound webhook payload — that token has a ~14-minute TTL and is bound to one specific user-initiated turn. Scheduled SkillRunner runs (the daily 9am cron, the manual `/run-agent`, retries) don't have a reply token; they're ambient. Trying to reuse `/channel-relay/reply/update` would land as `reply_token_missing_or_expired`.

The closest equivalent for SkillRunner is **edit-own-message** via Lark's `PATCH /open-apis/im/v1/messages/{message_id}` (text and card update endpoints already exist — the channel-runtime adapter uses them today for chat streaming). That gives SkillRunner the same "watch the message grow" UX without depending on a reply-token grant.

### A. Content depth (skill prompt + suggested data sources)

Rewrite `TryBuildDailyReportSpec` so the daily covers more surface area, with explicit structured sections and a length budget per section. Treat the new prompt as a **specification of what to fetch + how to summarize**, not a freeform creative brief.

Suggested sections (each with a hard ≤N-line budget, omitted entirely if empty rather than padded):

- **Shipped** — PRs merged in last 24h (title + repo + #PR), commits to default branch
[...]

PR #458, opened 2026-04-27
enhance(daily): structured-section prompt with omit-empty rule + repo-aware queries (#423 §A)

eanzhao:
## Summary

Implements **§A (content depth)** of #423. The current `/daily` prompt is a single
paragraph that asks for "3-6 concise bullet points" — produces thin reports and
pads when sources are silent. This PR rewrites `TryBuildDailyReportSpec` to treat
the system prompt as a **fetch-and-summarize specification**, not a freeform brief:

- **Explicit section order with per-section line budgets** — nine slots: Title (1) /
  Shipped (≤6) / In flight (≤6) / Reviews (≤4) / Issues (≤4) / CI (≤3, repo-mode
  only) / Trend (1, optional) / Blockers (1) / Source health (1, footer, conditional).
  §9 Source health is a real schema slot, not a freeform footer, so the `Blockers
  vs Source health` ordering is unambiguous when failures occur.
- **"Omit the section entirely if empty" rule** (header AND body) — the regression
  we're guarding against is the LLM padding silent sources with `no activity in this
  area` filler. The empty-day fallback (`Title + No measurable activity in the last
  24h.`) is gated on EVERY source returning 2xx; it explicitly does NOT fire when
  any source is unavailable.
- **Repo-aware query suggestions** — when `repositories=` is provided, the prompt
  steers the LLM to per-repo searches (`/search/issues?q=repo:{owner}/{repo}+...`,
  `/search/commits?q=repo:{owner}/{repo}+...`, `/repos/{owner}/{repo}/actions/runs`)
  and explicitly refuses collapsing the allowlist into a single global query. When
  the allowlist is empty, the suggestion list covers shipped PRs + commits, in
  flight, reviews, issues opened/closed/commented; CI is documented as omitted in
  no-repo mode (the global `/search/*` endpoints do not expose Actions run
  conclusions).
- **Source health distinguished from empty results** (refs #439) — 2xx with empty
  body is genuine zero data and the section is omitted; 4xx / 5xx / tool error
  envelopes (`{"error": true, ...}`, revoked OAuth grant, proxy timeout) are
  unavailable sources and surface in the §9 Source health footer. Silently masking
  credential expiration as "No measurable activity" was the failure mode #439 is
  about; the prompt now actively guards against it.
- **Substitution variables documented** — `{username}`, `{iso_date}`, and
  `{owner}/{repo}` are spelled out with concrete examples + a final "do not leave
  literal placeholders in outbound URLs" line.

Two new tests in `AgentBuilderToolTests` pin the nine-slot section schema, the
omit-empty rule, the §8 Blockers / §9 Source health ordering, the source-failure
classification, and the per-repo / no-repo query shapes so future copy edits
don't silently regress the spec.

## Out of scope (follow-ups for #423)

This PR addresses §A and the corresponding acceptance test. The remaining items
stay open under #423:

- **§B Option 2 (streaming-edit delivery)** — preferred per the issue. Needs a new
  `SkillRunnerStreamingReplySink` plus Lark `PATCH /open-apis/im/v1/messages/{id}`
  integration without a reply token. Larger surface; better as its own PR.
- **§B Option 1 (batched delivery)** — fallback shape; only worth landing if
  Option 2 hits a blocker.
- **§C failure-notification cross-tenant fallback** — `TrySendFailureAsync` going
  through the same `s/api-lark-bot` proxy that just rejected. Tracked in #423 §C and
  scorecard row C2; requires capturing a channel-bot fallback target at agent-create
  time.
- **`nyxid_proxy` tool error classification** — this PR teaches the prompt to
  surface non-2xx as §9 Source health, which is the user-facing half of #439, but
  the underlying `nyxid_proxy` envelope shape (`{"error": true, "status": ...}`)
  remains tracked there for the runtime-side fix.
- **Existing agents with frozen old prompt in actor state** — `SkillRunnerInitialized`
  freezes `SkillContent` at create time, so this PR only takes effect for newly
  created agents. The structural fix (versioned prompt templates) is tracked as #450.
  Migration path for users today: `/agents` → Delete → `/daily`.

The smoke test in the issue's acceptance ("commits authored empty but reviews
non-empty → Reviews renders, Shipped omitted") is a manual run-against-real-GitHub
check; the automated pinning test guards the prompt itself, not the LLM's adherence.

## Test plan

- [x] `dotnet test test/Aevatar.GAgents.ChannelRuntime.Tests/Aevatar.GAgents.ChannelRuntime.Tests.csproj` — 475 passed (32 in `AgentBuilderToolTests` including the two new pinning tests).
- [x] `bash tools/ci/architecture_guards.sh` — all guards pass.
- [ ] Manual smoke: create a fresh `/daily` agent, trigger via `/run-agent`, verify
  the output uses the new section structure and that empty sections are omitted.
- [ ] Manual smoke: revoke the GitHub grant for the user, trigger `/run-agent`,
  verify the output emits the §9 Source health footer instead of falling back to
  "No measurable activity in the last 24h." (this is the #439 regression guard).

Refs #423.

### DIFF
--- a/agents/Aevatar.GAgents.ChannelRuntime/AgentBuilderTemplates.cs
+++ b/agents/Aevatar.GAgents.ChannelRuntime/AgentBuilderTemplates.cs
@@ -131,9 +131,10 @@ internal static class AgentBuilderTemplates
             .AppendLine("5. Issues (≤4 lines) — issues opened, closed, or commented on by the user.")
             .AppendLine("6. CI (≤3 lines) — failing GitHub Actions runs on the tracked repos. Best-effort and only feasible in repo-allowlist mode; OMIT this section in no-repo mode (the global search endpoints do not expose Actions run conclusions).")
             .AppendLine("7. Trend (1 line, optional) — running totals vs the prior 24h, e.g. `Trend: shipped 3 (+1), reviews 5 (-2)`. Omit when the prior-window data could not be fetched.")
-            .AppendLine("8. Blockers (1 line, always last unless the report is empty-day) — `Blockers: <short list>` or `No blockers.` Auto-detect from: PRs >24h waiting on a review, CI red >2h, issues with labels `blocked` or `needs-info`.")
+            .AppendLine("8. Blockers (1 line) — `Blockers: <short list>` or `No blockers.` Auto-detect from: PRs >24h waiting on a review, CI red >2h, issues with labels `blocked` or `needs-info`. Position-locked at slot 8; the only section that may sit below it is the §9 Source health footer.")
+            .AppendLine("9. Source health (1 line, footer) — `Source health: <comma-separated list of unavailable sources with short reason>`. Emit ONLY when at least one source returned a non-2xx / error-shaped tool result. When emitted, this is always the final line — below Blockers, below everything.")
             .AppendLine()
-            .AppendLine("If the entire 24h window has no measurable activity across all sources, return ONLY the title line followed by `No measurable activity in the last 24h.` and nothing else (do not append a Blockers line in this case).")
+            .AppendLine("If EVERY source returned 2xx with no matching items (genuine empty day), return ONLY the title line followed by `No measurable activity in the last 24h.` and nothing else — do NOT emit Blockers or Source health. If ANY source failed, you are NOT on the empty-day path: emit at least the title line plus the §9 Source health footer (any other sections that have 2xx data render normally; §8 Blockers is also emitted).")
             .AppendLine("Do not invent activity. Do not paraphrase issue or PR titles into different wording. Keep each line short — Feishu text messages have a body cap, prefer trimming trailing detail over exceeding it.")
             .AppendLine()
             .AppendLine("# Suggested GitHub proxy calls")
@@ -171,12 +172,12 @@ internal static class AgentBuilderTemplates
 
         prompt
             .AppendLine()
-            .AppendLine("# Source health — distinguish empty results from source failures")
+            .AppendLine("# Source health — when to emit the §9 footer")
             .AppendLine()
             .AppendLine("Do NOT collapse transport, auth, or proxy failures into the empty-day fallback. Classify every tool result before mapping it to a section:")
-            .AppendLine("- 2xx with an empty list / no matching items → genuine zero data; the section is omitted per the schema.")
-            .AppendLine("- 4xx / 5xx / tool error envelope (e.g. `{\"error\": true, ...}`, revoked OAuth grant, proxy timeout) → the SOURCE is UNAVAILABLE, not zero. Append a final `Source health: <comma-separated list of unavailable sources with short reason>` line at the very bottom of the report (after Blockers).")
-            .AppendLine("- The empty-day fallback (`No measurable activity in the last 24h.`) is ONLY valid when EVERY source returned 2xx. If ANY source failed, emit at least the title plus a `Source health:` footer — silently masking credential expiration as `No measurable activity` is the bug we are guarding against.")
+            .AppendLine("- 2xx with an empty list / no matching items → genuine zero data; the section is omitted per the schema. Does NOT trigger §9.")
+            .AppendLine("- 4xx / 5xx / tool error envelope (e.g. `{\"error\": true, ...}`, revoked OAuth grant, proxy timeout) → the SOURCE is UNAVAILABLE, not zero. Add the source name + short reason to the §9 Source health footer.")
+            .AppendLine("- The empty-day fallback (`No measurable activity in the last 24h.`) is ONLY valid when EVERY source returned 2xx. If ANY source failed, you are NOT on the empty-day path — emit the title plus the §9 Source health footer at minimum. Silently masking credential expiration as `No measurable activity` is the bug we are guarding against.")
             .AppendLine("- Do not retry. Do not fall back to invented data. Do not leave any literal `{username}` / `{iso_date}` / `{owner}/{repo}` placeholders in outbound URLs.");
 
         return prompt.ToString();
--- a/agents/Aevatar.GAgents.ChannelRuntime/AgentBuilderTemplates.cs
+++ b/agents/Aevatar.GAgents.ChannelRuntime/AgentBuilderTemplates.cs
@@ -125,45 +125,53 @@ internal static class AgentBuilderTemplates
             .AppendLine("Each section has a hard line budget. If a section has zero data OR the source is unavailable, OMIT THE SECTION ENTIRELY (header and body) — do not pad with `no activity` or filler.")
             .AppendLine()
             .AppendLine("1. Title (1 line) — `Daily report — {username} — last 24h`.")
-            .AppendLine("2. Shipped (≤6 lines) — PRs merged in the last 24h and commits to the default branch. Format `- [owner/repo#NNN] title`.")
+            .AppendLine("2. Shipped (≤6 lines) — PRs merged AND commits authored by the user in the window. Format `- [owner/repo#NNN] title` for PRs, `- [owner/repo@sha7] subject` for commits.")
             .AppendLine("3. In flight (≤6 lines) — open PRs authored by the user. Append `(stale)` when the PR has had no activity for >24h.")
             .AppendLine("4. Reviews (≤4 lines) — PRs the user reviewed in the window. Include kind counts, e.g. `approved 2 / requested-changes 1 / commented 3`.")
             .AppendLine("5. Issues (≤4 lines) — issues opened, closed, or commented on by the user.")
-            .AppendLine("6. CI (≤3 lines) — failing builds on the default branch of the tracked repos (omit entirely when none are red).")
+            .AppendLine("6. CI (≤3 lines) — failing GitHub Actions runs on the tracked repos. Best-effort and only feasible in repo-allowlist mode; OMIT this section in no-repo mode (the global search endpoints do not expose Actions run conclusions).")
             .AppendLine("7. Trend (1 line, optional) — running totals vs the prior 24h, e.g. `Trend: shipped 3 (+1), reviews 5 (-2)`. Omit when the prior-window data could not be fetched.")
-            .AppendLine("8. Blockers (1 line, always last) — `Blockers: <short list>` or `No blockers.` Auto-detect from: PRs >24h waiting on a review, CI red >2h, issues with labels `blocked` or `needs-info`.")
+            .AppendLine("8. Blockers (1 line, always last unless the report is empty-day) — `Blockers: <short list>` or `No blockers.` Auto-detect from: PRs >24h waiting on a review, CI red >2h, issues with labels `blocked` or `needs-info`.")
             .AppendLine()
-            .AppendLine("If the entire 24h window has no measurable activity across all sources, return ONLY the title line followed by `No measurable activity in the last 24h.` and nothing else.")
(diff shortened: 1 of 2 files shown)


eanzhao:
Reviewing §A changes. A few issues:

---

### 1. Test gap: Section 7 (Trend) not asserted despite claiming full coverage

`AgentBuilderToolTests.cs` — the pinning test says *"Structural sections must all be named in order"* but only asserts sections 1–6 and 8, skipping section 7 entirely:

https://github.com/Aevatar-AI/aevatar/blob/feat/2026-04-27_daily-richer-content-prompt/test/Aevatar.GAgents.ChannelRuntime.Tests/AgentBuilderToolTests.cs#L75-L85

If the omission is intentional (Trend is optional), the comment should say so. Otherwise, add:

```csharp
skillContent.Should().Contain("7. Trend");
```

---

### 2. No CI query guidance when repo allowlist is empty

The output schema defines **§6 CI** but the no-repo query suggestion list has zero GitHub Actions / check-run endpoints:

https://github.com/Aevatar-AI/aevatar/blob/feat/2026-04-27_daily-richer-content-prompt/agents/Aevatar.GAgents.ChannelRuntime/AgentBuilderTemplates.cs#L142-L149

The per-repo branch correctly includes `/repos/{owner}/{repo}/actions/runs`, but the global-search branch has nothing equivalent. The LLM has no guidance on how to populate the CI section in the no-repo case, so it will either skip it or hallucinate. Either add a search endpoint for CI (e.g. `/search/commits?q=...+status:failure` as a rough proxy) or explicitly note that CI is only available when repos are pinned.

---

### 3. `{default_branch}` substitution variable undocumented

In the per-repo query list, the CI query uses `{default_branch}` but it's never explained alongside `{iso_date}` and `{username}`:

https://github.com/Aevatar-AI/aevatar/blob/feat/2026-04-27_daily-richer-content-prompt/agents/Aevatar.GAgents.ChannelRuntime/AgentBuilderTemplates.cs#L154

The LLM may leave the literal `{default_branch}` in the URL or guess `main`/`master`. Consider either documenting the substitution (e.g. *"replace `{default_branch}` with the repo's default branch — usually `main` or `master`"*) or switching to the unfiltered `/actions/runs?per_page=10` and filtering client-side.

---

Otherwise the extraction into `BuildDailyReportSkillPrompt` and the omit-empty semantics look solid.

agents/Aevatar.GAgents.ChannelRuntime/AgentBuilderTemplates.cs
```diff
@@ -122,6 +102,72 @@ public static bool TryBuildSocialMediaSpec(
...
+                .AppendLine("- GET /search/issues?q=reviewed-by:{username}+updated:>={iso_date}                // reviews")
+                .AppendLine("- GET /search/issues?q=author:{username}+is:issue+created:>={iso_date}            // issues opened")
+                .AppendLine("- GET /search/issues?q=author:{username}+is:issue+is:closed+closed:>={iso_date}   // issues closed")
+                .AppendLine("- GET /search/issues?q=commenter:{username}+updated:>={iso_date}                  // issues commented");
+        }
+        else
+        {
+            prompt
+                .AppendLine("Repository allowlist provided — run these per-repo (replace `{owner}/{repo}` with each entry, substitute `{iso_date}` with the start of the 24h window in ISO 8601 UTC). Do NOT collapse into one global query.")
+                .AppendLine("- GET /repos/{owner}/{repo}/pulls?state=closed&per_page=20               // filter merged_at >= {iso_date} client-side for shipped")
```
eanzhao: P2: This endpoint is not a reliable 24h merge-time source. `/repos/{owner}/{repo}/pulls?state=closed&per_page=20` returns only one page of closed PRs, and the list API is not keyed by `merged_at`; filtering that small page client-side can miss PRs merged in the last 24h when they were created earlier or fall past the first page in active repos. Since this feeds the Shipped section, repo-scoped reports can undercount real shipped work. Prefer a query source keyed by merge time, e.g. `repo:{owner}/{repo} author:{username} is:pr is:merged merged:>={iso_date}`, or explicitly paginate/sort enough to make the client-side `merged_at` filter complete.

eanzhao (commented):
Found one additional issue while reviewing PR 458.

eanzhao:
Additional review — supplementary to the three issues already raised:

---

### 4. `(always last)` for §8 Blockers conflicts with empty-day fallback

Line 134 in `AgentBuilderTemplates.cs`:

```
.AppendLine("8. Blockers (1 line, always last) — ...")
```

But two lines later (136) the empty-day rule says:

```
return ONLY the title line followed by `No measurable activity in the last 24h.` and nothing else.
```

On an empty day, the LLM is told both (a) "Blockers is always last" and (b) "return nothing else". A weaker model could synthesise both and emit Blockers anyway. Consider qualifying Blockers as: `(always last, unless the entire report is empty-day)`.

---

### 5. Per-repo `?state=closed` shipped query pulls closed-but-not-merged PRs

Line 157:

```
.AppendLine("- GET /repos/{owner}/{repo}/pulls?state=closed&per_page=20               // filter merged_at >= {iso_date} client-side for shipped")
```

`?state=closed` returns PRs closed without merge (e.g. rejected). The instruction says "filter merged_at" but does not call out that a null `merged_at` means "not shipped". Without explicit guidance, the LLM may count closed-unmerged PRs in the Shipped section. Suggestion: add a note like `"exclude PRs where merged_at is null"` or switch to `GET /repos/{owner}/{repo}/pulls?state=closed&sort=updated&direction=desc` with `"only count PRs whose merged_at falls in the window"`.

---

### 6. `{username}` substitution not documented alongside `{iso_date}` / `{owner}/{repo}`

In the no-repo query suggestions (lines 146–151), `{username}` appears in every URL but the substitution instruction (line 145) only documents `{iso_date}`. Given that issue #3 flags `{default_branch}` as undocumented, `{username}` has the same problem. Add a brief mention, e.g.:

```
(substitute `{username}` with {normalizedUser}, `{iso_date}` with the start of the 24h window in ISO 8601 UTC)
```

---

Otherwise the extraction into `BuildDailyReportSkillPrompt` and the omit-empty / section-budget structure are clean.

eanzhao:
All six issues from the previous review round are addressed in this update. Verified:

1. ✅ **Section 7 (Trend) test gap** — now asserted, comment explains the rationale
2. ✅ **No-repo CI guidance** — section 6 explicitly states "OMIT this section in no-repo mode", no-repo query list includes explicit skip comment
3. ✅ **`{default_branch}` removed** — CI query now uses unfiltered `/actions/runs?per_page=10` with client-side filtering; test asserts `NotContain("{default_branch}")`
4. ✅ **Blockers vs empty-day conflict** — §8 now says "always last **unless the report is empty-day**"; empty-day rule says "do not append a Blockers line in this case"; test pins both phrases
5. ✅ **Closed-but-not-merged PRs** — replaced `/pulls?state=closed` with search-API `is:pr+is:merged+merged:>=`; test asserts new query and `NotContain` on old one
6. ✅ **`{username}` substitution** — new substitution-variable line documents `{username}`, `{iso_date}`, and `{owner}/{repo}`; test pins all three

Bonus fixes caught in this revision are also solid: shipped-commits query added for both modes, repo-mode issues now include the commenter case, final instruction blocks literal placeholder leakage.

LGTM — no remaining issues from my side.

agents/Aevatar.GAgents.ChannelRuntime/AgentBuilderTemplates.cs
```diff
@@ -122,6 +102,80 @@ public static bool TryBuildSocialMediaSpec(
...
+                .AppendLine("- GET /search/issues?q=repo:{owner}/{repo}+reviewed-by:{username}+updated:>={iso_date}              // reviews")
+                .AppendLine("- GET /search/issues?q=repo:{owner}/{repo}+author:{username}+is:issue+updated:>={iso_date}          // issues authored (created/closed)")
+                .AppendLine("- GET /search/issues?q=repo:{owner}/{repo}+commenter:{username}+is:issue+updated:>={iso_date}       // issues commented")
+                .AppendLine("- GET /repos/{owner}/{repo}/actions/runs?per_page=10                                                // CI: filter `conclusion=failure` and `created_at >= {iso_date}` client-side; do NOT add a `branch=` filter (default branch varies; trim noise client-side instead)");
+        }
+
+        prompt
+            .AppendLine()
+            .AppendLine("If a query returns 4xx, 5xx, or empty, treat that source as zero for the affected section and continue — do not retry, do not fall back to invented data. Do not leave any literal `{username}` / `{iso_date}` / `{owner}/{repo}` placeholders in outbound URLs.");
```
eanzhao: P1: Do not collapse transport/auth failures into zero activity. `nyxid_proxy` currently returns error-shaped JSON for upstream failures (`{"error": true, "status": 401/5xx, ...}`), and #439 is specifically about the runner/LLM masking those as a successful "no activity" daily. This instruction preserves that failure mode by telling the model to treat 4xx/5xx as zero and continue, so a revoked GitHub grant or proxy outage can still produce `No measurable activity` with the agent left healthy. Please distinguish successful empty responses from tool/proxy failures: 2xx empty can be zero; 4xx/5xx/error-shaped tool results should be surfaced as a source failure (or at least not allowed to trigger the empty-day fallback).

eanzhao (commented):
The previous fix set looks mostly covered. I found one remaining issue around error handling semantics.

agents/Aevatar.GAgents.ChannelRuntime/AgentBuilderTemplates.cs
```diff
@@ -122,6 +102,86 @@ public static bool TryBuildSocialMediaSpec(
...
+            .AppendLine("3. In flight (≤6 lines) — open PRs authored by the user. Append `(stale)` when the PR has had no activity for >24h.")
+            .AppendLine("4. Reviews (≤4 lines) — PRs the user reviewed in the window. Include kind counts, e.g. `approved 2 / requested-changes 1 / commented 3`.")
+            .AppendLine("5. Issues (≤4 lines) — issues opened, closed, or commented on by the user.")
+            .AppendLine("6. CI (≤3 lines) — failing GitHub Actions runs on the tracked repos. Best-effort and only feasible in repo-allowlist mode; OMIT this section in no-repo mode (the global search endpoints do not expose Actions run conclusions).")
+            .AppendLine("7. Trend (1 line, optional) — running totals vs the prior 24h, e.g. `Trend: shipped 3 (+1), reviews 5 (-2)`. Omit when the prior-window data could not be fetched.")
+            .AppendLine("8. Blockers (1 line, always last unless the report is empty-day) — `Blockers: <short list>` or `No blockers.` Auto-detect from: PRs >24h waiting on a review, CI red >2h, issues with labels `blocked` or `needs-info`.")
```
eanzhao: P2: This still says Blockers is always last except for empty-day reports, but the new source-health rule later requires `Source health:` at the very bottom after Blockers when any source fails. That gives the model two incompatible ordering rules in the exact case we care about for #439, and a weaker run can either drop the source-health footer or place it before Blockers. Please qualify this as something like `always last among normal report sections unless empty-day or Source health is emitted`, or make Source health an explicit final section/footer in the section order and pin that in the test.

eanzhao (commented):
The source-health fix addresses the P1 behavior, but it leaves one prompt-order conflict.

eanzhao:
One non-code follow-up: the PR summary still says `4xx/5xx/empty → treat as zero`, but the latest implementation intentionally does the opposite for non-2xx / error-shaped tool results by emitting §9 `Source health`. Please update that bullet so the PR description matches the current behavior and does not re-document the #439 failure mode as intended behavior.

PR #469, opened 2026-04-28
enhance(daily): streaming-edit Lark delivery for SkillRunner (#423 §B)

eanzhao:
## Summary

Implements **#423 §B Option 2 (streaming-edit delivery)** so the daily report
*lands and grows in place* instead of arriving as one wall of text after
~30 s of silence.

- New `SkillRunnerStreamingReplySink` (modeled on `TurnStreamingReplySink`)
  sends the first non-empty LLM delta as a Lark `POST /open-apis/im/v1/messages`,
  captures `data.message_id`, then `PATCH /open-apis/im/v1/messages/{id}` for
  every later delta. The endpoint is reached through the same `s/api-lark-bot`
  proxy `SendOutputAsync` already uses — no new NyxID API surface required.
- `SkillRunnerGAgent.ExecuteSkillAsync` now constructs the sink, drives
  `OnDeltaAsync` from the existing chat-stream loop, and `FinalizeAsync` at the
  end. The redundant trailing `SendOutputAsync(output)` call is dropped from
  `HandleTriggerAsync`; the one-shot path stays for `TrySendFailureAsync` and
  as a fallback when streaming can't be configured.
- Throttle 300 ms (`SkillRunnerDefaults.StreamingEditThrottle`) ⇒ ~3.3
  edits/sec, well under Lark's 5/sec cap. Inside-window deltas collapse onto
  the latest text via the existing deferred-flush-timer + reflush-on-conflict
  pattern.
- 230002 fallback retry preserved on the initial POST so the cross-app
  same-tenant recovery from PR #412 doesn't regress. Edits past the initial
  POST do **not** fall back — same target, retry on next delta.
- Mid-stream POST/PATCH failures log + return; only finalize-time failures
  throw, matching the existing `HandleTriggerAsync` retry/persist contract.
  Reuses `BuildLarkRejectionMessage` so cross-tenant / cross-app recovery
  hints surface identically.
- 30 K-char length cap with a `…[truncated]` marker. Lark's actual text
  limit is much higher, but the JSON wrapper plus UTF-8 multi-byte expansion
  makes 30 K a safe ceiling.

12 new tests in `SkillRunnerStreamingReplySinkTests` pin the POST→PATCH
handoff, message_id capture, throttle window collapse, finalize-bypasses-
throttle, single-POST fallback when no deltas streamed, the 230002 fallback
retry, cross-tenant rejection at finalize, mid-stream rejection swallow +
next-delta retry, length-cap truncation, and dedupe of repeated text.

## Out of scope (follow-ups, still under #423)

- **§C failure-notification cross-tenant fallback** — `TrySendFailureAsync`
  still goes through the same `s/api-lark-bot` proxy, so the original
  99992364-style rejection blackout remains. Needs agent-create-time
  channel-bot fallback capture; separate PR.
- **§C section-boundary chunked delivery for >30 KB reports** — truncation
  is in; section-boundary multi-message split would let richer reports land
  in full. The 9-section schema from #458 §A is the right anchor for "where
  to split"; tracked as a follow-up.
- **Existing agents with frozen old prompt in actor state** — `OutboundConfig`
  is fixed at create time, so the new streaming-edit path takes effect for
  newly created agents. Existing agents keep working through the one-shot
  fallback (their `OutboundConfig` is complete; they get streaming-edit on
  the next run because the sink resolves at execute time, not at create
  time). Migration is no-op.

## Test plan

- [x] `dotnet test test/Aevatar.GAgents.ChannelRuntime.Tests/...` — 493 passed
  (12 new sink tests + 9 existing `SkillRunnerGAgentTests` regression).
- [x] `dotnet test test/Aevatar.AI.Tests/...` — 467 passed.
- [x] `bash tools/ci/architecture_guards.sh` — all guards pass.
- [x] `bash tools/docs/lint.sh` — 31 files, 0 errors.
- [ ] Manual smoke: create a fresh `/daily` agent, trigger via `/run-agent`,
  verify the Lark message lands within ~1 s and grows in place as the LLM
  streams sections.
- [ ] Manual smoke: trigger `/run-agent` against an agent whose
  `LarkReceiveId` is bot-not-in-chat, verify the initial POST falls back to
  `LarkReceiveIdFallback` (single message, not duplicates).

Refs #423.


agents/Aevatar.GAgents.ChannelRuntime/SkillRunnerStreamingReplySink.cs
```diff
@@ -0,0 +1,535 @@
...
+        var body = JsonSerializer.Serialize(new
+        {
+            content = JsonSerializer.Serialize(new { text }),
+        });
+
+        var response = await _client.ProxyRequestAsync(
+            _nyxApiKey,
+            _nyxProviderSlug,
+            $"open-apis/im/v1/messages/{Uri.EscapeDataString(platformMessageId)}",
+            "PATCH",
```
eanzhao: 这里用 `PATCH /open-apis/im/v1/messages/{message_id}` 会走 Lark 的「更新已发送卡片」接口，只支持 `interactive` card。当前初始 POST 发的是 `msg_type = "text"`，所以后续编辑会被 Lark 拒绝，streaming-edit 不会增长，finalize 还会把这次 run 标成失败/触发重试。

文本/富文本消息编辑应使用 `PUT /open-apis/im/v1/messages/{message_id}`，body 带 `msg_type` 和 `content`（[官方 Edit message 文档](https://open.feishu.cn/document/server-docs/im-v1/message/update)）。如果要保留 `PATCH`，初始消息也需要改成 `interactive` card 并发送 card JSON。建议把测试也改成校验真实方法和 body，避免 mock 放过这个协议差异。

eanzhao (commented):
发现一个会阻断 streaming-edit 的协议问题，见 inline。

eanzhao (commented):
## Code Review — PR #469 (streaming-edit Lark delivery)

Solid implementation overall. The sink design cleanly mirrors `TurnStreamingReplySink`, the throttle/reflush-on-conflict/drain patterns are correct, and the 12 tests cover the key behaviors well. A few observations:

---

### 1. Test gap: Nyx envelope error path (minor)

All sink tests use `HttpStatusCode.OK`, testing only the Lark business error path (HTTP 200 with `code ≠ 0`). The Nyx `{"error":true,"status":400,"body":"{\"code\":230002,...}"}` envelope path tested in `SkillRunnerGAgentTests` (e.g. `SendOutputAsync_ShouldRetryWithFallback_When_PrimaryRejectedAsBotNotInChat_ViaHttp400Envelope`) is not replicated in the streaming sink tests. The code path is correct since the sink calls `LarkProxyResponse.TryGetError` which handles both, but adding a corresponding test would match the existing regression coverage.

### 2. `_ = DispatchLoopAsync(...)` in timer callback + final-text path

In `TurnStreamingReplySink`, the timer fire-and-forget cannot surface unobserved exceptions because `DispatchOneAsync` catches all non-cancellation errors and never throws.

In the new `SkillRunnerStreamingReplySink`, the `DispatchLoopAsync` catch block re-throws after signaling `_drainTcs`. If `FinalizeAsync` stashes the final text while a timer-initiated dispatch is in progress, the dispatch loop picks it up with `currentIsFinal = true`. If that final dispatch then fails, `DispatchLoopAsync` throws → the fire-and-forget task is faulted and the exception becomes unobserved (`TaskScheduler.UnobservedTaskException`).

The caller of `FinalizeAsync` is correctly notified through `drainTask`, so correctness is fine. But the unobserved exception is noise that `TurnStreamingReplySink` never generates.

**Suggestion**: Add a `ContinueWith` handler on the fire-and-forget in `OnFlushTimerFired` to log final-text failures from the timer path, or assert that `_drainTcs` is always null when the timer fires.

### 3. `FlushAsync` clears `_hasPending` when text equals `_lastEmittedText` (design check — no bug)

```csharp
if (string.Equals(capped, _lastEmittedText, StringComparison.Ordinal))
{
    _pendingText = string.Empty;
    _hasPending = false;
    return;
}
```

When `_dispatchInProgress` is true, this check clears any pending text without creating `_drainTcs`. This is correct because the caller always passes the full accumulated text, so `capped == _lastEmittedText` implies the pending stash is redundant. Worth leaving a comment explaining the invariant.

---

### Summary

No correctness issues. The two actionable items (test gap, timer unobserved exception) are minor. Approve after considering the above.

agents/Aevatar.GAgents.ChannelRuntime/SkillRunnerStreamingReplySink.cs
```diff
@@ -0,0 +1,535 @@
...
+
+            _dispatchInProgress = true;
+            toDispatch = _pendingText;
+            _pendingText = string.Empty;
+            _hasPending = false;
+        }
+
+        if (toDispatch is not null)
+        {
+            _ = DispatchLoopAsync(toDispatch, firstIsFinal: false, CancellationToken.None);
```
eanzhao: **Unobserved exception risk.** `DispatchLoopAsync` 的 catch block 会 `throw;`，所以任何逃出 `DispatchOneAsync` 的异常都会成为 unobserved task exception。

`DispatchOneAsync` 的 catch filter 是 `!isFinal && ex is not OperationCanceledException`。Timer 触发的 dispatch `isFinal=false`，但 `OperationCanceledException`（及其子类 `TaskCanceledException`）不被拦截。`HttpClient` 超时恰好抛的是 `TaskCanceledException`，所以 **HTTP timeout 会直接穿透到 unobserved**。

建议至少在 call site 包一层：
```csharp
_ = DispatchLoopAsync(toDispatch, firstIsFinal: false, CancellationToken.None)
    .ContinueWith(t => _logger?.LogWarning(t.Exception!.InnerException!, "..."),
        TaskContinuationOptions.OnlyOnFaulted);
```

（注意到 `TurnStreamingReplySink` 有同样的问题，但既然这里是新增代码，顺手修了比较好。）

agents/Aevatar.GAgents.Scheduled/SkillRunnerGAgent.cs
```diff
@@ -241,14 +245,81 @@ private async Task<string> ExecuteSkillAsync(DateTimeOffset now, string? reason,
...
+        try
         {
-            if (!string.IsNullOrEmpty(chunk.DeltaContent))
+            await foreach (var chunk in ChatStreamAsync(prompt, requestId, metadata, ct))
+            {
+                if (string.IsNullOrEmpty(chunk.DeltaContent))
+                    continue;
                 content.Append(chunk.DeltaContent);
+                if (sink is not null)
+                    await sink.OnDeltaAsync(content.ToString(), ct);
```
eanzhao, line 267: **Per-delta O(n) allocation.** 每次 delta 都 `content.ToString()` 拷贝整个累积文本。对于最终 30K 字符的报告（几百个 delta），总分配量是 O(n²)。

Daily report 场景下影响不大（频率低、总文本量有限），但如果后续想复用这个 sink 模式，值得改成传递 `ReadOnlyMemory<char>` 或让 sink 持有 StringBuilder 引用、自行决定何时 snapshot。

目前作为 daily report 专用路径可以接受，留个 TODO 即可。

agents/Aevatar.GAgents.Scheduled/SkillRunnerGAgent.cs
```diff
@@ -241,14 +245,81 @@ private async Task<string> ExecuteSkillAsync(DateTimeOffset now, string? reason,
...
+    /// Constructs the streaming-edit sink for this run, or returns null when streaming cannot be
+    /// configured. The sink writes the first non-empty delta as a Lark
+    /// <c>POST /open-apis/im/v1/messages</c> (capturing the returned <c>message_id</c>) and edits
+    /// the same message via <c>PATCH /open-apis/im/v1/messages/{id}</c> for every later delta —
+    /// so the user sees the daily report land and grow in place rather than receiving one wall of
+    /// text after the LLM finishes.
+    /// </summary>
+    private SkillRunnerStreamingReplySink? TryCreateStreamingSink()
+    {
+        var client = _nyxIdApiClient ?? Services.GetService<NyxIdApiClient>();
```
eanzhao, line 303: `Services.GetService<T>()` 是 service locator 反模式。虽然同文件内已有先例（L507/L542 的 `IActorRuntime`），但 `NyxIdApiClient` 是 streaming path 的核心依赖——如果它在 DI 里没注册，sink 构造就会走到 `null` fallback，静默退化为 one-shot，没有告警。

建议至少加一条 `_logger?.LogWarning(...)` 表明 streaming 被降级了，或者直接在 `_nyxIdApiClient` 字段赋值时（构造函数或 ActivateAsync）就 resolve，fail fast 而不是运行时才发现。

agents/Aevatar.GAgents.Scheduled/SkillRunnerStreamingReplySink.cs
```diff
@@ -0,0 +1,535 @@
...
+/// close to publish it. Multiple deltas in the same window collapse onto the latest text.</item>
+/// <item>While a dispatch is in flight, additional deltas (caller-driven or timer-driven) are
+/// likewise stashed; the dispatch loop reflushes the most recent <c>_pendingText</c> after the
+/// in-flight chunk completes (reflush-on-conflict).</item>
+/// <item><see cref="FinalizeAsync"/> bypasses the throttle so the actor sees the complete text
+/// once the stream ends; if a dispatch is in flight, the final text reflushes after it and
+/// <see cref="FinalizeAsync"/> awaits the dispatch loop's drain signal before returning.</item>
+/// </list>
+/// </para>
+/// <para>
```
eanzhao, line 40: `_nyxApiKey` 以 plain string 存活整个 streaming session（可能持续几十秒）。如果 logger 或 exception message 意外引用到它就会泄露。建议考虑 `System.Security.SecureString` 或至少在 class doc 里标注 **never log this field** 的约束。

（同项目里其他地方也这么存的，severity 不高，只是安全 hygiene 建议。）

eanzhao (commented):
整体看下来设计和实现都很扎实——POST/PATCH 手off、throttle 折叠、finalize-bypass、fallback 重试、截断，边界语义都很清晰，测试覆盖也充分。几个我觉得值得处理的点列在 inline comments 里。最关键的是 **fire-and-forget 里 `OperationCanceledException`/`TaskCanceledException` 会成为 unobserved exception**（HTTP timeout 走 `TaskCanceledException`，不被 `DispatchOneAsync` 的 filter 拦截），这个应该至少在 call site 包一层 ContinueWith 或 try-catch 记日志。

eanzhao:
Pushed `7b55e2a4` + `df4da836` addressing review feedback. Quick map:

### P1 (correctness blockers)

**eanzhao on line 483 — Lark text edit needs PUT, not PATCH** — fixed. PATCH on `/im/v1/messages/{id}` is reserved for editing interactive cards; text/post (rich text) edits go through PUT with `msg_type` + `content` in the body. Verified against [the Lark Edit message docs](https://open.feishu.cn/document/server-docs/im-v1/message/update). Without this fix every PATCH would have been rejected and streaming-edit would have stopped growing past the placeholder. Tests now assert `HttpMethod.Put` and `msg_type:"text"` in the body. The canon doc explanation in §8 is also updated to call out the PUT-vs-PATCH split explicitly.

**chatgpt-codex on line 352 — TaskCanceledException leak** — fixed. New `IsTransientFailure(ex, ct)` helper distinguishes caller-requested cancel (propagate) from `HttpClient` request-timeout `TaskCanceledException` (transient retry). Both initial-POST and edit paths use it. Without this, a transport blip mid-stream would have aborted the whole run instead of being retried on the next delta.

### Cleanup follow-ups

**eanzhao on line 274 — unobserved exception on timer fire-and-forget** — fixed. Added `.ContinueWith(OnlyOnFaulted)` logging on the `OnFlushTimerFired` dispatch. The drain TCS path still notifies `FinalizeAsync` callers as before; the continuation just keeps `TaskScheduler.UnobservedTaskException` quiet.

**eanzhao on line 291 — service-locator silent downgrade** — fixed. `TryCreateStreamingSink` now emits a `LogWarning` when it falls back to one-shot (no client, or incomplete outbound config). Otherwise streaming-edit silently never engages and the only symptom is the wall-of-text UX users complained about.

**eanzhao on line 39 — `_nyxApiKey` "never log"** — added a class-level comment pinning the constraint. Both call sites already pass the value directly into `ProxyRequestAsync` without echoing.

### Test gap + invariant comment

**Nyx HTTP-400 envelope path test** — added `InitialPost_RejectedAsBotNotInChat_ViaHttp400Envelope_RetriesOnceWithFallbackTarget`, mirroring the existing non-streaming `SkillRunnerGAgentTests.SendOutputAsync_ShouldRetryWithFallback_When_PrimaryRejectedAsBotNotInChat_ViaHttp400Envelope`. `SequencedHandler` extended to support `(HttpStatusCode, string)` tuples so the same fixture covers both the HTTP-200 plain-Lark-error path AND the wrapped envelope.

**FlushAsync equality-clears-pending invariant** — added a comment explaining why clearing `_pendingText` when `capped == _lastEmittedText` is safe even mid-dispatch (callers always pass full accumulated text, so the pending stash can only be an older or equal-length prefix).

### Acknowledged but accepted as-is

**eanzhao on line 257 — per-delta O(n²) allocation** — acceptable for daily-report-sized output (≤30 KB capped, sink dedupes against `_lastEmittedText`). Added a comment capturing the tradeoff and the migration path (`(StringBuilder, Range)` snapshot or `ReadOnlyMemory<char>`) for a future skill with materially longer output.

### Tests + lint

- `dotnet test test/Aevatar.GAgents.ChannelRuntime.Tests/...` — **494 passed** (13 sink tests + 9 existing `SkillRunnerGAgentTests` regression).
- `bash tools/docs/lint.sh` — 31 files, 0 errors.

eanzhao (commented):
四个 review comment 都已 addressed：

1. **P0 fire-and-forget** → `ContinueWith(OnlyOnFaulted)` + `IsTransientFailure` 区分 caller-cancel vs transport-timeout（`TaskCanceledException` 不再穿透成 unobserved）。👍
2. **P2 service locator 降级静默** → 两个 `Logger.LogWarning` 覆盖 client-null 和 config-incomplete 两条路径。👍
3. **P2 `content.ToString()` O(n²)** → 内联注释 + TODO 标注 future `(StringBuilder, Range)` 切换。👍
4. **P3 `_nyxApiKey` doc** → `// ... Treat as a secret: NEVER include it in log messages` 约束。👍

额外发现并修了 **PATCH → PUT** 的重要 bug（Lark 按 msg_type 分动词，text 用 PUT 不是 PATCH），补了 HTTP-400 envelope 路径的 fallback 测试。

LGTM。

eanzhao (commented):
修复复核后，运行路径的 PUT/body 修正是对的；还有一个文档/注释残留需要清理，见 inline。

agents/Aevatar.GAgents.ChannelRuntime/SkillRunnerStreamingReplySink.cs
```diff
@@ -0,0 +1,590 @@
+using System.Text.Json;
+using Aevatar.AI.ToolProviders.NyxId;
+using Microsoft.Extensions.Logging;
+
+namespace Aevatar.GAgents.ChannelRuntime;
+
+/// <summary>
+/// Drives progressive (edit-in-place) rendering of a SkillRunner output by sending the first
+/// non-empty delta as a fresh Lark <c>POST /open-apis/im/v1/messages</c> and then editing the
+/// returned <c>message_id</c> via <c>PATCH /open-apis/im/v1/messages/{id}</c> for every later
```
eanzhao: 这里和下面 failure semantics 仍然写的是 `PATCH`，`SkillRunnerGAgent.TryCreateStreamingSink` 的 XML 注释以及 `docs/canon/daily-command-pipeline.md` 里也还有几处 “后续 PATCH / POST + PATCH” 的残留。代码已经修成 `PUT` + `msg_type:text`，但这些注释正好描述的是刚修掉的协议点，后续维护者很容易按注释把它改回 card-edit 语义。建议把所有 streaming text edit 描述统一改成 `PUT`，只在解释 Lark verb split 时保留 “PATCH is for interactive card”。

eanzhao:
Pushed `f32d9576` — final PATCH scrub. The follow-up review comment was right; I had three comments still claiming the verb was PATCH:

- `SkillRunnerGAgent.HandleTriggerAsync` post-execute comment ("POST initial + PATCH each delta")
- `SkillRunnerGAgent.TryCreateStreamingSink` xmldoc ("edits the same message via `PATCH /open-apis/im/v1/messages/{id}`")
- `SkillRunnerDefaults.StreamingEditThrottle` xmldoc ("Throttle for streaming-edit (Lark `PATCH ...`)")

All corrected to PUT, each with a one-liner reminding that PATCH is reserved for cards so a future maintainer doesn't "fix" it back.

The remaining `PATCH` matches in this PR are all in deliberate explainer comments that explicitly contrast PUT (used) with PATCH (reserved for interactive cards):
- `SkillRunnerStreamingReplySink.EditAsync` (the verb-split rationale)
- `SkillRunnerGAgent.HandleTriggerAsync` and `TryCreateStreamingSink` (now corrected with the explainer)
- `SkillRunnerDefaults.StreamingEditThrottle` (now corrected with the explainer)
- `SkillRunnerStreamingReplySinkTests` line 50 (test pinning the verb assertion)
- `daily-command-pipeline.md` §8 (the documentation of the verb-split invariant)

Those stay so future readers can grep `PATCH` and land on the verb-split documentation rather than an out-of-date implementation claim.

Also resolved the dev merge conflict from the directory rename (`Aevatar.GAgents.ChannelRuntime/` → `Aevatar.GAgents.Scheduled/` + `Aevatar.GAgents.Channel.Runtime/` + `Aevatar.GAgents.Authoring.Lark/` + …). The sink moved to `Aevatar.GAgents.Scheduled` alongside `SkillRunnerGAgent`/`SkillRunnerDefaults`; namespace and Lark-types `using` updated accordingly. PR is now MERGEABLE.

Tests: 591 passing post-merge. Docs lint: 32 files, 0 errors.

PR #471, opened 2026-04-28
fix(skill-runner): defense-in-depth safety net for nyxid_proxy tool failures (#439)

eanzhao:
## Summary

- **Layer 2 (tool)** — `NyxIdProxyTool` annotates every JSON-object response with `_aevatar_tool_status` (`error` / `ok`) so downstream consumers can distinguish a 2xx-with-data response from a NyxID-wrapped 4xx/5xx envelope, an approval-blocked call, or a Lark business-error envelope (`code != 0`).
- **Layer 3 (runner)** — Instance-owned `IToolCallMiddleware` on `SkillRunnerGAgent` reads the marker and increments a per-run counter; `ExecuteSkillAsync` calls `EnsureToolStatusAllowsCompletion` after the LLM stream completes. When every `nyxid_proxy` call this run failed, it throws — routing through the existing `HandleTriggerAsync` catch path so retries are honored and (after the budget is exhausted) `SkillRunnerExecutionFailedEvent` is persisted instead of a clean `Completed` event. Mixed runs still complete normally.

## Why

Issue #439 reported the daily report falsely surfacing `No meaningful public GitHub activity` for a user who actually had 52 commits + 114 issues/PRs in the same window. PR #458 fixed the prompt layer (§9 Source health footer), but the issue itself called out that pure prompt fixes are fragile: *"LLMs can and do drop format requirements under load."* This adds the two remaining layers from the issue's defense-in-depth design so a weaker model that ignores §9 still can't smuggle a fake-success past the runner.

## Affected code

- `src/Aevatar.AI.ToolProviders.NyxId/Tools/NyxIdProxyTool.cs` — adds `AnnotateWithToolStatus`, funnels every return path through it
- `agents/Aevatar.GAgents.Scheduled/SkillRunnerGAgent.cs` — wires the counting middleware via private chained ctor, resets counter per run, calls `EnsureToolStatusAllowsCompletion`
- `agents/Aevatar.GAgents.Scheduled/SkillRunnerToolFailureCounter.cs` — new
- `agents/Aevatar.GAgents.Scheduled/NyxIdProxyToolFailureCountingMiddleware.cs` — new

## Issue acceptance criteria (#439)

- [x] When `nyxid_proxy` returns 4xx/5xx/7xxx, the daily report output explicitly names the failing endpoint(s) — addressed by PR #458 prompt layer + this PR's structural marker
- [x] When all GitHub tool calls fail, `SkillRunnerGAgent` persists `SkillRunnerExecutionFailedEvent` (not `Completed`) — runner-layer safety net
- [x] When some calls succeed and some fail, the report includes partial data and lists the failed queries — prompt §9 plus mixed-policy in `EnsureToolStatusAllowsCompletion`
- [x] Test fixtures cover (a) all-fail; (b) mixed; (c) genuinely-empty — see `SkillRunnerToolFailureSafetyNetTests`

## Test plan

- [x] `dotnet build test/Aevatar.GAgents.ChannelRuntime.Tests/Aevatar.GAgents.ChannelRuntime.Tests.csproj` — clean, 0 errors, 24 warnings (pre-existing)
- [x] `dotnet test test/Aevatar.GAgents.ChannelRuntime.Tests/...` — 582 passed, 0 failed (24 new + 558 pre-existing)
- [x] `bash tools/ci/architecture_guards.sh` — relevant guards passed (playground asset drift requires pnpm vite build, unrelated to backend changes)


eanzhao (commented):
Overall: solid defense-in-depth design. The two-layer approach (tool annotation + runner safety net) is well-scoped and the test coverage is thorough.

A few issues to address:

---

**1. Constructor chain calls static method before `this()` — C# evaluation order**

`SkillRunnerGAgent.cs` public ctor delegates to private ctor via `this(BuildToolMiddlewareChain(toolMiddlewares), ...)`. In C#, arguments to `this(...)` / `base(...)` are evaluated before the constructor body, but `BuildToolMiddlewareChain` is static and allocates a new `SkillRunnerToolFailureCounter` + `NyxIdProxyToolFailureCountingMiddleware` each time. This works today but is fragile: if anyone changes `BuildToolMiddlewareChain` to reference instance members (e.g., logger), it silently breaks.

This is fine as-is, just flagging it for awareness.

---

**2. `SkillRunnerToolFailureCounter` uses `Interlocked` + `Volatile` — justified but worth a comment**

AGENTS.md says `禁止在模块内使用 lock/Monitor/ConcurrentDictionary 作为并发补丁来维护事实状态`. The counter uses `Interlocked.Increment` and `Volatile.Read/Write`. This is **correct and necessary** because `StreamingToolExecutor` dispatches read-only tools concurrently on the thread pool, so multiple middleware invocations can race.

However, `Reset()` writes two fields non-atomically (`Volatile.Write(ref _failureCount, 0); Volatile.Write(ref _successCount, 0);`). If a concurrent tool call's middleware runs between these two writes, you get a transient inconsistent state (failure=0, success=N from old run). This is benign in practice because `ExecuteSkillAsync` awaits the full `ChatStreamAsync` before returning, so no concurrent middleware should be in-flight when `Reset()` is called. But adding a brief comment at `Reset()` explaining this invariant would help future readers.

---

**3. `EnsureToolStatusAllowsCompletion` is `internal static` — consider the no-tool-calls edge case more carefully**

`Policy_NoToolCallsAtAll_Allows` passes when both counts are zero. This means a skill that *should* have called `nyxid_proxy` but the LLM decided to skip all tool calls entirely (e.g., hallucinated a report from memory) will pass the safety net. The issue description says "pure LLM transformation" is a valid use case, but for the daily-report skill specifically, if zero tool calls were made and the LLM outputs a report, that's exactly the failure mode #439 describes.

Consider whether the safety net should also have a mode that fails when the *expected* tool was never called (not just when it was called and failed). This may be out of scope for this PR, but worth noting.

---

**4. `_aevatar_tool_status` field leaks into LLM context**

The marker is injected into the JSON response that the LLM sees. While the PR description mentions this is intentional ("LLM prompt language"), the underscore-prefixed field name could confuse weaker models or get accidentally included in the LLM's output text. The LLM now sees something like:

```json
{"_aevatar_tool_status": "ok", "total_count": 52, "items": [...]}
```

This is fine for strong models but weaker ones may echo it. Since the prompt layer fix (#458) already addresses the problem at the LLM level, this extra LLM-visible annotation is redundant for the LLM and only needed by the middleware. Consider whether the middleware should strip the marker after reading it, or alternatively, store it in a side-channel (e.g., `context.Items`) rather than mutating the LLM-visible response.

---

**5. Test file location**

Tests are in `test/Aevatar.GAgents.ChannelRuntime.Tests/` but the production code is in `agents/Aevatar.GAgents.Scheduled/`. Is this the intended test project for `Scheduled` agents? The test project name suggests it's for ChannelRuntime-specific tests. Not a blocker, just checking.

---

**6. Minor: `AnnotateWithToolStatus` allocates `MemoryStream` + `Utf8JsonWriter` on every tool call**

This runs on every single `nyxid_proxy` response, including the happy path. The allocation is small but adds up in high-throughput scenarios. Not a blocker, but consider pooling or reusing the writer if this becomes a hot path.

agents/Aevatar.GAgents.Scheduled/SkillRunnerToolFailureCounter.cs
```diff
@@ -0,0 +1,29 @@
...
+    private int _failureCount;
+    private int _successCount;
+
+    public int FailureCount => Volatile.Read(ref _failureCount);
+
+    public int SuccessCount => Volatile.Read(ref _successCount);
+
+    public void Reset()
+    {
+        Volatile.Write(ref _failureCount, 0);
```
eanzhao, line 30: Reset() 写两个字段不是原子的。虽然实际安全（ExecuteSkillAsync 在 ChatStreamAsync 完成后才调用 Reset，不会有并发 middleware），但这个不变量依赖调用顺序，建议加注释说明。

agents/Aevatar.GAgents.Scheduled/NyxIdProxyToolFailureCountingMiddleware.cs
```diff
@@ -0,0 +1,70 @@
...
+
+        // Only the proxy tool's results carry the structural marker; other tools opt out.
+        if (!string.Equals(context.ToolName, "nyxid_proxy", StringComparison.Ordinal))
+            return;
+
+        var result = context.Result;
+        if (string.IsNullOrEmpty(result))
+            return;
+
+        switch (TryReadToolStatus(result))
```
eanzhao: 建议在 middleware 读完 marker 后从 Result 中移除 _aevatar_tool_status，或改用 context.Items 传递状态。当前 marker 会泄漏到 LLM 上下文中，弱模型可能把它写进输出文本。

agents/Aevatar.GAgents.Scheduled/SkillRunnerGAgent.cs
```diff
@@ -250,7 +302,32 @@ private async Task<string> ExecuteSkillAsync(DateTimeOffset now, string? reason,
...
+        var normalized = string.IsNullOrWhiteSpace(output) ? "No update generated." : output;
+        EnsureToolStatusAllowsCompletion(_toolFailureCounter.FailureCount, _toolFailureCounter.SuccessCount);
+        return normalized;
+    }
+
+    /// <summary>
+    /// Runner-layer safety net for issue #439: when every nyxid_proxy call in a run failed,
+    /// the LLM's plain-text output is structurally indistinguishable from a real "no
+    /// activity" report — the prompt-layer §9 Source health footer can be silently dropped
+    /// by a weaker model, and the runner has no other way to tell. Throwing here routes
```
eanzhao, line 407: 两个计数器都为 0 意味着 LLM 可能根本没有调用 nyxid_proxy（比如直接从记忆编造报告）。对 daily-report 场景，这正是 #439 要防止的 failure mode。考虑是否需要区分「没有 tool call」和「tool call 全部成功」两种情况。

eanzhao (commented):
## Review feedback

### 1. `ClassifyToolStatus` `code != 0` 覆盖面过宽，存在假阳性风险

`src/Aevatar.AI.ToolProviders.NyxId/Tools/NyxIdProxyTool.cs` 中 `ClassifyToolStatus` 将任意 `code != 0` 的 JSON 响应均判定为 `ToolStatusError`。当前这覆盖了 NyxID approval (7000/7001) 和 Lark business error 两种实际错误形态，但如果下游 API 在正常响应中使用了 `code` 字段（例如 `{"code": 200, "data": {...}}` 或返回 HTTP 风格的状态码），会被错误地记为 failure。

虽然 safety net 仅在 all-fail 时触发，但 `AnnotateWithToolStatus` 是 `internal static` 方法，标注着可复用。建议：

- 考虑将 `code != 0` 分支收紧为需要同时存在 `msg` 字段（Lark 模式），或仅匹配已知错误码区间（7000/7001 + Lark 业务错误）
- 或者至少在 XML doc 中明确标注这是 best-effort, 使用时需考虑下游 API 的响应形状

### 2. 端到端接线测试较弱

`test/Aevatar.GAgents.ChannelRuntime.Tests/SkillRunnerToolFailureSafetyNetTests.cs` 中 `Runner_ExposesCounter_ThatIsTheSameInstanceTheMiddlewareWritesTo` 测试名称暗示验证了 middleware ↔ agent 的连线正确性，但实际上只验证了 `ToolFailureCounterForTesting` 属性返回同一引用（同语反复）。真正的连线测试应该是：通过构造函数创建 agent → 获取 counter → 构造 middleware 拿到同一个 counter → 执行 middleware → 验证 agent 侧 counter 也记录了相同的值。当前的测试覆盖不了 `BuildToolMiddlewareChain` 中拼接链路的回归。

### 3. `Utf8JsonWriter` 整文档重序列化（性能）

`AnnotateWithToolStatus` 对每个 JSON-object 做 `JsonDocument.Parse` + `Utf8JsonWriter` 全量重写。对于 daily-report 场景（5-10 个小响应）影响很小，但如果后续将此模式扩展到高频 proxy 调用的场景，建议用字符串拼接（在 `{` 后直接插入 marker key-value + 逗号）避免 O(N) 的全文档重序列化开销。

### 总结

整体设计方向正确，三层防御逻辑自洽。中间件只计 `nyxid_proxy` 调用 + all-fail 才 throw + 利用现有 retry/catch 路径，不创造新的异常路径，这些都是好的设计决策。建议处理上述第 1 点（`code != 0` 跨度问题）后再合入，其余为优化建议。

src/Aevatar.AI.ToolProviders.NyxId/Tools/NyxIdProxyTool.cs
```diff
@@ -117,9 +117,100 @@ public async Task<string> ExecuteAsync(string argumentsJson, CancellationToken c
...
+    /// <summary>
+    /// Inject <see cref="ToolStatusFieldName"/> at the top of a JSON-object response. Returns
+    /// the input unchanged when the body isn't a JSON object (raw text, arrays, etc.) — the
+    /// marker is opportunistic, never destructive. Detection rules:
+    /// - <c>"error"</c> property truthy → error (matches <see cref="NyxIdApiClient.SendAsync"/>'s
+    ///   non-2xx wrapper and exception wrapper)
+    /// - <c>"code"</c> numeric and non-zero → error (matches NyxID approval codes 7000/7001 and
+    ///   any business-error envelope using the same shape)
+    /// </summary>
+    internal static string AnnotateWithToolStatus(string? response)
```
eanzhao: 这个 annotator 没覆盖 `slug` 为空的 discovery 分支：`ExecuteAsync` 在 discovery mode 直接 `return await DiscoverMergedServicesAsync(...)`。成功 discovery 返回数组可以不计入数据 fetch，但如果 `DiscoverProxyServicesAsync` 返回 `NyxIdApiClient` 的非 2xx wrapper（`{"error": true, ...}`），middleware 看不到 `_aevatar_tool_status` 会忽略，计数保持 `0/0`，`EnsureToolStatusAllowsCompletion` 仍会放行。建议 discovery 返回值也走 annotation；双 token merge 里也要先识别 error object，避免两个 discovery error 被合成 `[]`。

src/Aevatar.AI.ToolProviders.NyxId/Tools/NyxIdProxyTool.cs
```diff
@@ -117,9 +117,100 @@ public async Task<string> ExecuteAsync(string argumentsJson, CancellationToken c
...
+            return response;
+        }
+    }
+
+    private static string ClassifyToolStatus(System.Text.Json.JsonElement root)
+    {
+        if (root.TryGetProperty("error", out var errorProp) && IsTruthyError(errorProp))
+            return ToolStatusError;
+
+        if (root.TryGetProperty("code", out var codeProp)
```
eanzhao: 这个判断对通用 proxy 太宽了：`nyxid_proxy` 不只代理 Lark，任何成功业务响应只要顶层有 numeric `code` 且非 0（例如 `{ "code": 42, ... }` 这种领域字段）都会被标成 `error`。`SkillRunnerGAgent` 是通用 scheduled skill runner，这会让本来成功的单次 proxy 调用被计为 all-fail 并触发失败。建议把 `code != 0` 收窄到明确的错误 envelope（例如 Lark/Feishu 形态同时有 `msg`，或由 NyxID 的服务配置/HTTP 状态 wrapper 提供错误语义），不要对所有顶层 `code` 字段泛化判错。

eanzhao (commented):
Reviewed the PR and left two inline issues.

eanzhao (commented):
Fix commit 291a730 looks good. All three review items addressed:

1. **Marker leak fixed** - AnnotateWithToolStatus removed entirely from NyxIdProxyTool. Classification now lives in the middleware (ClassifyResult), reads raw response without mutation. LLM never sees the marker. NyxIdProxyToolStatusAnnotationTests.cs deleted.

2. **Reset() comment added** - SkillRunnerToolFailureCounter.Reset() now has thorough doc comment explaining the non-atomic invariant.

3. **No-tool-calls edge case acknowledged** - Policy_NoToolCallsAtAll_Allows test has explicit comment noting out-of-scope for this PR. Fair.

Additional improvements: LooksLikeCodeBasedErrorEnvelope now requires paired message field (msg/message/error_msg) to avoid false positives on domain code fields. Wiring test upgraded from tautological to driving actual registered middleware via reflection. Wiring_PreservesCallerInjectedMiddleware ensures DI middleware survives. Discovery arrays now classify as Ok in mixed runs.

No blocking issues remaining. Ready to merge.

agents/Aevatar.GAgents.Scheduled/NyxIdProxyToolFailureCountingMiddleware.cs
```diff
@@ -0,0 +1,133 @@
...
+    internal static ResultClassification ClassifyResult(string? result)
+    {
+        if (string.IsNullOrEmpty(result))
+            return ResultClassification.Unknown;
+
+        try
+        {
+            using var doc = JsonDocument.Parse(result);
+
+            if (doc.RootElement.ValueKind != JsonValueKind.Object)
```
eanzhao, line 84: 这里把所有非 object JSON 都计为 `Ok`，修了真实 array 成功的场景，但还会漏掉双 token discovery 的失败路径。`NyxIdProxyTool.DiscoverMergedServicesAsync` 在 user/org 两次 `DiscoverProxyServicesAsync` 都返回 `NyxIdApiClient` error object 时，会成功 parse 两个 object、`ParseServiceSlugs` 得到空集合，然后 serialize 成 `[]` 返回；这里再把 `[]` 计为 success，最终 `0 failure / 1 success`，runner 会完成而不是失败。建议在 discovery merge 里先识别 `{"error": ...}` 并原样返回/抛出，或者让 classifier 能区分“真实 discovery array”和“由 error objects 合成的空 array”。

agents/Aevatar.GAgents.Scheduled/NyxIdProxyToolFailureCountingMiddleware.cs
```diff
@@ -0,0 +1,133 @@
...
+            || !codeProp.TryGetInt64(out var code)
+            || code == 0)
+        {
+            return false;
+        }
+
+        // Require a paired message field so we don't false-flag domain responses that
+        // happen to use `code` for a non-error meaning. The Lark/Feishu envelope is
+        // {code, msg, ...}; NyxID approval is {code, message, approval_request_id, ...}.
+        return root.TryGetProperty("msg", out _)
```
eanzhao: 这个收窄仍然会误伤通用 proxy 的正常成功响应。很多 API 会返回 `{ "code": 200, "message": "success", ... }` 或 `{ "code": 1, "message": "ok" }` 这类业务成功 envelope；当前只要 `code != 0` 且有 `message` 就计为 failure，单次成功 proxy 调用会被 all-fail safety net 打成失败。建议只把 `msg` 作为 Lark/Feishu envelope，`message` 分支则限制到 NyxID/approval 的已知错误码（例如 7000/7001，或实际 NyxID `error_code` wrapper），不要把 generic `code + message` 当错误。

eanzhao (commented):
Reviewed the latest fix; two edge cases still look actionable.

Issue #473, opened 2026-04-28
[Bug] GitHub Daily Report 返回大量 422 invalid user/permission 且提交数据不准确


YueZh127:
###  Description

在使用 `/daily` 生成 GitHub 日报时，即使输入有效用户名（如 `Yuezh0127`），系统仍在所有数据源返回：

`422 invalid user/permission`

同时 commit（提交）数据出现异常：

- 返回“unrelated global results”（全局无关结果）
- 标记为“not attributable to Yuezh0127”（无法归属于该用户）

导致日报无法生成有效内容。

---

###  Steps to Reproduce
1. 执行命令：/daily username 
2. 查看返回结果

3. 出现以下问题：

- 所有数据源均报错：
- shipped PRs
- in-flight PRs
- reviews
- issues opened / closed / commented
- commit 数据异常：
- 返回全局无关结果
- 无法归属于目标用户

---

###  Actual Behavior

- 所有查询返回 `422 invalid user/permission`
- commit 查询退化为全局搜索
- 返回数据无法使用
- 错误信息无法区分具体原因

---

### Expected Behavior

- 对有效 GitHub 用户：
- 应返回对应用户的 PR / issues / reviews / commits 数据
- commit 查询应：
- 严格基于 `author:username` 过滤
- 错误应明确区分：
- 用户不存在
- 权限不足
- API 限流
- 查询参数错误
- ❌ 不应 fallback 到无关的全局数据

---

eanzhao:
Triaged. This is a real bug, not a duplicate of any existing ticket — but it's adjacent to several. Sharing the search results so context is consolidated.

**No existing issue tracks the 422 root cause.** Closest tickets:

- #411 (closed) and #417 (closed) — both fixed the **403 ApiKeyScopeForbidden** path (NyxID api-key allowlist using catalog `DownstreamService.id` instead of per-user `UserService.id`). With #417 deployed, `/rate_limit` returns 200, but `/search/*` can still 422 for orthogonal reasons (OAuth scope sufficiency for GitHub's search engine). The preflight from #411 is too narrow to catch this.
- #439 (open) — the runner currently masks GitHub tool failures as silent "no activity" successes. Note that in **your** report the §9 Source health footer IS surfacing the failures correctly (you can see the `422 invalid user/permission` text), so the prompt-side classifier is working. The remaining problem is preventing those 422s in the first place, plus the secondary symptom you flagged: commits falling back to "unrelated global results."
[...]

PR #479, opened 2026-04-28
fix(daily): preflight /search/* with bound username to catch GitHub 422 at create-time (#474)

### LINKED ISSUE #474: bug(daily): GitHub /search/* returns 422 at runtime — preflight (rate_limit) only screens 401/403, search-API permission/scope failures slip through
> Tracks the same operational bug as #473 (user-facing report). This issue captures the engineering analysis and fix direction; #473 is the QA-side reproduction.

## Symptom

After a `/daily Yuezh0127` run, every GitHub data source in the report comes back as `422 invalid user/permission`:

- shipped PRs (`/search/issues?q=author:...+is:pr+is:merged`)
- in-flight PRs (`/search/issues?q=author:...+is:pr+is:open`)
- reviews (`/search/issues?q=reviewed-by:...`)
- issues opened / closed / commented (`/search/issues?q=author:...+is:issue+...` and `commenter:...`)

Commit data (`/search/commits?q=author:...+author-date:>=...`) misbehaves differently — it returns content the LLM flags as **\"unrelated global results, not attributable to Yuezh0127\"**. Net user impact: the daily report is unusable.

The §9 Source health footer (added under #439's prompt-side hardening) IS surfacing the failures with a `422 invalid user/permission` reason — so this is **not** a regression of the silent-fake-success bug. The classification layer works; the underlying 422s are real and unexpected.

This is also distinct from #411 / #417, which were **403 ApiKeyScopeForbidden** root causes (NyxID-side allowlist using catalog `DownstreamService.id` instead of per-user `UserService.id`). After #417 landed, `allowed_service_ids` carries the right value and `/rate_limit` returns 200 — but `/search/*` can still 422 for reasons orthogonal to api-key scope.

## Root cause (working hypothesis)

The fix in #411/#417 ended at the api-key allowlist boundary. The preflight gate in `AgentBuilderTool.PreflightGitHubProxyAsync` only probes `/rate_limit` and only fail-fasts on **401/403**:

```csharp
// agents/Aevatar.GAgents.Authoring.Lark/AgentBuilderTool.cs:1663-1700
var probe = await nyxClient.ProxyRequestAsync(apiKey, \"api-github\", \"/rate_limit\", \"GET\", ...);
...
if (status != (int)HttpStatusCode.Unauthorized && status != (int)HttpStatusCode.Forbidden)
    return null;  // ← anything else, including 422, is treated as healthy
```

`/rate_limit` is a privileged-token rate metadata endpoint — it succeeds with **any** valid token, including tokens whose OAuth grant lacks the scopes GitHub's search engine actually checks (`public_repo` for public commit/issue search, `repo` for private). So the preflight rubber-stamps an api-key that will fail every single `/search/*` call at runtime.

GitHub's documented 422 from `/search/commits` and `/search/issues` looks like:

```json
{
  \"message\": \"Validation Failed\",
  \"errors\": [{
    \"message\": \"The listed users, organizations or repositories cannot be searched either because the resources do not exist or you do not have permission to view them.\",
    \"resource\": \"Search\",
    \"field\": \"q\",
    \"code\": \"invalid\"
  }]
}
```

The LLM faithfully compresses this to `422 invalid user/permission` for the §9 footer. So the **string the user sees is correct** — what's wrong is that this state is reachable in production at all.

Most likely chain (pending verification against a real failing token):

1. NyxID GitHub OAuth grant is established with a minimal scope (e.g. `read:user`) sufficient for `BuildGitHubAuthorizationResponseAsync` to report \"connected\" and for `/rate_limit` / `/users/{username}` probes to return 200.
2. The minted agent api-key inherits this grant via `UserService.id` allowlist (post-#417).
3. At runtime, `/search/issues` and `/search/commits` are routed through the same proxy + token, but GitHub's search engine enforces stricter scope requirements than the basic user endpoints. With insufficient scope it returns **422 \"users... cannot be searched\"** rather than 401/403.
4. The proxy passes 422 through verbatim (`NyxIdProxyTool.cs:111-120` returns `result` as-is — same default-to-pass-through pattern called out in #439, but for a different status code). The LLM has no way to convert this into a fix-fast signal at agent-create time because no preflight ever exercised the search endpoints.

Secondary observation — the commit \"unrelated global results\" symptom:

For most search endpoints, a 422 returns no items, so the LLM has nothing to display and emits the source-health line. For `/search/commits` specifically, the same `422 invalid user/permission` body comes back, but I suspect the LLM is **either** falling back to a relaxed query (the prompt at `AgentBuilderTemplates.cs:181` says \"Do not retry\" but LLMs do drop format rules), **or** the URL is being constructed without proper `author:` filter qualification, giving GitHub a relevance-only query that returns unfiltered global commits. Worth verifying which it is in production logs before fixing.

## Why the existing layers don't catch this

| Layer | Coverage | Gap for #473 |
| --- | --- | --- |
| `BuildGitHubAuthorizationResponseAsync` | OAuth grant present at NyxID | Doesn't verify scope sufficiency for search APIs |
| `PreflightGitHubProxyAsync` (#411 + #417) | api-key can hit `/rate_limit` (401/403) | `/rate_limit` is scope-light; doesn't exercise `/search/*`; only screens 401/403 |
| `NyxIdProxyTool.ExecuteAsync` | Approval-error detection (codes 7000/7001) | All other non-2xx pass through as opaque JSON to the LLM |
| Daily prompt §9 Source health | Surfaces 4xx/5xx into output | Surfaces *correctly*, but the runtime call still cost a turn and the report is empty |
| SkillRunner completion classification (#439, in flight) | Will downgrade tool-error outputs to `Failed` | Doesn't prevent the failed run from being scheduled in the first place |

The right place to plug this is **agent-create time**, the same gate that #411 added. Catching it at runtime via #439's classifier means we log a `Failed` run after wasting a tool call and a Lark notification — useful, but the agent should never have been persisted.

## Suggested fix direction (not prescriptive)

The minimum viable shape is to widen the preflight to actually exercise the endpoints the daily report depends on, and treat 422 as fail-fast for those endpoints:

- **Preflight (P0)**: in `PreflightGitHubProxyAsync`, after `/rate_limit` returns 200, run a no-op `/search/issues?q=author:{username}+type:pr&per_page=1` (and a `/search/commits?q=author:{username}&per_page=1` with the appropriate `Accept` header). Treat **422 from these endpoints** as `github_search_unauthorized` and surface a structured error pointing to OAuth scope rather than NyxID binding. Revoke the freshly minted api-key on failure (same pattern as #411).
- **Error mapping (P1)**: distinguish at least four 422 reasons in the user-facing message — (a) user does not exist, (b) scope insufficient, (c) query malformed, (d) GitHub search rate limit (different envelope from REST rate limit). Today they all collapse to one string; per #473's expected behavior, they need to be separable.
- **Tool layer (P2, overlap with #439)**: shape the proxy response to flag tool-error envelopes structurally so the LLM can't mis-classify a 422 as \"the user just had no PRs.\" (#439 already proposes this for the runner side; same hook helps here.)
- **Commit fallback (P2)**: investigate the \"unrelated global results\" symptom from production logs — confirm whether the LLM is retrying without the author filter (prompt-discipline failure → tighten the prompt + add a runner-side regex check for `author:` in the URL) or whether the URL is mis-substituted upstream. Decide based on the log evidence; don't pre-commit to a fix until that's known.

Pure preflight hardening alone isn't enough either — once an OAuth grant is downgraded out-of-band (user goes to github.com/settings/applications and shrinks the scope), an existing scheduled agent will still 422 at runtime. So preflight catches the create-time case; the runtime classifier (#439) catches the in-flight degradation case. They're complementary.

## Acceptance criteria

- [ ] `/daily` create-time preflight calls `/search/issues` and `/search/commits` with the bound `github_username` and a `per_page=1` window, in addition to `/rate_limit`. A 422 from either is fail-fast.
- [ ] User-facing error on 422 distinguishes \"user does not exist\" / \"scope insufficient\" / \"query malformed\" / \"search rate limited\" — at minimum the first two, since they are the paths a user can act on.
- [ ] When create-time preflight fails on 422, the freshly minted api-key is best-effort revoked (same as #411's 401/403 path).
- [ ] At runtime, when every search-call returns 422, `SkillRunnerExecutionFailedEvent` is persisted (overlaps with #439's coverage; this issue does not duplicate that fix but should not regress it).
- [ ] Test fixture: a fake NyxID proxy returning 422 for `/search/*` and 200 for `/rate_limit` — current code persists a happy agent, fixed code rejects creation.
- [ ] Test fixture: production-shape 422 body (`{\"message\":\"Validation Failed\",\"errors\":[...]}`) round-trips into the structured error envelope, not raw passthrough.

## Affected paths

- `agents/Aevatar.GAgents.Authoring.Lark/AgentBuilderTool.cs:1655-1725` — `PreflightGitHubProxyAsync` (probe + 401/403 classifier; needs to grow `/search/*` coverage and a 422 case)
- `agents/Aevatar.GAgents.Authoring.Lark/AgentBuilderTemplates.cs:115-184` — daily prompt + §9 footer rules (the surfacing layer is fine; the prompt's anti-fallback wording should grow a per-section guard against retrying a failed `/search/commits` without `author:`)
- `src/Aevatar.AI.ToolProviders.NyxId/Tools/NyxIdProxyTool.cs:71-121` — proxy passthrough (overlap with #439's structured-error proposal)
- `src/Aevatar.AI.ToolProviders.NyxId/NyxIdApiClient.cs` — `ProxyRequestAsync` shape; verify URL encoding of `+` in `q=` query strings against GitHub's parser to rule out the commit-fallback symptom being a transport-level encoding bug

## Related

- #473 — user-facing duplicate of this bug (this issue is the engineering analysis)
- #439 — sibling: runtime classifier that downgrades fake-success runs to `Failed`. Complementary (covers in-flight degradation; this issue covers create-time prevention).
- #411 — original preflight that landed `/rate_limit`-only screening for 401/403. This issue is the natural follow-up: extend the same preflight to the search surface that actually matters for the daily report.
- #417 — fixed the per-user `UserService.id` allowlist bug; ruled out the 403 path. With #417 deployed, 422 is the next-most-common failure shape.
- #436 / #437 — separate per-Lark-user binding bug, unrelated root cause.

eanzhao:
## Summary

- `PreflightGitHubProxyAsync` now probes `/search/issues` and `/search/commits` with the bound `github_username` after the existing `/rate_limit` step. 422 responses (the "users... cannot be searched..." surface from #473) trigger fail-fast as `github_search_unauthorized`, distinct from the existing 401/403 `github_proxy_access_denied` envelope, with a stable `reason_code` for operators (`scope_insufficient_or_user_not_found` vs. `validation_failed`).
- The freshly minted api-key is best-effort revoked on the new 422 path, mirroring the cleanup contract from PR #418 so `/daily` retries don't accumulate orphan proxy-scoped keys.
- Three new unit tests pin the behavior: `/search/issues` 422 fail-fast (production-shape envelope), commits-only 422 (failure on the second probe, matching #473's reported degradation), and empty-200 success (so we don't regress into rejecting legitimate "no recent activity" creations).

Closes #474. Related: #473 (user-facing reproduction); complementary to #439 (runtime classifier — not duplicated here).

## Why this is needed

`/rate_limit` is scope-light. It returns 200 for any valid OAuth token, including tokens whose grant lacks `public_repo` / `repo`. That gap was invisible to the create-time preflight from #411 / #417, which only checked 401/403. So agents persisted with a healthy preflight, only to 422 every `/search/*` call at runtime — every scheduled daily report came back as `422 invalid user/permission` across all sections, and commits degraded to "unrelated global results, not attributable to {user}". Catching this at agent-create time is strictly better than catching it after the agent is scheduled, persisted, and has wasted a tool call + Lark notification on every cron tick.

## Test plan

- [x] `dotnet build agents/Aevatar.GAgents.Authoring.Lark/Aevatar.GAgents.Authoring.Lark.csproj` — clean
- [x] `dotnet test test/Aevatar.GAgents.ChannelRuntime.Tests/...` — 561 passed (including 3 new tests for the 422 paths)
- [x] `dotnet test test/Aevatar.AI.Tests/...` — 475 passed (covers `NyxIdApiClient` envelope shape)
- [x] `bash tools/ci/architecture_guards.sh` — passes through the proto/projection/channel/workflow guards; `playground_asset_drift_guard` exits with the same 254 on a clean `dev` checkout (environmental on this worktree, not caused by the change). Verified by stashing the diff and re-running.
- [ ] CI green
- [ ] Manual verification path (out of scope for this PR; tracked in #474 acceptance criteria): exercise `/daily` against a NyxID account whose GitHub OAuth grant has `read:user` only — expect `github_search_unauthorized` with `reason_code: scope_insufficient_or_user_not_found` instead of a persisted-but-broken agent.


eanzhao (commented):
LGTM. Thorough review — no blocking issues found.

**What this does**: Adds `/search/issues` and `/search/commits` probes (with the bound `github_username`) to the GitHub proxy preflight, catching the 422 failure mode where `/rate_limit` passes but the OAuth grant lacks `public_repo`/`repo` scope. This prevents persisting daily-report agents that would 422 every scheduled run.

**Code quality**:
- `IsErrorEnvelope` correctly extracts the shared `NyxIdApiClient.SendAsync` error envelope shape, deduplicating the parsing logic between `ClassifyRateLimitProbeFailure` and `ClassifySearchProbeFailure`.
- Backward compatible: when `githubUsername` is null/empty, search probes are skipped and only `/rate_limit` runs — same behavior as before.
- Error classification is conservative: `ClassifyGitHubSearch422Body` falls through to `validation_failed` for unknown 422 bodies rather than guessing.
- All three new tests pass and cover: 422 on `/search/issues`, 422 on `/search/commits` only (issues succeed), and empty-200 success (no regression on legitimate "no recent activity" users).

**Non-blocking nit**: The `detail` variable in `ClassifySearchProbeFailure` shadows the local `detail` output from `IsErrorEnvelope` — the right-hand ternary uses the local, the left-hand is the anonymous type property. This is correct C# but slightly confusing to read. Consider renaming the local (e.g., `envelopeDetail` or `nyxMessage`) for clarity in a follow-up.

eanzhao (commented):
Overall: the fix is well-structured — the two-step probe, the refactoring into separate classifiers, and the test coverage for the main 422 paths are all solid. A few gaps worth addressing before merge:

**1. Missing test: empty/null username skips search probes**
`normalizedUser` empty returns `null` immediately, but no test verifies the search probes are never issued when username is absent. The existing 403 test short-circuits at Step 1, so it does not cover this. A dedicated test would protect against a future regression where someone removes the empty guard.

**2. Missing test: `validation_failed` reason_code**
Only `scope_insufficient_or_user_not_found` is exercised. A 422 with a body that does not contain "cannot be searched" (e.g. a query syntax error) should produce `validation_failed`, but this default branch is untested.

**3. `{username}` in hint is a literal, not interpolated**
The `hint` string contains `https://github.com/{username}` as a regular string literal (not `$""`), so it outputs the literal text `{username}` rather than the actual username. Since `github_username` is already a field in the response, this is cosmetic, but it would be clearer to either interpolate the actual value or drop the template and reference the field.

**4. Sequential search probes (not blocking)**
Both probes go through the NyxID proxy to GitHub, each adding a full round-trip. For the create-time path this is acceptable, but if latency becomes a concern, `Task.WhenAll` would halve the wait on the happy path.

eanzhao:
Pushed [f77d83f7](https://github.com/aevatarAI/aevatar/pull/479/commits/f77d83f7) addressing review feedback:

**Codex P1 ([r3152148327](https://github.com/aevatarAI/aevatar/pull/479#discussion_r3152148327))** — repo-scoped permissions probe added. After the global `/search/*` step, when `repositories` is configured, preflight loops the allowlist and runs `/search/issues?q=repo:{owner}/{repo}+author:{username}` and `/search/commits?q=repo:{owner}/{repo}+author:{username}` for each. Fail-fasts with the failing repo embedded in `github_path`. Carried `Repositories` through `DailyReportTemplateSpec` so the preflight reads the same normalized list the prompt template emits — no duplicate comma-split.

**eanzhao review (`scope_insufficient_or_user_not_found` only)** — added `ExecuteAsync_CreateAgent_DailyReport_FailsClosed_When_GithubSearchReturns422_WithUnknown422Body` covering the conservative fall-through to `validation_failed` for 422 bodies without the "cannot be searched" markers.

**eanzhao review (no test for `/rate_limit` short-circuit)** — added an assertion to the existing 403 test pinning that no `/search/*` request fires when `/rate_limit` returns 401/403. Catches the empty-username regression scenario by structural symmetry — both early returns are in the same method, so a re-ordering that breaks one breaks both.

**eanzhao review (`{username}` literal in hint)** — fixed; the hint now references the `github_username` field already in the response envelope.

**eanzhao non-blocking nit (`detail` shadowing)** — renamed local to `envelopeMessage` in `ClassifyRateLimitProbeFailure` and `ClassifySearchProbeFailure` so the anonymous-type `detail` property isn't shadowed.

**Skipped** (`Task.WhenAll` parallelization, eanzhao non-blocking #4): kept sequential. Add latency at create-time is bounded; sequential is simpler to read and predictable to fail-fast on the first repo that doesn't pass.

Tests: 583 pass on `Aevatar.GAgents.ChannelRuntime.Tests` (up from 581 — 2 new tests for the unknown-422 fall-through and the repo-scoped 422 fail-fast paths).

eanzhao:
Review follow-up verified on commit `f77d83f` — all feedback addressed:

1. **`validation_failed` reason_code** — New test `FailsClosed_When_GithubSearchReturns422_WithUnknown422Body` covers the fall-through branch. ✅
2. **Rate-limit short-circuit guard** — 403 test now asserts `handler.Requests` contains no `/proxy/s/api-github/search/` calls, pinning that step 2+3 are never reached on step 1 failure. ✅
3. **`{username}` literal in hint** — Replaced with `"the bound github_username (see field above)"`. ✅
4. **`detail` shadow** — Renamed to `envelopeMessage` at call sites in both classifiers. ✅
5. **Codex P1: repo-scoped probes** — Step 3 loops `IReadOnlyList<string> repositories` with `repo:{owner}/{repo}+author:{user}` queries. `DailyReportTemplateSpec` gains `Repositories` field. New test `FailsClosed_When_RepoScopedSearchReturns422` pins the behavior. ✅

No remaining blocking issues. LGTM.

eanzhao:
## PR Review: feat(lark-bot): swap typing reaction to done after reply

### Scope Check: DRIFT DETECTED

**Stated intent** (from PR title/description): Swap Lark typing reaction from "Typing" → "DONE" after reply lands.

**Delivered**: The typing→DONE swap feature *plus* a full revert of the OwnerScope/caller-scoping infrastructure from issue #466, removal of `SkillRunnerStreamingReplySink`, `AgentBuilderActionIds`, `AgentBuilderJson`, `AgentBuilderCardContent.FormatListAgentsResult`, `ICallerScopeResolver` chain, `IUserAgentDeliveryTargetReader`, and substantial frontend deletions across studio pages.

**Out-of-scope changes:**
- Complete removal of `OwnerScope` proto message and all caller-scoping resolvers
- Deletion of `SkillRunnerStreamingReplySink` (591 lines) and its tests
- Removal of `AgentBuilderActionIds` constants (inlined into consumers)
- Removal of `AgentBuilderJson` shared helpers (duplicated into 3 consumers)
- Frontend studio script-draft storage cleanup
- `AgentDeliveryTargetTool` re-exposes `nyx_api_key` and `owner_nyx_user_id` as LLM tool parameters

---

### CRITICAL Findings

**[P1] (confidence: 8/10) `AgentDeliveryTargetTool` — `nyx_api_key` re-exposed to LLM**

The PR adds `nyx_api_key` back as an LLM-overridable tool parameter in `AgentDeliveryTargetTool.ParametersSchema`. The issue #466 review specifically removed this to close an impersonation surface: an LLM could be prompted to supply a different API key, letting it act as a different identity for outbound delivery. The PR description doesn't acknowledge this regression.

Location: `src/Aevatar.AI.ToolProviders.AgentCatalog/AgentDeliveryTargetTool.cs` (new `nyx_api_key` and `owner_nyx_user_id` params)

**[P1] (confidence: 9/10) `QueryAllAsync` in `UserAgentCatalogQueryPort` — hard cap at 1000, no pagination**

```csharp
var result = await _documentReader.QueryAsync(new ProjectionDocumentQuery { Take = 1000 }, ct);
```

This silently truncates at 1000 entries. The old `QueryByCallerAsync` had cursor-based pagination with a 5000 hard cap. `AgentBuilderTool.QueryAgentsForOwnerAsync` then filters client-side with `.Where(x => string.IsNullOrWhiteSpace(ownerFilter) || ...)`, so users with >1000 agents across all owners will see silently incomplete results. Even worse, `UserAgentCatalogRuntimeQueryPort.QueryAllAsync` loads *all* credentials into a dictionary (also capped at 1000 by `ProjectionDocumentQuery { Take = 1000 }`), and the credential set may not align 1:1 with the document set.

Location: `agents/Aevatar.GAgents.Scheduled/UserAgentCatalogQueryPort.cs:28`, `UserAgentCatalogRuntimeQueryPort.cs:37`

**[P2] (confidence: 8/10) `QueryAgentsForOwnerAsync` — client-side owner filter after fetching ALL agents**

```csharp
var entries = await queryPort.QueryAllAsync(ct);
return entries
    .Where(x => string.IsNullOrWhiteSpace(ownerFilter) || string.Equals(x.OwnerNyxUserId, ownerFilter, StringComparison.Ordinal))
    ...
```

The old implementation pushed the owner filter into the projection store via `ProjectionDocumentFilter`. Now every `list_agents` call fetches every non-tombstoned agent in the system and filters in-process. If `ownerFilter` is null/empty (fallback path), it returns ALL agents from all owners — a cross-tenant information disclosure.

Location: `agents/Aevatar.GAgents.Authoring.Lark/AgentBuilderTool.cs:807-812`

---

### INFORMATIONAL Findings

**[P2] (confidence: 7/10) `AgentBuilderJson` helpers duplicated into 3 files**

`TryReadError`, `TryReadString`, `TryReadBool`, `TryReadOptional` are now copy-pasted into `AgentBuilderCardContent`, `AgentBuilderCardFlow`, and `NyxRelayAgentBuilderFlow`. The deleted `AgentBuilderJson.cs` comment explicitly warned about this divergence hazard. Any future fix to JSON parsing logic must be replicated in 3 places.

**[P2] (confidence: 7/10) `AgentBuilderActionIds` constants inlined**

The action ID constants are now scattered as string literals across `AgentBuilderCardFlow` and `AgentBuilderCardContent`. A typo in a button argument would silently route to a fallback branch with no compile-time signal.

**[P3] (confidence: 6/10) `UserAgentCatalogCommandPort.Matches` now compares `NyxApiKey`**

```csharp
private static bool Matches(UserAgentCatalogEntry entry, UserAgentCatalogUpsertCommand command) =>
    ...
    && string.Equals(entry.NyxApiKey, command.NyxApiKey, StringComparison.Ordinal);
```

The `NyxApiKey` is populated from the runtime query port (which reads from the credential document). The old code explicitly excluded `NyxApiKey` from the match because it wasn't projected to the catalog document. If the credential read model lags behind the upsert, this match will fail and the command port will time out waiting for observation.

**[P3] (confidence: 5/10) Typing→DONE race on relay path**

The relay path does `_ = TrySwapTypingReactionToDoneAsync(...)` (fire-and-forget). If the Lark API is slow, the swap may complete after the user has already seen the message. Not a correctness bug, but worth noting — the direct-reply path correctly awaits-with-timeout via `AwaitTypingReactionThenSwapAsync`.

---

### Positive Notes

- The typing→DONE swap design is clean: list-based discovery avoids caching reaction_id on the singleton runner, `operator_type=app` filter correctly preserves user-added reactions.
- The `AwaitTypingReactionThenSwapAsync` race guard is well-thought-out with the 2-second timeout cap.
- `OnReplyDeliveredAsync` hook for the streaming completion path is the right abstraction.
- Test coverage is solid: new tests cover the swap sequence, streaming path hook, relay-path registration failure graceful degradation, and pagination.
- `ShouldSwapTypingReaction` correctly gates on `om_` prefix for Lark message IDs.

---

### Verdict

The typing→DONE feature is well-implemented. The OwnerScope revert is the main concern — it reintroduces the `nyx_api_key` as an LLM parameter, removes store-level owner filtering, and creates a cross-tenant disclosure path when `ownerFilter` is empty. These need to be addressed before merging.

eanzhao:
## PR #479 Review — LGTM with minor notes

核心修复正确：`/rate_limit` 是 scope-light 的，无法捕获 `/search/*` 422 的生产故障模式。新增的三层 probe（rate_limit → global search → per-repo search）覆盖了所有 failure surface。

### 确认的改进

- `IsErrorEnvelope` 提取为公共解析方法，rate-limit 和 search classifier 统一走同一个 envelope 解析，消除了之前重复的 JSON 解析逻辑
- `ClassifyGitHubSearch422Body` 的保守分类策略正确：匹配到 "cannot be searched" / "do not have permission" → `scope_insufficient_or_user_not_found`，其他 422 → `validation_failed`，不猜测
- best-effort api-key revoke 在所有 preflight 失败路径上都触发，跟 #418 的 cleanup 契约一致
- per-repo probe 用 `per_page=1` 控制 payload，`Uri.EscapeDataString` 只用在 username 段，repo 的 slash 保留原文 — 正确处理 GitHub 的 `owner/repo` 格式

### 测试覆盖

6 个测试 pin 了关键行为：
1. `/rate_limit` 403 fail-fast + **短路后续所有 `/search/*` 调用**（关键断言：`handler.Requests.Should().NotContain(r => r.Path.Contains("/search/"))`）
2. `/search/issues` 422 production envelope shape
3. `/search/commits`-only 422（issues 200 但 commits 422，exercises 第二个 probe）
4. 空 200 success（不误杀合法的 "no recent activity" 用户）
5. 未知 422 body → `validation_failed`（保守 fallback）
6. 全局 probe 200 但 repo-qualified probe 422（新 failure mode：token 有 `public_repo` 但 allowlist 含 private repo）

### 小 note（不阻断）

`PreflightGitHubProxyAsync` 签名从 4 参数涨到 6 参数（加了 `githubUsername` + `repositories`）。目前还在可接受范围，但如果后续再加 probe 维度，考虑封装一个 `PreflightRequest` record。现在不需要改。

**无阻塞项。** 可以合并。

eanzhao:
## Re-review after dev merge

The three-dot diff (what this PR actually adds beyond `dev`) is now **5 files, +845/-14 lines** — purely the typing→DONE reaction swap feature. The OwnerScope revert, `SkillRunnerStreamingReplySink` deletion, frontend cleanup, and other cleanup changes were merged into dev via separate PRs (#478, #483, #484, #490).

### Previous critical findings — status:

| Finding | Status |
|---|---|
| `nyx_api_key` re-exposed to LLM | Addressed in #478 (merged to dev) |
| `QueryAllAsync` hard cap at 1000 | Addressed in #478 (merged to dev) |
| Client-side owner filter cross-tenant disclosure | Addressed in #478 (merged to dev) |

### Remaining feature-only review:

**No new critical issues found.** The typing→DONE implementation is clean:

- `TrySwapTypingReactionToDoneAsync` correctly uses list-based discovery with `operator_type=app` filter
- `AwaitTypingReactionThenSwapAsync` race guard with 2s timeout cap is well-designed
- `OnReplyDeliveredAsync` hook correctly covers the streaming completion path
- `ShouldSwapTypingReaction` gates on `om_` prefix and Lark/Feishu platform
- Pagination with `MaxListPages=10` safety cap handles misbehaving Lark responses
- Test coverage is thorough: swap sequence, streaming hook, relay registration failure, pagination, non-Lark no-op, race guard, reply-failure skip

### Nit (informational, confidence 6/10):

**[P3] `OnReplyDeliveredAsync` fire-and-forget on streaming path**

In `ConversationGAgent.cs`, the streaming completion path calls `_ = ResolveRunner().OnReplyDeliveredAsync(...)` — the swap is fire-and-forget. If the Lark API is slow, the user may see the DONE reaction appear after they've already moved on. Not a correctness bug, just a UX observation. The non-streaming paths (`RunLlmReplyAsync`, `TryHandleAgentBuilderAsync`) have the same pattern for the relay path (`_ = TrySwapTypingReactionToDoneAsync`), so this is consistent.

### Verdict: LGTM — ready to merge.

PR #487, opened 2026-04-28
feat(lark-bot): swap typing reaction to done after reply

eanzhao:
## Summary
- Immediate inbound reaction changed from `OK` → `Typing` (Lark's keyboard-typing hands) so the user sees the bot working on it.
- After the reply lands, the bot lists its own `Typing` reactions (filter `operator_type=app` so user-added Typing is preserved), deletes each, and adds `DONE`.
- Uses list-based discovery instead of caching `reaction_id` on the singleton runner — keeping cross-turn reaction state on a service-layer dictionary would violate the "禁止中间层进程内缓存作为事实源" rule.
- Wired post-reply at `TryHandleAgentBuilderAsync` (direct slash/card replies) and `RunLlmReplyAsync` (the deferred LLM path, which is the main message → reply flow). `RunLlmReplyAsync` now resolves registration even on the relay path so `NyxProviderSlug` is available for the swap; the relay reply itself still doesn't need it.

## Test plan
- [x] `dotnet build agents/Aevatar.GAgents.NyxidChat/Aevatar.GAgents.NyxidChat.csproj` — clean
- [x] `dotnet test test/Aevatar.GAgents.ChannelRuntime.Tests/Aevatar.GAgents.ChannelRuntime.Tests.csproj` — 580/580 pass, including:
  - Updated `RunInboundAsync_ShouldSendImmediateLarkReaction_*` asserts `"emoji_type":"Typing"`
  - New `RunLlmReplyAsync_ShouldSwapTypingReactionToDone_AfterSuccessfulRelayReply` verifies GET → DELETE (bot reaction only) → POST DONE sequence and that user-owned Typing reactions are left alone
  - New `RunLlmReplyAsync_ShouldNotSwapReaction_WhenReplyFails` verifies no swap fires on reply failure
- [x] `bash tools/ci/test_stability_guards.sh` — clean (TCS-based deterministic wait, no polling)


agents/Aevatar.GAgents.NyxidChat/ChannelConversationTurnRunner.cs
```diff
@@ -124,29 +125,32 @@ public async Task<ConversationTurnResult> RunLlmReplyAsync(
...
+        var result = await SendReplyAsync(
             outboundIntent,
             sentSeed,
             reply.Activity.Conversation,
             inbound,
             registration,
             runtimeContext,
             ct);
+        if (result.Success)
+            _ = TrySwapTypingReactionToDoneAsync(inbound, registration, ct);
```
eanzhao, line 183: **P1: Cover the default streaming relay completion path**

This swap only runs when `ConversationGAgent` reaches `RunLlmReplyAsync`, but relay streaming is enabled by default (`NyxIdRelayOptions.StreamingRepliesEnabled = true`). On successful streamed replies, `HandleLlmReplyReadyAsync` returns from `TryCompleteStreamedReplyAsync` after final `RunStreamChunkAsync`/`PersistStreamedCompletionAsync`, so this code is never called. That leaves the inbound `Typing` reaction in place for the normal relay LLM path. The swap needs to be triggered from the streamed completion/final-chunk path as well, or exposed as a runner hook that `TryCompleteStreamedReplyAsync` can invoke after the final visible reply succeeds.

agents/Aevatar.GAgents.NyxidChat/ChannelConversationTurnRunner.cs
```diff
@@ -314,14 +318,17 @@ public async Task<ConversationStreamChunkResult> RunStreamChunkAsync(
...
         var result = await SendReplyAsync(
             replyContent,
             activity.Id,
             activity.Conversation,
-            ToInboundMessage(activity),
+            inbound,
             registration,
             runtimeContext,
             ct);
+        if (result.Success)
```
eanzhao, line 379: **P2: Avoid racing the fire-and-forget Typing POST on direct replies**

`RunInboundAsync` starts `TrySendImmediateLarkReactionAsync` without awaiting it, then direct slash/card flows can reach this swap almost immediately. If the list request runs before Lark has persisted the initial `Typing` reaction, `ParseAppReactionIds` sees no bot reaction, `DONE` is added, and the original Typing POST can complete afterwards, leaving both `Typing` and `DONE` on the message. Direct-reply paths need to sequence the initial reaction with the swap, or make the swap tolerate this race with a bounded retry/re-list before adding the final `DONE`.

eanzhao (commented):
Found two additional issues in the reaction swap path.

eanzhao (commented):
## Review Summary

Two real issues, both confirmed by code path tracing:

**P1 — Streaming relay path skips the reaction swap entirely.** This is the biggest gap. `StreamingRepliesEnabled` defaults to `true`, and when streaming succeeds, `HandleLlmReplyReadyAsync` returns early from `TryCompleteStreamedReplyAsync` at ConversationGAgent.cs:477, completely bypassing `RunLlmReplyAsync` (and its swap). The existing test `HandleLlmReplyReadyAsync_WhenStreamingSucceeded_PersistsCompletedWithoutInvokingRunLlmReply` explicitly asserts `runner.LlmReplyCount.ShouldBe(0)`. So on the most common production path (streamed LLM replies), the Typing reaction is never swapped to DONE.

**P2 — Race on direct-reply slash commands.** `RunInboundAsync` fires the Typing POST (line 62) with `_ =` fire-and-forget, then `TryHandleAgentBuilderAsync` can complete a fast direct reply and trigger the swap at line 331 before Lark has persisted the Typing reaction. The swap's GET list finds nothing, adds DONE, then the original Typing POST completes, leaving both reactions. Narrow window, but architecturally real. Card actions are safe (gate excludes non-Message types). LLM path is safe (LLM latency >> reaction POST latency).

The non-streaming relay and direct-slash paths that go through `RunLlmReplyAsync` look correct. The list-filter-delete-add sequence, `operator_type=app` filtering, error handling, and fire-and-forget exception safety are all well done.

eanzhao (commented):
## Fix Review — LGTM

Both P1 and P2 are addressed correctly.

**P1 (streaming path):** `OnReplyDeliveredAsync` hook with default no-op on the interface, called from `ConversationGAgent` only on the streaming-early-return branch. Non-streaming path goes through `RunLlmReplyAsync` with its own swap, no double-fire. Tests assert `OnReplyDeliveredCount` = 1 (streaming) and 0 (non-streaming).

**P2 (race condition):** `TrySendImmediateLarkReactionAsync` Task is captured and threaded through to `TryHandleAgentBuilderAsync` → `AwaitTypingReactionThenSwapAsync` which awaits it with a 2-second cap before the swap GET. `RunLlmReplyAsync` skips this guard (LLM latency >> POST latency). `TypingReactionGateHandler` test is well designed — parks the typing POST, asserts zero swap requests before release, then verifies POST→GET→DELETE→POST DONE ordering.

No issues found. Ship it.

PR #490, opened 2026-04-28
test(daily): cross-user isolation regression test for #437

### LINKED ISSUE #437: [Bug] `/daily` binding causes cross-user data leakage
#### 📌 Description
When using the Lark bot (Aevatar), after one user binds their account via /daily username, other users who run /daily will see the bound username’s data instead of their own.

Note: This issue also occurs when users interact with the bot via separate 1:1 (private) chats, not only in group chats.

This results in user data being shared across different users incorrectly.

---

#### 🔁 Steps to Reproduce

1. User A  sends a private message to the bot: /daily usernameA(binding completed)
2. User B sends a private message to the bot: /daily
3. Result:
- Returns data for `usernameA` ❌
- Instead of User B’s own data ✅

---

#### ✅ Expected Behavior
- `/daily` should return data based on the current user (user_id / open_id)
- Each user should have an isolated binding and data context

---

#### ❌ Actual Behavior
- `/daily` returns the most recently bound username (global state)
- No user-level isolation

eanzhao:
## Summary
- Adds a regression test pinning the exact cross-user scenario from issue #437: User A saves `alice-gh` via `/daily alice-gh`, User B runs `/daily` in a separate p2p chat with the same bot, and User B must NOT see `alice-gh`.
- The per-user scope fix (`ChannelUserConfigScope`, PR #438) already prevents the leak — this test ensures it stays fixed.
- Asserts that User B's per-user scope (`scope-1:lark:ou_bob`) is queried, not User A's (`scope-1:lark:ou_alice`) or the bot scope (`scope-1`).

## Test plan
- [x] New test `CrossUserIsolation_UserBDoesNotSeeUserASavedPreference` passes
- [x] All 9 existing github_username tests still pass
- [x] Full ChannelRuntime test suite (482 tests) passes

Closes #437


eanzhao:
## Review

Test logic looks solid — good use of the sentinel value `WRONG-bot-scope-leak` and the `DidNotReceive` assertions on Alice's scope / bot scope to catch regressions.

### Branch naming

The branch is `fix/2026-04-28_daily-cross-user-isolation`, but this PR only adds a regression test — the actual fix is already in #438. Per the naming convention (`<type>/YYYY-MM-DD_<purpose>`, where `<type>` ∈ {feat, fix, refactor, docs, test, chore}), this should be `test/2026-04-28_daily-cross-user-isolation`. The commit scope already correctly uses `test(daily):`, so the branch type should match.

### Minor observation

The existing `DerivesGithubUsername_FromNyxProxy_WhenArgumentAndPreferenceMissing` test doesn't assert `github_username` in the result JSON — it only checks the `HandleEventAsync` payload. The new test asserts both, which is strictly better, but the inconsistency is worth noting in case the result JSON contract isn't documented elsewhere.

Otherwise LGTM.

PR #492, opened 2026-04-28
enhance(daily): failure-notification slug fallback + chunked delivery (#423 §C)

### LINKED ISSUE #423: enhance(daily): richer report content + progressive delivery (streaming-edit or batched)

eanzhao:
## Summary

Closes the remaining acceptance items on #423 — the §C cross-cutting safety net.
After PRs #458 (§A structured prompt) and #469 (§B streaming-edit), this PR
delivers the last two items so the issue can close cleanly:

1. **§C-1 Failure-notification cross-tenant fallback.** When the primary outbound
   proxy is structurally rejected (e.g. cross-tenant `99992364`), the previous
   `TrySendFailureAsync` retried through the *same* proxy and the user saw
   nothing. Now the agent captures the inbound channel-bot's NyxID slug at
   create time — by definition reachable, since the user just messaged it —
   and `TrySendFailureAsync` tries that slug first, falling back to the primary
   only when the captured slug also rejects (or wasn't captured / equals
   primary).

2. **§C-2 Section-boundary chunked delivery.** Outputs > 30 KB used to be
   truncated at the cap with a `…[truncated]` marker. The new
   `SkillRunnerOutputChunker.Split()` paragraph-aware splitter keeps the full
   report by emitting `[part k/N • continued ↑/continues ↓]` chunks. chunk[0]
   flows through the existing streaming-edit sink (the in-flight message
   lands as part 1); chunks 1..N go as fresh one-shot POSTs.

## Implementation

| Layer | Change |
|---|---|
| Proto (`skill_runner.proto`) | `SkillRunnerOutboundConfig.failure_notification_provider_slug` (field 12) |
| Channel runtime (`ChannelMetadataKeys.cs`, `ChannelConversationTurnRunner.cs`) | New `InboundChannelBotProxySlug` key; `BuildReplyMetadata` populates it from `ChannelInboundEvent.NyxProviderSlug` |
| Agent builder (`AgentBuilderTool.cs`) | Refactored `ResolveProxyServiceIdsAsync` to also return the eligible slug map (no second `/user-services` round-trip). `ResolveFailureNotificationContext` decides whether to capture the inbound slug + expand `allowed_service_ids` to include its `UserService.id` |
| Skill runner (`SkillRunnerGAgent.cs`) | `SendOutputAsync` parameterized by `providerSlugOverride`; `TrySendFailureAsync` tries failure slug first then falls back to primary; `ExecuteSkillAsync` invokes `SkillRunnerOutputChunker.Split()` and dispatches chunk[0] via sink + chunks 1..N via `SendOutputAsync` |
| Output chunker (`SkillRunnerOutputChunker.cs`, new) | Pure-function paragraph-boundary splitter with char-boundary fallback for runaway single-paragraph inputs |
| Docs (`docs/canon/daily-command-pipeline.md`) | §3, §8, §9.3, §11, §17 updated to mark §C landed and reflect the new flow |

The two paths are independent and could ship separately, but they're both
small and both close the same issue, so one PR keeps the change ergonomically
scoped to "everything #423 §C asked for".

## Out of scope (deliberate)

- **Catalog (`UserAgentCatalogEntry`) not extended.** Runtime routing reads from
  actor State; the catalog is for listing/status. If a future workflow agent
  needs the same fallback exposed at the catalog reader level, add the field
  then — premature now.
- **Streaming sink does NOT mid-stream switch messages when accumulated text
  passes 30 KB.** Truncate-and-finalize-then-chunk keeps the streaming-edit
  UX (one growing message) for the common case and only fans out to N
  messages after the LLM finishes. The issue text recommends "split on section
  boundaries, send N messages" — this is what landed. A future enhancement
  could do multi-message streaming-edit, but it's a bigger surface and not
  required for §C.

## Test coverage

- **15 new tests, 633 existing tests still pass:**
  - `SkillRunnerOutputChunkerTests` (7) — empty, exactly-at-cap, multi-paragraph, no-paragraph hard split, content preservation, tail-only overflow, all-content-round-trip
  - `SkillRunnerGAgentTests` (5) — primary distinct → failure slug used; failure slug rejects → fall back; failure slug == primary → no double POST; empty failure slug (legacy) → primary path; both reject → swallowed
  - `AgentBuilderToolTests` (3) — captured + svc-id in `allowed_service_ids`; inbound == primary → empty; inbound missing from user-services → empty + creation succeeds + allowed_service_ids unchanged

## Test plan

- [x] `dotnet test test/Aevatar.GAgents.ChannelRuntime.Tests/...` — 648 passed
- [x] `dotnet test test/Aevatar.AI.Tests/...` — 483 passed
- [x] `dotnet test test/Aevatar.AI.Core.Tests/...` — 68 passed
- [x] `dotnet test test/Aevatar.Architecture.Tests/...` — 105 passed
- [x] `dotnet test test/Aevatar.GAgents.Channel.Protocol.Tests/...` — 123 passed
- [x] Targeted CI guards (query_projection_priming, projection_state_version, projection_state_mirror_current_state, workflow_binding_boundary, test_stability) — all pass
- [x] `bash tools/docs/lint.sh` — 32 files, 0 errors
- [ ] Manual smoke: revoke outbound `s/api-lark-bot` access for a test user → trigger `/run-agent` → verify the failure message lands via the inbound channel-bot slug
- [ ] Manual smoke: force a > 30 KB daily report → verify chunked delivery (streamed part 1 + N additional messages with continuation markers)

Closes #423.

### DIFF
--- a/agents/Aevatar.GAgents.Authoring.Lark/AgentBuilderTool.cs
+++ b/agents/Aevatar.GAgents.Authoring.Lark/AgentBuilderTool.cs
@@ -266,9 +266,22 @@ public sealed class AgentBuilderTool : IAgentTool
             return gitHubAuthorizationResponse;
 
         var providerSlug = (args.Str("nyx_provider_slug") ?? "api-lark-bot").Trim();
-        var requiredServiceIds = await ResolveProxyServiceIdsAsync(nyxClient, token, templateSpec!.RequiredServiceSlugs, ct);
-        if (requiredServiceIds.errorJson != null)
-            return requiredServiceIds.errorJson;
+        var serviceResolution = await ResolveProxyServiceIdsAsync(nyxClient, token, templateSpec!.RequiredServiceSlugs, ct);
+        if (serviceResolution.ErrorJson != null)
+            return serviceResolution.ErrorJson;
+
+        // Issue #423 §C — capture the inbound channel-bot slug as a failure-notification
+        // fallback. By definition the user can be reached through the bot they just
+        // messaged, so when a primary outbound delivery is rejected (e.g. cross-tenant
+        // Lark `99992364`) the failure-notification message can still land if the agent's
+        // API key is allowed to route through the inbound bot. Optional: if the inbound
+        // slug is not registered as a per-user UserService row (or equals the primary,
+        // in which case the fallback would just hit the same proxy), we leave the field
+        // empty and TrySendFailureAsync degrades to the current single-attempt behavior.
+        var failureNotificationContext = ResolveFailureNotificationContext(
+            providerSlug,
+            serviceResolution.RequiredIds!,
+            serviceResolution.EligibleIdBySlug);
 
         var agentId = string.IsNullOrWhiteSpace(args.Str("agent_id"))
             ? SkillRunnerDefaults.GenerateActorId()
@@ -276,7 +289,7 @@ public sealed class AgentBuilderTool : IAgentTool
 
         var createKeyResponse = await nyxClient.CreateApiKeyAsync(
             token,
-            BuildCreateApiKeyPayload(agentId, requiredServiceIds.value!),
+            BuildCreateApiKeyPayload(agentId, failureNotificationContext.AllowedServiceIds),
             ct);
 
         if (IsErrorPayload(createKeyResponse))
@@ -322,6 +335,7 @@ public sealed class AgentBuilderTool : IAgentTool
             LarkReceiveIdFallback = deliveryTarget.Fallback?.ReceiveId ?? string.Empty,
             LarkReceiveIdTypeFallback = deliveryTarget.Fallback?.ReceiveIdType ?? string.Empty,
             OwnerScope = caller.Clone(),
+            FailureNotificationProviderSlug = failureNotificationContext.FailureSlug ?? string.Empty,
         };
 #pragma warning restore CS0612
 
@@ -439,17 +453,17 @@ public sealed class AgentBuilderTool : IAgentTool
         // hardcoded `[providerSlug, publishProviderSlug]` was fine for two slugs but would
         // drift if a third slug were ever added; route through the spec so the source of
         // truth lives next to the workflow YAML.
-        var requiredServiceIds = await ResolveProxyServiceIdsAsync(
+        var serviceResolution = await ResolveProxyServiceIdsAsync(
             nyxClient,
             token,
             templateSpec!.RequiredServiceSlugs,
             ct);
-        if (requiredServiceIds.errorJson != null)
-            return requiredServiceIds.errorJson;
+        if (serviceResolution.ErrorJson != null)
+            return serviceResolution.ErrorJson;
 
         var createKeyResponse = await nyxClient.CreateApiKeyAsync(
             token,
-            BuildCreateApiKeyPayload(agentId, requiredServiceIds.value!),
+            BuildCreateApiKeyPayload(agentId, serviceResolution.RequiredIds!),
             ct);
 
         if (IsErrorPayload(createKeyResponse))
@@ -1040,29 +1054,45 @@ public sealed class AgentBuilderTool : IAgentTool
     /// slug, and skip org-shared rows the caller cannot use as a proxy target — those would later
     /// surface as a less-actionable <c>org_role_insufficient</c> error.</para>
     /// </remarks>
-    private async Task<(IReadOnlyList<string>? value, string? errorJson)> ResolveProxyServiceIdsAsync(
+    /// <summary>
+    /// Result of <see cref="ResolveProxyServiceIdsAsync"/>. <see cref="RequiredIds"/> /
+    /// <see cref="ErrorJson"/> are mutually exclusive (success vs. blocking error). Even on
+    /// success, callers can use <see cref="EligibleIdBySlug"/> to look up <em>optional</em>
+    /// slugs that were not in <c>requiredSlugs</c> — e.g. the inbound channel-bot slug for
+    /// SkillRunner's failure-notification fallback (issue #423 §C). Optional lookups must
+    /// not block agent creation, so they go through this map instead of being added to
+    /// <c>requiredSlugs</c> (which would cause <see cref="ResolveProxyServiceIdsAsync"/> to
+    /// return a <c>service_not_connected</c> error if the slug is missing).
+    /// </summary>
+    private readonly record struct ProxyServiceResolutionResult(
+        IReadOnlyList<string>? RequiredIds,
+        string? ErrorJson,
+        IReadOnlyDictionary<string, string> EligibleIdBySlug);
+
+    private async Task<ProxyServiceResolutionResult> ResolveProxyServiceIdsAsync(
         NyxIdApiClient client,
         string token,
         IReadOnlyList<string> requiredSlugs,
         CancellationToken ct)
     {
+        var emptyEligible = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
         if (requiredSlugs.Count == 0)
         {
-            return (null, JsonSerializer.Serialize(new
+            return new ProxyServiceResolutionResult(null, JsonSerializer.Serialize(new
             {
                 error = "no_required_slugs",
                 hint = "At least one required Nyx proxy service slug must be provided.",
-            }));
+            }), emptyEligible);
         }
 
         var response = await client.ListUserServicesAsync(token, ct);
         if (IsErrorPayload(response))
         {
-            return (null, JsonSerializer.Serialize(new
+            return new ProxyServiceResolutionResult(null, JsonSerializer.Serialize(new
             {
                 error = "user_services_unavailable",
                 hint = "Could not list connected Nyx user-services. Try again or check NyxID availability.",
-            }));
+            }), emptyEligible);
         }
 
         try
@@ -1120,17 +1150,29 @@ public sealed class AgentBuilderTool : IAgentTool
                 bestBySlug[slug] = candidate;
             }
 
+            // Snapshot the eligible (slug → id) map before the per-required-slug check so
+            // callers can look up optional slugs (e.g. inbound channel-bot for failure-
+            // notification fallback) without re-listing user-services. Ineligible rows are
+            // intentionally excluded — including them would let optional lookups silently
+            // pick up an inactive or org-viewer-only service the API key cannot route through.
+            var eligibleBySlug = bestBySlug
+                .Where(static pair => pair.Value.IsEligible)
+                .ToDictionary(
+                    pair => pair.Key,
+                    pair => pair.Value.Id,
+                    StringComparer.OrdinalIgnoreCase);
+
             var ids = new List<string>(requiredSlugs.Count);
             foreach (var slug in requiredSlugs.Distinct(StringComparer.OrdinalIgnoreCase))
             {
                 if (!bestBySlug.TryGetValue(slug, out var resolution))
                 {
-                    return (null, JsonSerializer.Serialize(new
+                    return new ProxyServiceResolutionResult(null, JsonSerializer.Serialize(new
                     {
                         error = "service_not_connected",
                         slug,
                         hint = $"NyxID has no connected user-service for slug `{slug}`. Connect the provider at NyxID before creating this agent.",
-                    }));
+                    }), emptyEligible);
                 }
 
                 if (resolution.IsEligible)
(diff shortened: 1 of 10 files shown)


eanzhao (commented):
Found one blocking edge case in the optional failure-notification fallback path.

agents/Aevatar.GAgents.Authoring.Lark/AgentBuilderTool.cs
```diff
@@ -1182,6 +1227,51 @@ private readonly record struct ServiceResolution(
...
+        // route through it. If it's not, leaving the failure-notification field empty
+        // keeps the runtime on the existing single-attempt path — better than persisting
+        // a slug whose every send would 403 at proxy enforcement time.
+        if (!eligibleIdBySlug.TryGetValue(inboundSlug, out var inboundId))
+            return new FailureNotificationContext(null, requiredIds);
+
+        // Dedupe — if the inbound slug's UserService.id is already in requiredIds the
+        // expanded list is identical, but we still surface the slug on OutboundConfig so
+        // the runtime knows to use it for failure notifications.
+        var allowed = requiredIds.Contains(inboundId, StringComparer.Ordinal)
```
eanzhao, line 1274: This optional fallback can accidentally make `/daily` creation fail for org-shared inbound bots. `/api/v1/user-services` includes org-inherited rows as `credential_source.type = org` with `allowed = true`, so `eligibleIdBySlug` can return an org-owned `UserService.id`. But NyxID `POST /api/v1/api-keys` validates `allowed_service_ids` against the new key owner's own active `UserService` rows (`user_id == key owner`). If the required services are personal but the inbound channel-bot slug only resolves through an org-owned service, this append sends an org-owned id in `allowed_service_ids` and the key creation fails with “UserService ... not found or not owned by user”. That turns the optional failure-notification fallback into a creation blocker. Please either only capture personal rows for this optional slug, or create/scope the child key through a contract that can accept the org-owned id, and add a regression test with `credential_source: { type: "org", allowed: true }`.

eanzhao (commented):
## Review Summary

Overall this is a well-structured PR — two independent features (failure-notification fallback + chunked delivery) cleanly separated with solid test coverage (15 new tests). The proto field numbering, backward-compatibility handling, and graceful degradation paths are all correct.

I found a few issues worth addressing before merge:

---

### 1. Case-sensitivity mismatch in slug comparison (should fix)

`AgentBuilderTool.cs:1255` — `ResolveFailureNotificationContext` compares `inboundSlug` vs `primarySlug` with `StringComparison.Ordinal` (case-sensitive), but the `eligibleIdBySlug` dictionary is built with `StringComparer.OrdinalIgnoreCase`. If inbound and primary slugs differ only in casing (e.g. `"API-LARK-BOT"` vs `"api-lark-bot"`), the equality check passes through, the dictionary lookup succeeds, and the failure-notification slug gets set to what is effectively the same proxy — the exact scenario the guard is meant to prevent.

```csharp
// Line 1255 — Ordinal but dictionary is OrdinalIgnoreCase
if (string.Equals(inboundSlug, primarySlug, StringComparison.Ordinal))
```

Should use `StringComparison.OrdinalIgnoreCase` to match the dictionary semantics, or align both consistently.

### 2. `MarkerOverhead = 60` is insufficient for >=100 chunks (theoretical but worth a defensive fix)

`SkillRunnerOutputChunker.cs:35` — For outputs producing >=100 chunks (each ~30K), marker text like `"[part 100/100 • continues ↓]"` is 31 chars. A middle chunk with both prefix and suffix would be `31 + rawContent + 31 = 30,002`, exceeding `MaxLarkTextLength = 30,000`. 

Practically this requires ~3MB output which won't happen for daily reports, but the invariant "every rendered chunk <= MaxLarkTextLength" should hold regardless. Options:
- Compute overhead dynamically from `rawChunks.Count` (two-digit vs three-digit formatting width)
- Or clamp and re-split at the render stage

### 3. Sink `FinalizeAsync` applies `TruncateForLark` on top of chunker output (harmless but confusing)

`SkillRunnerStreamingReplySink.cs:179` — `FinalizeAsync` calls `TruncateForLark` which truncates at 30K. The chunker already ensures each chunk is <= 30K, so this truncation never fires for chunked output. However, it means the sink silently re-truncates if someone changes the chunker logic without realizing the sink has its own cap. Consider adding a brief comment or assertion that the sink's truncation is a safety net for chunk[0] only.

### 4. Missing integration test for end-to-end chunked delivery

There's no test that exercises `DispatchOutputChunksAsync` with multi-chunk output through the sink path — verifying chunk[0] lands via `FinalizeAsync` and chunks 1..N via `SendOutputAsync`. The chunker tests and agent tests cover their respective units, but the glue between them is untested. A single integration test producing >30K output would close this gap.

### 5. Hard-split test doesn't assert rendered chunk size

`SkillRunnerOutputChunkerTests.Split_WithNoParagraphBoundaries_ShouldHardSplitAtBudget` doesn't assert that rendered chunks (with markers) stay <= `MaxLarkTextLength`. The paragraph-boundary test does this check; adding it to the hard-split test would catch a future `MarkerOverhead` regression.

eanzhao:
## Review: §C failure-notification fallback + chunked delivery

Overall: clean design, good test coverage (15 new tests), and the two features are well-scoped to #423. One real bug, a few nits.

### P1: Org-owned UserService rows can block `/daily` creation

Already raised by @eanzhao — `eligibleIdBySlug` includes org-inherited rows (`credential_source.type = org`), but `POST /api/v1/api-keys` validates `allowed_service_ids` against the owner's personal `UserService` rows. If the inbound channel-bot only resolves through an org-owned service, appending its id to `allowed_service_ids` will fail key creation with "not found or not owned by user". The optional fallback becomes a creation blocker.

Fix: filter `eligibleIdBySlug` to personal-only rows, or catch the key-creation error and retry without the inbound slug.

### P2: Chunker `LastIndexOf` comment is misleading

`SkillRunnerOutputChunker.cs:68-73` — the comment says the `count` argument "intentionally overshoots by 1 at the trailing edge so a `\n\n` anchored at `(offset + contentBudget - 1)` with its second `\n` just outside the budget is still picked up". But `LastIndexOf(string, int startIndex, int count)` searches the range `[startIndex - count + 1, startIndex]` — with `count = contentBudget` and `startIndex = offset + contentBudget - 1`, the window is `[offset, offset + contentBudget - 1]`. A `\n\n` *starting* at `offset + contentBudget - 1` has its second char at `offset + contentBudget`, which is outside the window and won't be found. The "overshoot" doesn't help here.

Practical impact: if the only `\n\n` in the budget window straddles the boundary, we hard-split instead of splitting at the paragraph. No data loss, just slightly suboptimal chunking. Either fix the comment to match reality, or adjust the search to `count = contentBudget + 1` to actually catch the straddling case.

### P3: Case-sensitivity inconsistency in slug comparison

`ResolveFailureNotificationContext` compares `inboundSlug` vs `primarySlug` with `StringComparison.Ordinal` (case-sensitive), but the `eligibleIdBySlug` dictionary uses `StringComparer.OrdinalIgnoreCase`. If NyxID ever returns mixed-case slugs, the equality check could miss a match that the dictionary finds. Low risk today (slugs are lowercase), but worth a `StringComparison.OrdinalIgnoreCase` on the equality check for consistency.

### Test quality

Tests are thorough: happy path, fallback, same-slug skip, legacy empty-slug, both-reject swallow. The `SequencedHandler` for multi-request assertions is a nice pattern. Chunker tests cover empty, exact-cap, paragraph boundaries, no-paragraph hard split, content preservation, and tail-only overflow. Solid.

**Verdict:** P1 blocks merge (can break `/daily` creation for org-shared bots). P2-P3 are follow-ups.

PR #495, opened 2026-04-28
Fix retired ChannelRuntime startup cleanup

eanzhao:
## Problem

Persisted runtime actors from the retired `Aevatar.GAgents.ChannelRuntime` assembly can be activated during pod startup before the new registration projection startup path rebuilds them. Orleans then fails to resolve old agent/materialization-context type names and the host can abort startup.

## Solution

- Add a maintenance-only event-store reset capability and implement it for InMemory, File, and Garnet event stores.
- Add `RetiredChannelRuntimeActorCleanupHostedService` in Mainnet host composition. It takes a one-time event-store marker lease, probes persisted actor runtime types, removes retired root/projection actors, removes stale relays, clears stale read models, resets retired actor streams, and writes a completion marker.
- Register the cleanup hosted service before ChannelRuntime, Device, and Scheduled projection startup services.
- Document the upgrade/runbook contract for operators.

## Impact Paths

- `src/Aevatar.Mainnet.Host.Api/Hosting/Migration/*`
- `src/Aevatar.Foundation.*/*EventStore*`
- `src/Aevatar.Mainnet.Host.Api/Hosting/MainnetHostBuilderExtensions.cs`
- `docs/operations/2026-04-28-retired-channelruntime-cleanup-runbook.md`

## Validation

- `dotnet test test/Aevatar.Foundation.Core.Tests/Aevatar.Foundation.Core.Tests.csproj --nologo --filter "ResetStreamAsync"`
- `dotnet test test/Aevatar.Hosting.Tests/Aevatar.Hosting.Tests.csproj --nologo --filter "RetiredChannelRuntimeActorCleanupHostedServiceTests|AddAevatarMainnetHost_ShouldRunRetiredChannelRuntimeCleanup"`
- `bash tools/ci/test_stability_guards.sh`
- `bash tools/ci/query_projection_priming_guard.sh`

PR #497, opened 2026-04-28
refactor(skill-runner): split into SkillDefinitionGAgent + SkillExecutionGAgent

### LINKED ISSUE #491: refactor(skill-runner): split SkillRunnerGAgent into SkillDefinitionGAgent + SkillExecutionGAgent
## Background

Supersedes #447. The lifecycle-split diagnosis in #447 is correct — `SkillRunnerGAgent` mixes long-lived subscription state with per-execution session state, causing retry/schedule collisions and making #439 hard to fix cleanly. However, the proposed solution (per-template actor types like `DailyReportSubscriptionGAgent` + `DailyReportRunGAgent`) is wrong:

- It bakes business-specific templates into code-level types — every new template requires two new actor classes + proto + events + commands (O(n) expansion cost).
- `DailyReportSubscriptionGAgent` and `SocialMediaPostSubscriptionGAgent` would be near-identical code, differing only in proto field names — duplicate abstraction.
- Violates "单一主干，插件扩展" and "内核最小化" — template differentiation should be plugin/configuration, not parallel type hierarchies.

The core insight: **SkillRunnerGAgent + different system prompt + toolset = a new agent**. Business differentiation is data-driven (skill_content, toolset config, execution_prompt), not type-driven. This is already the correct model — the fix should preserve it.

## Proposed direction

Split into two **generic** actors, keeping template differences purely in instance data:

| Responsibility | Actor | Lifecycle |
|---|---|---|
| Subscription / schedule / configuration | `SkillDefinitionGAgent` | Long-lived, fact owner |
| Single execution | `SkillExecutionGAgent` | Session-scoped, one per fire |

### SkillDefinitionGAgent (renamed from SkillRunnerGAgent)

- **State**: `skill_name`, `template_name`, `skill_content` (system prompt), `execution_prompt`, `schedule_cron`, `schedule_timezone`, `provider_name`, `model`, `temperature`, `max_tokens`, `max_tool_rounds`, `max_history_messages`, `enabled`, `scope_id`, `outbound_config`, `toolset_ref` (if applicable)
- **Commands**: `InitializeSkillDefinition`, `Enable`, `Disable`, `UpdateSchedule`, `UpdateSkillContent`, `Delete`
- **Events**: `SkillDefinitionInitialized`, `SkillDefinitionEnabled`, `SkillDefinitionDisabled`, `NextRunScheduled`, `SkillDefinitionTombstoned`
- **Zero per-execution fields** — no `last_run_at`, `error_count`, `last_output`, `retry_attempt`
- Owns `ChannelScheduleRunner`; on each cron fire, spawns a `SkillExecutionGAgent` with the current config snapshot
- Business naming: "a user-configured scheduled skill" IS the business entity

### SkillExecutionGAgent (new, generic)

- **State**: `subscription_id`, `scheduled_at`, `started_at`, `completed_at`, `output`, `errors[]`, `retry_attempts`, `status`
- **Commands**: `StartExecution`, internal continuations
- **Events**: `ExecutionStarted`, `ExecutionStepCompleted`, `ToolCallFailed`, `ExecutionCompleted`, `ExecutionFailed`
- Receives system prompt + toolset config + execution context from the definition actor at spawn time
- Owns retry semantics in full — definition actor doesn't know what "retry" is
- Naturally discardable / archivable; run history is its own projection

### Why this is better than #447

- **New template = new configuration, zero code changes** — daily report, social media, any future template are all `SkillDefinitionGAgent` instances with different `skill_content` + toolset
- #439 fix localizes to `SkillExecutionGAgent` — `ExecutionFailed` event when tool calls fail; definition state stays clean
- Schedule-vs-retry collision eliminated — different actors, different lifetimes
- Existing commands (`/daily`, `/run-agent`, `/disable-agent`, `/enable-agent`, `/delete-agent`) keep working — just routes to renamed actor underneath
- Fully aligned with CLAUDE.md: "单一主干，插件扩展", "内核最小化", "Actor 以业务命名", "默认短生命周期", "长期 actor 限定事实拥有者"

## Acceptance

- [ ] No actor named `SkillRunnerGAgent` in the codebase (renamed to `SkillDefinitionGAgent`)
- [ ] `SkillDefinitionGAgent` state contains zero per-execution fields
- [ ] Each execution is one `SkillExecutionGAgent` activation; on retry the same execution actor resumes
- [ ] Different templates (daily_report, social_media, etc.) are purely instance-level configuration differences — no per-template actor types
- [ ] Existing slash commands all keep working with new actor names underneath
- [ ] Migration path: existing `SkillRunnerGAgent` instances convert to `SkillDefinitionGAgent` on next activation

## Dependencies

- #refactor-credential-actor — credential actor referenced by definition
- #refactor-prompt-template — versioned prompt referenced by definition

## Related

- #447 — superseded by this issue (lifecycle split correct, per-template types wrong)
- #439 — silent failure; fix becomes natural after split
- #423 — content / delivery enhancements; cleaner to layer on top of split actors

eanzhao:
## Summary

Splits `SkillRunnerGAgent` into two generic actors, keeping template differences purely in instance data (#491):

- **`SkillDefinitionGAgent`** (long-lived fact owner): owns subscription, schedule, config, outbound config. On each cron fire, spawns a session-scoped execution actor with a config snapshot.
- **`SkillExecutionGAgent`** (session-scoped): owns LLM execution, streaming-edit delivery, tool failure safety net (#439), and retry semantics. Naturally discardable; run history is its own projection.
- New template = new configuration, zero code changes — daily report, social media, any future template are all `SkillDefinitionGAgent` instances with different `skill_content` + toolset.
- CI guard `tools/ci/skill_runner_lifecycle_split_guard.sh` validates: no `SkillRunnerGAgent` class, no per-execution fields in definition state, no per-template actor types.
- Legacy aliases map existing persisted `SkillRunnerState` events to `SkillDefinitionState` for seamless migration on next activation.

## Changes

| Area | What changed |
|------|-------------|
| Proto | New `skill_definition.proto` (definition state with `reserved 8,10,11,12` for dropped execution fields) and `skill_execution.proto` (session state + config snapshot) |
| Actors | `SkillRunnerGAgent` → `SkillDefinitionGAgent` (GAgentBase, no LLM) + new `SkillExecutionGAgent` (AIGAgentBase, LLM execution) |
| Command port | `ISkillRunnerCommandPort` → `ISkillDefinitionCommandPort` + `SkillDefinitionCommandPort` |
| Defaults | `SkillRunnerDefaults` → `SkillDefinitionDefaults` (same constant values for backward compat) |
| DI | Updated `ScheduledServiceCollectionExtensions` registration |
| Integration | `AgentBuilderTool`, `AgentBuilderCardFlow`, `NyxRelayAgentBuilderFlow` updated to new types |
| Migration | Legacy aliases + backward-compat command handlers on `SkillDefinitionGAgent` for existing persisted events/durable callbacks |
| CI guard | `tools/ci/skill_runner_lifecycle_split_guard.sh` |
| Tests | All 667 channel runtime tests + 105 architecture tests pass |

## Test plan

- [x] `dotnet build aevatar.slnx` — build succeeds
- [x] `dotnet test test/Aevatar.GAgents.ChannelRuntime.Tests/` — 667/667 pass
- [x] `dotnet test test/Aevatar.Architecture.Tests/` — 105/105 pass
- [x] `bash tools/ci/skill_runner_lifecycle_split_guard.sh` — ok
- [x] `bash tools/ci/architecture_guards.sh` — all relevant guards pass

Closes #491


eanzhao:
## Review: SkillDefinitionGAgent + SkillExecutionGAgent Split

Overall the split is clean — long-lived definition actor vs session-scoped execution actor is the right architecture per #491 and aligns with the AGENTS.md lifecycle rules. A few issues worth addressing:

---

### 1. HandleTriggerAsync: fire-and-forget execution dispatch with no error propagation

In `SkillDefinitionGAgent.HandleTriggerAsync`, the execution actor is created and dispatched inside a catch-all that only logs:

```csharp
catch (Exception ex)
{
    Logger.LogWarning(ex, "Skill definition {ActorId} failed to spawn execution {ExecutionId}", Id, executionId);
}
await Scheduler.ScheduleNextRunAsync(now, CancellationToken.None);
```

If the execution actor spawn or dispatch fails (actor runtime overloaded, transient infrastructure error), the definition actor silently schedules the next cron. No `SkillExecutionFailedEvent` is ever emitted, no registry status update occurs, and the user has zero visibility into the missed execution. Per AGENTS.md: _"已提交领域事件必须可观察"_ — if the failure is never committed as a domain event, it is invisible to the system.

**Suggestion:** On spawn/dispatch failure, emit a domain event and update the registry status so the missed run is visible in `/agent-status`.

---

### 2. SkillExecutionGAgent.HandleStartAsync: retry counter resets are lost on crash

In `HandleStartAsync`, on failure the retry is scheduled via `ScheduleRetryAsync` which stores the `_retryLease` as an in-memory field. If the actor deactivates between retries (it's session-scoped, so this is plausible), `_retryLease` is lost. When the durable callback fires and `HandleRetryAsync` runs, it receives `command.RetryAttempt` from the callback payload — so the retry attempt counter is correct. However, the actor will re-execute from scratch, and the `HandleStartAsync` path on fresh activation won't be triggered (only `HandleRetryAsync` fires from the callback). This seems fine for the retry path.

**But:** There's a subtle bug — `HandleStartAsync` does `await PersistDomainEventAsync(started)` then runs `ExecuteSkillAsync`. If the actor crashes mid-execution (e.g. process restart), on reactivation the actor replays `SkillExecutionStartedEvent` and has `status = "running"` in state, but **no one will ever trigger the retry or mark it failed**. The durable timeout for retry is only scheduled _after_ `ExecuteSkillAsync` throws. A crash before that point means the execution is permanently stuck in `"running"` with no recovery path.

**Suggestion:** Either schedule a watchdog timeout immediately after persisting the started event (so a crash is recovered), or document that this is acceptable for session-scoped actors.

---

### 3. UpdateRegistryExecutionAsync uses definitionId as the registry key, but error_count is overwritten to 0 on success

In `HandleStartAsync` success path:
```csharp
await UpdateRegistryExecutionAsync(
    SkillDefinitionDefaults.StatusRunning,
    Timestamp.FromDateTimeOffset(now),
    0,          // errorCount always 0
    string.Empty,
    CancellationToken.None);
```

And in the failure path:
```csharp
await UpdateRegistryExecutionAsync(
    SkillDefinitionDefaults.StatusError,
    Timestamp.FromDateTimeOffset(now),
    1,          // hardcoded to 1, not cumulative
    ex.Message,
    CancellationToken.None);
```

The `errorCount` in the failure path is always `1` regardless of how many retries actually failed. Since this updates the _definition's_ registry entry, it should probably reflect the actual retry attempt count (e.g. `State.RetryAttempts + 1`).

---

### 4. _retryLease is a mutable field on the actor but is not part of persisted state

`_retryLease` is an in-memory `RuntimeCallbackLease?` field. Per AGENTS.md: _"Actor/模块运行态只能在事件处理主线程修改"_ — this is fine for single-threaded actor turns. But if the actor is re-activated after deactivation, `_retryLease` is null, and `CancelRetryLeaseAsync` becomes a no-op. This means the old durable callback lease is leaked (never cancelled) if the execution completes after a reactivation. The durable callback will eventually fire and call `HandleRetryAsync` on a completed/stale execution, which will try to re-execute.

The `HandleRetryAsync` method doesn't guard against re-execution on an already-completed actor (no status check like `if (State.Status != "running") return;`).

**Suggestion:** Add an idempotency guard at the top of `HandleRetryAsync` (and `HandleStartAsync`).

---

### 5. SkillRunnerToolFailureCounter — name not updated

The class `SkillRunnerToolFailureCounter` and `SkillRunnerStreamingReplySink` still carry the `SkillRunner` prefix while everything else has been renamed to `SkillDefinition`/`SkillExecution`. The CI guard (`skill_runner_lifecycle_split_guard.sh`) only checks for `SkillRunnerGAgent` class, so these are technically allowed, but the naming inconsistency makes the codebase harder to navigate. Consider renaming to `SkillExecutionToolFailureCounter` etc. in a follow-up.

---

### Minor

- `SkillExecutionState` field 9 is skipped (jumps from 8 to 10). This is fine for protobuf but worth a comment in the proto for future readers.
- The `SkillDefinitionDefaults.GenerateExecutionId` method uses `Guid.NewGuid():N` which is non-deterministic. If `HandleTriggerAsync` is retried (e.g. at-least-once delivery), a new execution actor is created each time. This may be intentional but should be documented.

eanzhao:
## Review

The split of `SkillRunnerGAgent` into `SkillDefinitionGAgent` (long-lived fact owner) + `SkillExecutionGAgent` (session-scoped) is architecturally sound and well-aligned with the CLAUDE.md rules. Proto design is clean with proper `reserved` fields. Legacy migration via aliases + compat handlers is thorough. Good CI guard.

**Issues to fix:**

### 1. HandleTriggerAsync swallows spawn failures and still advances the schedule

`SkillDefinitionGAgent.cs:124-168`: If `actorRuntime.CreateAsync` or `dispatchPort.DispatchAsync` throws, the catch block logs a warning and falls through to `ScheduleNextRunAsync`. This means if the execution actor can never be created (e.g., DI misconfiguration), the definition will keep firing on schedule, logging warnings, and producing no output indefinitely — with no error surfaced to the catalog or user.

**Recommendation**: Either persist a `SkillDefinitionExecutionFailedEvent` (so the catalog reflects the error), or re-throw after logging so the durable callback system can surface the failure. At minimum, update the registry status to `StatusError` in the catch block.

### 2. `_retryLease` should not be mutable instance state on the actor

`SkillExecutionGAgent.cs:26`: The field `_retryLease` is written in `ScheduleRetryAsync` and `CancelRetryLeaseAsync` during event handling. While safe due to single-threaded actor turns, this is a durable callback handle that can be restored from state on reactivation. If the execution actor is passivated and reactivated mid-retry, `_retryLease` will be null and `CancelRetryLeaseAsync` will silently skip cancellation — the retry will still fire (durable callback survives passivation), but the agent won't be able to cancel it if it completes before the retry fires.

**Recommendation**: Either (a) don't store the lease at all (let the durable callback fire and handle it idempotently in `HandleRetryAsync`), or (b) persist the lease in actor state so it survives passivation.

### 3. `HandleEnableAsync` skips the event when already enabled

`SkillDefinitionGAgent.cs:197-210`: If `State.Enabled` is already `true`, no `SkillDefinitionEnabledEvent` is persisted but `ScheduleNextRunAsync` and `UpdateRegistryStatusAsync` are still called. This is subtly different from `HandleDisableAsync` which always persists an event. The asymmetry is intentional (idempotent enable), but worth documenting with a comment since it could look like an oversight.

### 4. `UpdateRegistryStatusAsync` always sends `ErrorCount = 0`

`SkillDefinitionGAgent.cs:256-264`: This method hardcodes `ErrorCount = 0` and `LastError = string.Empty`. It's called from `UpsertRegistryAsync` (after initialization) and `HandleDisableAsync`/`HandleEnableAsync`. While the execution agent's `UpdateRegistryExecutionAsync` corrects the error count after execution, there's a window after `HandleDisableAsync` where the error count from a previous failed execution gets wiped to 0. Consider preserving the existing error count or only resetting it on `HandleEnableAsync`.

### 5. Minor: test file still named `SkillRunnerGAgentTests.cs`

The test class was renamed to `SkillDefinitionGAgentTests` but the file is still `SkillRunnerGAgentTests.cs`. Not a blocker but inconsistent with the naming convention.

---

Otherwise the design is solid. The execution config snapshot pattern (copy definition config into `StartSkillExecutionCommand` → persist in `SkillExecutionStartedEvent`) ensures execution state is self-contained and survives replay. The `SkillRunnerToolFailureCounter` being per-agent-instance via `BuildToolMiddlewareChain` is correct — no cross-execution contamination. The CI guard script is well-structured.

eanzhao (commented):
**P2: `SkillExecutionState.scheduled_at` never populated**

`SkillExecutionGAgent.cs` `HandleStartAsync` receives `StartSkillExecutionCommand.scheduled_at` (set by the definition actor at `SkillDefinitionGAgent.cs:133`) but the value is discarded:

1. `SkillExecutionStartedEvent` has no `scheduled_at` field
2. `ApplyStarted` (`SkillExecutionGAgent.cs:515`) does not set `next.ScheduledAt`

Result: `SkillExecutionState.scheduled_at` stays null forever. Either propagate the field through the event and applier, or add `reserved 2` to `SkillExecutionState` to match how `SkillDefinitionState` handled fields 8/10/11/12.

agents/Aevatar.GAgents.Scheduled/SkillDefinitionCommandPort.cs
```diff
@@ -24,52 +24,49 @@ public SkillRunnerCommandPort(
...
-        await EnsureSkillRunnerActorAsync(agentId, ct);
-        await DispatchAsync(agentId, new EnableSkillRunnerCommand { Reason = reason ?? string.Empty }, ct);
+        await EnsureDefinitionActorAsync(agentId, ct);
+        await DispatchAsync(agentId, new EnableSkillDefinitionCommand { Reason = reason ?? string.Empty }, ct);
     }
 
-    private async Task EnsureSkillRunnerActorAsync(string agentId, CancellationToken ct)
+    private async Task EnsureDefinitionActorAsync(string agentId, CancellationToken ct)
     {
         _ = await _actorRuntime.GetAsync(agentId)
```
eanzhao, line 71: Existing scheduled agents were created with the Orleans runtime state `AgentTypeName` set to the old `SkillRunnerGAgent` assembly-qualified type. `GetAsync(agentId)` only checks that this stored type name is non-empty, so it will return a non-null actor here, but when the grain later handles the envelope it calls `Type.GetType` on the deleted `SkillRunnerGAgent` type, fails initialization, and drops the command. That means existing `/daily` agents cannot be triggered/disabled/enabled after this deploy. Please add an actor-type migration/alias path before treating existing actors as valid `SkillDefinitionGAgent`s, or keep a compatible old type that forwards to the new definition actor.

agents/Aevatar.GAgents.Scheduled/SkillDefinitionGAgent.cs
```diff
@@ -0,0 +1,354 @@
...
+            TemplateName = evt.TemplateName,
+            SkillContent = evt.SkillContent,
+            ExecutionPrompt = evt.ExecutionPrompt,
+            ScheduleCron = evt.ScheduleCron,
+            ScheduleTimezone = evt.ScheduleTimezone,
+            OutboundConfig = evt.OutboundConfig,
+            Enabled = evt.Enabled,
+            ScopeId = evt.ScopeId,
+            ProviderName = evt.ProviderName,
+            Model = evt.Model,
```
eanzhao, line 714: This legacy replay path drops the optional LLM tuning fields from persisted `SkillRunnerInitializedEvent`s (`temperature`, `max_tokens`, `max_tool_rounds`, `max_history_messages`). Existing agents that customized these will silently replay into `SkillDefinitionState` with temperature/max_tokens unset and max tool/history reset to defaults, changing runtime behavior after the split. Please copy the `Has*` fields here (and in `HandleLegacyInitializeAsync`) the same way `HandleInitializeAsync` does.

eanzhao (commented):
## Review

整体架构拆分方向正确，生命周期分离清晰，proto 定义和 legacy alias 的迁移设计合理。发现几个问题需要确认或修改：

### 1. `SkillDefinitionGAgent.HandleTriggerAsync` 中 spawn execution 失败的异常被静默吞掉

`SkillDefinitionGAgent.cs` 的 `HandleTriggerAsync` 方法中，`try-catch` 在 `CreateAsync`/`DispatchAsync` 失败时只记录 warning，不会向调用方或 registry 报告失败。这意味着：
- 用户手动 trigger 后，如果 execution actor 创建失败，没有任何错误反馈
- registry 中的 `last_error` / `error_count` 不会更新
- 用户无法感知到持续失败

建议：在 catch 中至少更新 registry status 为 error，或者 persist 一个事件让 projection 可以观察到失败。

### 2. `SkillExecutionGAgent.HandleRetryAsync` 没有更新 `State.RetryAttempts`

`SkillExecutionGAgent.cs` 的 `HandleRetryAsync` 使用 `command.RetryAttempt` 做判断，但 `State.RetryAttempts` 在 retry 链中不会被更新（`ApplyStarted` 没有设置它，retry 路径也没有 persist 事件更新它）。当前 `MaxRetryAttempts = 1` 所以这个 bug 不会触发，但如果将来增加重试次数，会导致基于 state 的判断出错。建议增加一个 `SkillExecutionRetryStartedEvent` 来更新 `RetryAttempts`。

### 3. `ChannelArchitectureTests` 中 `SkillDefinitionGAgent` 的 proactiveCallerPatterns 注释已过时

`ChannelArchitectureTests.cs` 中 `SkillDefinitionGAgent` 被加入 `proactiveCallerPatterns`，但它现在不再直接调用任何 channel/Conversation 方法（只负责创建 `SkillExecutionGAgent`）。注释描述 "must dispatch through ConversationGAgent command envelopes rather than directly invoking IChannelOutboundPort.ContinueConversationAsync" 对 `SkillDefinitionGAgent` 已不准确。建议更新注释说明，或考虑是否需要将其保留在该列表中。

### 4. `SkillExecutionState` proto 中 field 9 跳过缺少说明

`skill_execution.proto` 中 `SkillExecutionState` 从 field 8 (`status`) 跳到 field 10 (`skill_content`)，跳过了 field 9。虽然 protobuf 允许，但建议加一行注释说明 field 9 的预留意图，避免后续维护者困惑。

### 5. 缺少 `SkillExecutionGAgent` 核心逻辑（retry、failure）的独立单元测试

当前 `SkillDefinitionGAgentTests.cs` 主要测试了 `SendOutputAsync`，但 `HandleStartAsync` 的 catch-retry 路径、`HandleRetryAsync` 的最终失败路径等核心逻辑缺少独立的单元测试。虽然这些路径在旧版中也没有被直接测试，但拆分后 `SkillExecutionGAgent` 是一个新的核心 actor，建议至少增加 retry 调度和失败处理的 mock 测试（无需真实 LLM provider）。

---

其他方面（proto reserved 字段、legacy alias、backward-compat handlers、CI guard）都处理得很好，没有问题。

eanzhao:
**Blocking on #498**

This PR deletes `Aevatar.GAgents.Scheduled.SkillRunnerGAgent` and replaces it
with `SkillDefinitionGAgent` + `SkillExecutionGAgent`. Live SkillRunner actors
in production have `RuntimeActorGrainState.AgentTypeName =
"Aevatar.GAgents.Scheduled.SkillRunnerGAgent"` written by every prior daily
activation. After deployment, `RuntimeActorGrain.OnActivateAsync` calls
`ResolveAgentType` → `Type.GetType` + `AppDomain.GetAssemblies()` and finds
nothing — the class no longer exists. Activation fails; subsequent envelopes
are dropped (`Dropping envelope for actor {ActorId}: initialization failed`).

`SkillRunnerLegacyAliases.cs` (`[LegacyClrTypeName]` / `[LegacyProtoFullName]`)
only handles payload codec compatibility via
`ProtobufContractCompatibility.IsCompatibleClrTypeName` — it does **not**
rewrite the runtime actor implementation CLR type stored in
`RuntimeActorGrainState.AgentTypeName`. This is exactly the activation-failure
mode PR #495 fixes for the deleted `ChannelRuntime` types, but applied to live
business actors (every scheduled daily / social / report skill) instead of
retired ones.

Per #498, the correct resolution is the stable `AgentKind` + `IAgentKindRegistry`
identity model, not a one-off retirement spec or `preserve/migrate` runtime-type
rewrite. Once #498 Phase 1 lands in `dev`, this PR ships as a kind-alias
registration on `SkillDefinitionGAgent` (e.g. `[LegacyAgentKind("skill-runner")]`)
with zero production state mutation and natural mixed-version rollout safety.

**Do not merge until #498 Phase 1 is in `dev` and this PR is rebased on top of
it.** Architecture-correct path is the long-term direction; we are not
shortcutting it with a one-time governance patch.

agents/Aevatar.GAgents.Scheduled/SkillExecutionGAgent.cs
```diff
@@ -0,0 +1,642 @@
...
+        {
+            var output = await ExecuteSkillAsync(now, command.Reason, CancellationToken.None);
+
+            await PersistDomainEventAsync(new SkillExecutionCompletedEvent
+            {
+                CompletedAt = Timestamp.FromDateTimeOffset(now),
+                Output = output,
+            });
+
+            await UpdateRegistryExecutionAsync(
```
eanzhao: **[codex]** severity=blocker, category=arch

`SkillExecutionGAgent` can complete after `SkillDefinitionGAgent.HandleDisableAsync` has already written `disabled`, but this unconditional execution-side catalog update sets the definition row back to `running` (and the failure path sets it to `error`). That violates CLAUDE.md: single authoritative owner / actor lifecycle because enabled status is owned by the long-lived definition actor, not the session actor. Route execution completion/failure back to the definition actor with an execution id/config version and let the definition actor update the catalog only if the result is still current.

agents/Aevatar.GAgents.Scheduled/SkillDefinitionGAgent.cs
```diff
@@ -0,0 +1,418 @@
...
+            return;
+        }
+
+        var now = DateTimeOffset.UtcNow;
+        var executionId = SkillDefinitionDefaults.GenerateExecutionId(Id);
+
+        Exception? dispatchFailure = null;
+        try
+        {
+            var actorRuntime = Services.GetRequiredService<IActorRuntime>();
```
eanzhao: **[Consensus: 2 models]** severity=major, category=arch

Both reviewers flag service-locator usage at the trigger handler entry: `Services.GetRequiredService<IActorRuntime>()` and `Services.GetRequiredService<IActorDispatchPort>()` are resolved per-call inside the GAgent. This is a service-locator anti-pattern — depends are hidden, testability suffers, and resolution cost is paid on every trigger. Inject these via the constructor (or via the GAgent base if framework-mediated) and remove the `Services.GetRequiredService<>` calls.

<details>

- **v4-pro**: `Services.GetRequiredService<IActorRuntime>()` and `Services.GetRequiredService<IActorDispatchPort>()` (line 170) are resolved via service locator inside an event handler. Per CLAUDE.md command/envelope rules: "传输载体必须可替换 … 上层依赖投递契约，不依赖具体载体" and "禁止在基类、业务层或中间适配层绕过标准 publisher 直接操作底层传输对象". These are core runtime services — they should be constructor-injected (even if optional, with `= null` fallback) rather than resolved per-invocation from `IServiceProvider`. The constructor chain already accepts nullable DI params; add `IActorRuntime?` and `IActorDispatchPort?` via constructor injection, then fall back to `Services.GetRequiredService` only when the ctor param is null (backward compat for unit tests that don't compose full DI).
- **mimo-v2.5-pro**: Service-locator anti-pattern: `Services.GetRequiredService<IActorRuntime>()` and `Services.GetRequiredService<IActorDispatchPort>()` (line 170) resolve dependencies inline inside `HandleTriggerAsync`. This violates the repo convention of constructor injection and makes the handler harder to unit test (requires a full `IServiceProvider` setup). Inject both `IActorRuntime` and `IActorDispatchPort` as `private readonly` constructor parameters, matching the pattern used by `SkillDefinitionCommandPort` which already injects both.
</details>

agents/Aevatar.GAgents.Scheduled/SkillDefinitionGAgent.cs
```diff
@@ -0,0 +1,418 @@
...
+            {
+                ExecutionId = executionId,
+                FailedAt = Timestamp.FromDateTimeOffset(now),
+                Reason = command.Reason?.Trim() ?? "schedule",
+                Error = ex.Message,
+            });
+            dispatchFailure = ex;
+        }
+
+        await Scheduler.ScheduleNextRunAsync(now, CancellationToken.None);
```
eanzhao, line 242: **[codex]** severity=major, category=bug

After the split, scheduling the next durable callback no longer projects the new `State.NextRunAt` to `UserAgentCatalog`; the execution actor later sends `UserAgentCatalogExecutionUpdateCommand` without `NextRunAt`, and the catalog reducer overwrites the existing value with null. `/agent-status` will lose the next-run timestamp after the first run. Have the definition actor publish the next-run catalog update after scheduling, and make execution result updates unable to clear definition-owned schedule fields.

agents/Aevatar.GAgents.Scheduled/SkillExecutionGAgent.cs
```diff
@@ -0,0 +1,642 @@
...
+        {
+            DefinitionId = command.DefinitionId ?? string.Empty,
+            ScheduledAt = command.ScheduledAt,
+            StartedAt = Timestamp.FromDateTimeOffset(DateTimeOffset.UtcNow),
+            Reason = command.Reason ?? string.Empty,
+            SkillContent = command.SkillContent ?? string.Empty,
+            ExecutionPrompt = command.ExecutionPrompt ?? string.Empty,
+            OutboundConfig = command.OutboundConfig?.Clone() ?? new SkillRunnerOutboundConfig(),
+            ScopeId = command.ScopeId ?? string.Empty,
+            ProviderName = command.ProviderName ?? string.Empty,
```
eanzhao, line 160: **[glm-5.1]** severity=major, category=bug

**Pre-scheduled retry outside try-catch; no cancellation on success.**

The old `SkillRunnerGAgent` scheduled retries **only in the failure catch path** and explicitly called `CancelRetryLeaseAsync` on success. The new code pre-schedules the retry **before** the try block and never cancels it:

```csharp
await PersistDomainEventAsync(started);
if (SkillDefinitionDefaults.MaxRetryAttempts > 0)
    await ScheduleRetryAsync(State.RetryAttempts + 1, CancellationToken.None); // outside try

var now = DateTimeOffset.UtcNow;
try
{
    var output = await ExecuteSkillAsync(...);
    // success path: no retry cancellation
}
```

Two problems:
1. If `ScheduleSelfDurableTimeoutAsync` throws, the started event was already persisted (status = "running") but no retry is scheduled and no execution occurred — the actor is **stuck** with no recovery path.
2. On success, the durable timeout fires after 30s and is silently ignored by the status guard — a leaked timer slot per execution.

Suggested fix — move retry scheduling into the catch path (matching old semantics) and track the lease for cancellation:
```csharp
try
{
    var output = await ExecuteSkillAsync(now, command.Reason, CancellationToken.None);
    await PersistDomainEventAsync(new SkillExecutionCompletedEvent { ... });
    await CancelRetryLeaseAsync(CancellationToken.None); // cancel if retry was pending
    ...
}
catch (Exception ex)
{
    if (State.RetryAttempts < SkillDefinitionDefaults.MaxRetryAttempts)
    {
        await ScheduleRetryAsync(State.RetryAttempts + 1, CancellationToken.None);
        return;
    }
    ...
}
```

agents/Aevatar.GAgents.Scheduled/SkillExecutionGAgent.cs
```diff
@@ -0,0 +1,642 @@
...
+            await ScheduleRetryAsync(State.RetryAttempts + 1, CancellationToken.None);
+
+        var now = DateTimeOffset.UtcNow;
+        try
+        {
+            var output = await ExecuteSkillAsync(now, command.Reason, CancellationToken.None);
+
+            await PersistDomainEventAsync(new SkillExecutionCompletedEvent
+            {
+                CompletedAt = Timestamp.FromDateTimeOffset(now),
```
eanzhao: **[v4-pro]** severity=major, category=bug

`completed_at` timestamp is recorded as `Timestamp.FromDateTimeOffset(now)` where `now` was captured at line 159 *before* `ExecuteSkillAsync` ran (which may take 30–60 s of LLM streaming). The catalog's `last_run_at` will report the start time, not the actual completion time. Same bug exists in `HandleRetryAsync` at line 233. Fix: capture a fresh `var completedAt = DateTimeOffset.UtcNow` after `ExecuteSkillAsync` returns, and use it for both `CompletedAt` and `LastRunAt`.

```csharp
var completedAt = DateTimeOffset.UtcNow;
await PersistDomainEventAsync(new SkillExecutionCompletedEvent
{
    CompletedAt = Timestamp.FromDateTimeOffset(completedAt), ...
});
await UpdateRegistryExecutionAsync(..., Timestamp.FromDateTimeOffset(completedAt), ...);
```

test/Aevatar.Architecture.Tests/Rules/ChannelArchitectureTests.cs
```diff
@@ -146,7 +146,7 @@ public void OutboundSendAsync_Callers_Must_StayInsideAdapters_Or_TurnContext()
     [Fact]
     public void ProactiveCallers_MustNot_Invoke_ContinueConversationAsync_Directly()
     {
-        // Rule: proactive actor-to-actor paths (SkillRunnerGAgent, WorkflowAgentGAgent, admin endpoint
+        // Rule: proactive actor-to-actor paths (SkillExecutionGAgent, WorkflowAgentGAgent, admin endpoint
```
eanzhao, line 149: **[Consensus: 2 models]** severity=major, category=bug

The architecture-rule comment was updated to reference `SkillExecutionGAgent` (the actor that now performs outbound delivery), but the actual guarded pattern in the array doesn't match the comment — the rule still targets `SkillDefinitionGAgent` (or vice-versa). After the split, the proactive-caller pattern that previously named `SkillRunnerGAgent` should be re-aimed at the new owner. Either fix the guard pattern to match the comment, or fix the comment to match the pattern, then re-run the architecture tests to confirm the rule still bites the right actor.

<details>

- **v4-pro**: Comment at line 149 was updated to say `SkillExecutionGAgent` (the actor that does outbound delivery) but the proactive-caller pattern array at line 156 was changed to `SkillDefinitionGAgent` (which only spawns execution actors and never calls `ContinueConversationAsync` directly). The guard should match both actors (or at minimum `SkillExecutionGAgent` which actually sends outbound messages). After the split, `SkillExecutionGAgent` uses `NyxIdApiClient.ProxyRequestAsync`, not `ContinueConversationAsync`, so it may not need a guard — but the comment/pattern mismatch will confuse future maintainers. Either align the pattern to the comment (`SkillExecutionGAgent`) or drop both actors from the list entirely since neither invokes `ContinueConversationAsync`.
- **codex**: The rule comment was updated to name `SkillExecutionGAgent`, but the guarded pattern was changed to `SkillDefinitionGAgent`. The proactive outbound code moved to `SkillExecutionGAgent`, so this guard now misses the class it is supposed to protect; replace/add `SkillExecutionGAgent` here so direct `ContinueConversationAsync` calls in the execution actor fail the architecture test.
</details>

agents/Aevatar.GAgents.Scheduled/SkillDefinitionGAgent.cs
```diff
@@ -0,0 +1,418 @@
...
+
+        var now = DateTimeOffset.UtcNow;
+        var executionId = SkillDefinitionDefaults.GenerateExecutionId(Id);
+
+        Exception? dispatchFailure = null;
+        try
+        {
+            var actorRuntime = Services.GetRequiredService<IActorRuntime>();
+            _ = await actorRuntime.GetAsync(executionId)
+                ?? await actorRuntime.CreateAsync<SkillExecutionGAgent>(executionId, CancellationToken.None);
```
eanzhao: **[v4-pro]** severity=minor, category=concurrency

`CancellationToken.None` is passed to `CreateAsync`, `DispatchAsync`, `ScheduleNextRunAsync`, and `UpdateRegistryExecutionErrorAsync` inside `HandleTriggerAsync`. Event handler methods in this framework don't receive a cancellation token parameter, so this is a framework limitation rather than a caller bug. However, it means runtime shutdown cannot cancel an in-flight spawn-dispatch-schedule chain. Consider adding a `CancellationToken` parameter to the `[EventHandler]` method signature if the framework supports it, or wiring a timeout through the dispatch port.

agents/Aevatar.GAgents.Scheduled/SkillDefinitionGAgent.cs
```diff
@@ -0,0 +1,418 @@
...
+    {
+        await HandleEnableAsync(new EnableSkillDefinitionCommand { Reason = command.Reason });
+    }
+
+    private async Task UpsertRegistryAsync(string status, CancellationToken ct)
+    {
+#pragma warning disable CS0612 // legacy field reads/writes during owner_scope migration
+        var legacyOwnerNyxUserId = State.OutboundConfig?.OwnerNyxUserId ?? string.Empty;
+        var legacyPlatform = ResolvePlatform(State.OutboundConfig?.Platform);
+        var ownerScope = State.OutboundConfig?.OwnerScope
```
eanzhao, line 405: **[glm-5.1]** severity=minor, category=di

**Service-locator pattern on every trigger.** `IActorRuntime` and `IActorDispatchPort` are resolved from `Services.GetRequiredService` on each trigger invocation. These are singleton services — resolve once in `OnActivateAsync` and cache as fields to avoid per-trigger service-location overhead.

```csharp
private IActorRuntime? _actorRuntime;
private IActorDispatchPort? _dispatchPort;

protected override async Task OnActivateAsync(CancellationToken ct)
{
    await base.OnActivateAsync(ct);
    _actorRuntime = Services.GetRequiredService<IActorRuntime>();
    _dispatchPort = Services.GetRequiredService<IActorDispatchPort>();
    await Scheduler.BootstrapOnActivateAsync(ct);
}
```

test/Aevatar.GAgents.ChannelRuntime.Tests/SkillDefinitionGAgentTests.cs
```diff
@@ -16,11 +16,11 @@
 
 namespace Aevatar.GAgents.ChannelRuntime.Tests;
 
-public sealed class SkillRunnerGAgentTests : IAsyncLifetime
+public sealed class SkillDefinitionGAgentTests : IAsyncLifetime
```
eanzhao, line 20: **[mimo-v2.5-pro]** severity=minor, category=test

Misleading test class name: the class is `SkillDefinitionGAgentTests` but the majority of tests (SendOutputAsync, TrySendFailureAsync, BuildExecutionMetadata, all fallback/retry tests) exercise `SkillExecutionGAgent` behavior via `CreateActivatedExecutionAgent` helper. Consider either (a) splitting into `SkillDefinitionGAgentTests` + `SkillExecutionGAgentTests`, or (b) renaming the file to match the class (`SkillDefinitionGAgentTests.cs` → already correct class name, but file still named `SkillRunnerGAgentTests.cs`).

eanzhao:
## Fix-check follow-up

5 of 6 multi-model reviewers re-evaluated the 13 inline comments after the latest push. **11 resolved, 1 partial, 1 unresolved.** Big-ticket items (state-ownership blocker, NextRunAt projection, retry placement, stale `completed_at`, legacy field preservation, service-locator anti-pattern) all confirmed resolved by majority.

Two items still need attention:

### ⚠️ Partial — `test/Aevatar.GAgents.ChannelRuntime.Tests/SkillDefinitionGAgentTests.cs:19` (4/5 partial, 1 unresolved)

The file was renamed `SkillRunnerGAgentTests.cs → SkillDefinitionGAgentTests.cs` so file/class names now align. However, the class still mixes execution-actor tests (`SendOutputAsync`, `TrySendFailureAsync`, `BuildExecutionMetadata`, fallback/retry tests via `CreateActivatedExecutionAgent`) into a class named for the definition actor. glm-5.1 dissents to *unresolved*; the others see it as a meaningful but incomplete fix.

**Suggested fix:** split into `SkillDefinitionGAgentTests` + `SkillExecutionGAgentTests`.

### ❌ Unresolved — `agents/Aevatar.GAgents.Scheduled/SkillDefinitionGAgent.cs:146` (3/5 unresolved, 2 cannot_determine)

`CancellationToken.None` is still passed to `CreateAsync`, `DispatchAsync`, `ScheduleNextRunAsync`, and registry update calls inside `HandleTriggerAsync`. The original comment acknowledges this is a framework limitation (`[EventHandler]` methods don't receive a token), and 2 reviewers (mimo, kimi) explicitly cite that as `cannot_determine`. **Likely acceptable to leave as-is** — but if the framework gains a per-handler timeout/cancel hook later, this is the call-site to revisit.

### Codex dissent on the blocker (#3166478164, otherwise resolved 4/5)

The lifecycle/ownership fix routes execution results through `SkillDefinitionGAgent`, which the majority confirms resolves the original concern. Codex marks this *partial* because the `Report*Command` path carries `execution_id` but no config/version stamp — if a definition is disabled+re-enabled with new config mid-run, the stale execution's completion could still update the catalog. Worth a quick check in `HandleExecutionCompletedAsync` to confirm `State.Enabled` alone is sufficient, or whether a config-version field is needed.

---

<sub>Generated by `/opencode-pr-fix-check`. Reviewers: deepseek-v4-pro, glm-5.1, mimo-v2.5-pro, kimi, codex. Gemini dropped (V8 OOM on the diff size).</sub>

agents/Aevatar.GAgents.Scheduled/SkillDefinitionGAgent.cs
```diff
@@ -0,0 +1,592 @@
...
+            {
+                Reason = command.Reason?.Trim() ?? string.Empty,
+            });
+        }
+
+        // Enable is intentionally idempotent: already-enabled actors refresh scheduling
+        // and catalog visibility without committing another lifecycle event.
+        await Scheduler.ScheduleNextRunAsync(DateTimeOffset.UtcNow, CancellationToken.None);
+        await UpdateRegistryStatusAsync(
+            SkillDefinitionDefaults.StatusRunning,
```
eanzhao, line 305: **[Consensus: 4 models]** severity=major, category=bug

Four reviewers independently flag that `HandleEnableAsync` calls `UpdateRegistryStatusAsync(preserveExecutionState: false)`, which sends `LastRunAt = null`, `ErrorCount = 0`, and an empty `LastError` through `UserAgentCatalogExecutionUpdateCommand`. The reducer applies these fields directly, so re-enabling a previously-disabled definition wipes the accumulated execution state (last completed run timestamp, accumulated error count, last error message) — the user loses observability into what happened before disable. The old `SkillRunnerGAgent.HandleEnableAsync` preserved this. **Fix:** pass `preserveExecutionState: true` from `HandleEnableAsync` (and audit `HandleDisableAsync` for the same issue) so enable/disable only changes status, not execution-history fields.

<details>

- **v4-pro** (line 271): `HandleEnableAsync` calls `UpdateRegistryStatusAsync(preserveExecutionState: false)`, which sets `PreserveLastRunAt = false` and leaves `LastRunAt` unset (null). This clears the catalog's `last_run_at` field on every enable. The old `SkillRunnerGAgent.HandleEnableAsync` preserved `State.LastRunAt`. Since `SkillDefinitionState` no longer tracks `LastRunAt`, this is structurally inevitable, but consider setting `PreserveLastRunAt = true` and omitting `LastRunAt` to avoid wiping the catalog value on enable.
- **mimo-v2.5-pro** (line 258): Re-enable now clears the catalog's `last_error` / `error_count` (`preserveExecutionState: false`). The old `SkillRunnerGAgent.HandleEnableAsync` forwarded existing `State.ErrorCount` / `State.LastError` to the catalog update, so a re-enabled agent preserved its failure history in `/agent-status`. If this is intentional (clean slate on enable), consider adding a code comment explaining the rationale; otherwise, pass `preserveExecutionState: true` to keep the error breadcrumb visible to operators.
- **codex** (line 272): `Enable` passes `preserveExecutionState: false`, which sends `LastRunAt = null`, `ErrorCount = 0`, and an empty `LastError` through `UpdateRegistryStatusAsync`. Re-enabling a disabled agent therefore erases the visible last execution state, unlike `Disable` and the previous `SkillRunnerGAgent` behavior. Preserve last-run state on enable, or split the preserve flags so enable can update status/next-run while explicitly deciding whether to clear only the error fields.
- **glm-5.1** (line 303): `preserveExecutionState: false` causes `UpdateRegistryStatusAsync` to emit a `UserAgentCatalogExecutionUpdateCommand` with `PreserveLastRunAt = false` and `LastRunAt = null` (protobuf default), which **clears the catalog entry's `last_run_at`**.

The old `SkillRunnerGAgent.HandleEnableAsync` preserved `last_run_at` explicitly:
```csharp
await UpdateRegistryExecutionAsync(
    SkillRunnerDefaults.StatusRunning, State.LastRunAt, State.NextRunAt,
    State.ErrorCount, State.LastError, CancellationToken.None);
```

After this PR, a disable → enable cycle makes `/agent-status` show *"Last run: n/a"* instead of the actual last run time.

**Fix:** change to `preserveExecutionState: true` so the catalog preserves both `last_run_at` and error state on re-enable (matching old behavior). If clearing error state on re-enable is intentional, split the flags: set `PreserveLastRunAt = true` while keeping `PreserveErrorState = false`.
</details>

agents/Aevatar.GAgents.Scheduled/SkillDefinitionCommandPort.cs
```diff
@@ -24,52 +24,49 @@ public SkillRunnerCommandPort(
 
     public async Task InitializeAsync(
         string agentId,
-        InitializeSkillRunnerCommand command,
+        InitializeSkillDefinitionCommand command,
         bool runImmediately,
         CancellationToken ct = default)
     {
         ArgumentException.ThrowIfNullOrWhiteSpace(agentId);
         ArgumentNullException.ThrowIfNull(command);
```
eanzhao, line 33: **[kimi]** severity=major, category=arch

The comment justifying `_catalogProjectionPort.EnsureProjectionForActorAsync` was deleted, yet the call remains in a command-port hot path. Per AGENTS.md §"不得 query-time priming", projection scope activation must not happen synchronously inside command/query dispatch. Removing the comment papers over the violation rather than fixing it. Either move priming to an explicit activation/lease/binder flow, or restore the comment and add a TODO linking to the refactor issue.

agents/Aevatar.GAgents.Scheduled/SkillDefinitionGAgent.cs
```diff
@@ -0,0 +1,592 @@
...
+        {
+            ExecutionId = command.ExecutionId.Trim(),
+            CompletedAt = completedAt,
+        };
+        if (command.HasDefinitionConfigRevision)
+            completed.DefinitionConfigRevision = command.DefinitionConfigRevision;
+
+        await PersistDomainEventAsync(completed);
+
+        await UpdateRegistryExecutionResultAsync(
```
eanzhao, line 350: **[codex]** severity=major, category=concurrency

`IsStaleExecutionReport` only filters reports from older config revisions, so two executions spawned from the same definition revision can complete out of order and the older run can overwrite the newer run's catalog status/`LastRunAt`/error state. Include a definition-owned run sequence or scheduled-at token in `StartSkillExecutionCommand` and both report commands, then ignore reports older than the latest observed run; add an overlap test where run B finishes before run A.

agents/Aevatar.GAgents.Scheduled/SkillDefinitionGAgent.cs
```diff
@@ -0,0 +1,592 @@
...
+
+    private static SkillDefinitionState ApplyExecutionFailed(
+        SkillDefinitionState current,
+        SkillDefinitionExecutionFailedEvent _) => current;
+
+    private static SkillDefinitionState ApplyDisabled(SkillDefinitionState current, SkillDefinitionDisabledEvent _)
+    {
+        var next = current.Clone();
+        next.Enabled = false;
+        next.NextRunAt = null;
```
eanzhao, line 688: **[kimi]** severity=major, category=design

`ApplyDisabled` and `ApplyEnabled` both increment `ConfigRevision`, which causes `IsStaleExecutionReport` to silently drop in-flight execution reports when a definition is toggled while an execution is running. The catalog read-model then misses the actual last-run timestamp and error state. Lifecycle events (disable/enable) should not conflate with config mutations; consider only bumping `ConfigRevision` in `ApplyInitialized` so spawned executions remain reportable unless their actual config snapshot is obsolete.

agents/Aevatar.GAgents.Scheduled/SkillExecutionGAgent.cs
```diff
@@ -0,0 +1,685 @@
...
+        }
+
+        var completedAt = DateTimeOffset.UtcNow;
+        await PersistDomainEventAsync(new SkillExecutionCompletedEvent
+        {
+            CompletedAt = Timestamp.FromDateTimeOffset(completedAt),
+            Output = output,
+        });
+
+        await ReportExecutionCompletedAsync(
```
eanzhao, line 219: **[codex]** severity=major, category=concurrency

After `SkillExecutionCompletedEvent` is persisted, `ReportExecutionCompletedAsync` is the only path that tells the definition/catalog about the terminal state. If `DispatchPort.DispatchAsync` throws, or the process dies after the local event commit and before this dispatch, replay leaves the execution in `completed`/`failed` and duplicate `Start` is ignored, so the definition never observes the run. Persist a pending-report state/event and deliver it via durable self-retry/outbox semantics, or at minimum catch report dispatch failures and schedule a retry before ending the turn.

test/Aevatar.GAgents.ChannelRuntime.Tests/SkillExecutionGAgentTests.cs
```diff
@@ -36,255 +34,27 @@ public Task DisposeAsync()
         return Task.CompletedTask;
```
eanzhao, line 36: **[glm-5.1]** severity=major, category=test

The execution agent's core lifecycle has no direct test coverage. The following paths are untested:

- `HandleStartAsync`: duplicate-start guard, execution failure → retry scheduling, execution failure after retries exhausted → `SkillExecutionFailedEvent` + report-back
- `HandleRetryAsync`: stale-status guard, retry exhaustion path
- `ReportExecutionCompletedAsync` / `ReportExecutionFailedAsync`: round-trip dispatch back to the definition actor
- The definition ↔ execution report flow end-to-end (including `IsStaleExecutionReport` rejection)

The old `SkillRunnerGAgentTests` covered the monolithic trigger path. After the split, the `SkillExecutionGAgentTests` only tests output/send behavior by seeding state via event store, and `SkillDefinitionGAgentTests` covers the definition side. The execution agent's state machine transitions are the riskiest untested surface in this refactor.

tools/ci/skill_runner_lifecycle_split_guard.sh
```diff
@@ -0,0 +1,91 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd -- "${SCRIPT_DIR}/../.." && pwd)"
+cd "${REPO_ROOT}"
+
+# Issue #491 guard: enforce SkillRunner → SkillDefinition + SkillExecution lifecycle split.
```
eanzhao, line 9: **[codex]** severity=major, category=test

This adds a standalone guard, but it is not invoked by `tools/ci/architecture_guards.sh` or any changed CI entrypoint, so the #491 invariant is not actually enforced unless someone runs this script manually. Wire this script into the existing architecture guard/CI workflow so regressions fail automatically.

agents/Aevatar.GAgents.Scheduled/SkillDefinitionGAgent.cs
```diff
@@ -0,0 +1,592 @@
...
+                OutboundConfig = State.OutboundConfig?.Clone() ?? new SkillRunnerOutboundConfig(),
+                ScopeId = State.ScopeId ?? string.Empty,
+                ProviderName = State.ProviderName ?? SkillDefinitionDefaults.DefaultProviderName,
+                Model = State.Model ?? string.Empty,
+                DefinitionConfigRevision = State.ConfigRevision,
+            };
+
+            if (State.HasTemperature)
+                startCommand.Temperature = State.Temperature;
+            if (State.HasMaxTokens)
```
eanzhao, line 211: **[Consensus: 2 models]** severity=minor, category=design

Two reviewers flag complementary aspects of the same concurrency hole in `HandleTriggerAsync`: (a) the next cron run is scheduled **immediately after dispatch**, before the current execution completes (glm-5.1@169); (b) every trigger unconditionally spawns a new execution actor with a fresh GUID, with no in-flight dedup (v4-pro@179). Together this means: if execution duration > cron interval, the next trigger fires while the previous run is still active, and a parallel execution actor is spawned — the old `SkillRunnerGAgent` serialised this in a single actor. **Fix options:** (1) move `ScheduleNextRunAsync` into the completion handler so the next run is only scheduled after the current finishes; or (2) add an in-flight execution check in `HandleTriggerAsync` and skip dispatch (or queue) when a previous execution is still active.

<details>

- **v4-pro** (line 179): `HandleTriggerAsync` unconditionally spawns a new execution actor (`GenerateExecutionId` includes a GUID), which means concurrent triggers (e.g. cron + manual) produce parallel executions. The old `SkillRunnerGAgent` was single-threaded and serialized triggers naturally. This behavioral change is likely intentional per the session-scoped design, but worth documenting that overlapping runs are now possible.
- **glm-5.1** (line 169): The next cron run is scheduled **immediately after dispatching** to the execution actor, not after execution completes. The old `SkillRunnerGAgent` scheduled after execution finished, which naturally serialized runs — a slow execution would delay the next schedule. In the new design, a long-running execution (> cron interval) causes the next trigger to spawn a second concurrent execution actor for the same definition. This is fine for daily reports (the primary use case) but is a behavioral change worth documenting. If future skills use frequent crons, consider adding a guard (e.g., an `ActiveExecutionId` field on definition state that prevents spawning while an execution is in-flight).
</details>

agents/Aevatar.GAgents.Scheduled/ScheduledRetiredActorSpec.cs
```diff
@@ -258,13 +258,13 @@ private static bool IsGeneratedUserAgent(string? agentId, string? agentType)
...
 
-        if (string.Equals(agentType, SkillRunnerDefaults.AgentType, StringComparison.Ordinal) ||
+        if (string.Equals(agentType, SkillDefinitionDefaults.AgentType, StringComparison.Ordinal) ||
             string.Equals(agentType, WorkflowAgentDefaults.AgentType, StringComparison.Ordinal))
         {
             return true;
         }
 
-        return normalizedId.StartsWith($"{SkillRunnerDefaults.ActorIdPrefix}-", StringComparison.Ordinal) ||
+        return normalizedId.StartsWith($"{SkillDefinitionDefaults.ActorIdPrefix}-", StringComparison.Ordinal) ||
```
eanzhao, line 270: **[v4-pro]** severity=minor, category=bug

The `IsGeneratedUserAgent` ID-prefix fallback matches `normalizedId.StartsWith("skill-runner-")`. Execution actor IDs follow the pattern `"skill-runner-{guid}-exec-{guid}"`, which also starts with `"skill-runner-"`. If an execution actor is persisted without agent-type metadata (edge case), the retired-actor cleanup could incorrectly classify it as a generated user agent. The agent-type check on line 261 would normally prevent this, but the prefix fallback remains a latent mismatch.

agents/Aevatar.GAgents.Scheduled/SkillDefinitionGAgent.cs
```diff
@@ -0,0 +1,592 @@
...
+    private static SkillDefinitionState ApplyInitialized(SkillDefinitionState current, SkillDefinitionInitializedEvent evt)
+    {
+        var next = current.Clone();
+        next.SkillName = evt.SkillName ?? string.Empty;
+        next.TemplateName = evt.TemplateName ?? string.Empty;
+        next.SkillContent = evt.SkillContent ?? string.Empty;
+        next.ExecutionPrompt = evt.ExecutionPrompt ?? string.Empty;
+        next.ScheduleCron = evt.ScheduleCron ?? string.Empty;
+        next.ScheduleTimezone = NormalizeTimezone(evt.ScheduleTimezone);
+        next.OutboundConfig = evt.OutboundConfig?.Clone() ?? new SkillRunnerOutboundConfig();
```
eanzhao, line 590: **[mimo-v2.5-pro]** severity=minor, category=design

`ApplyExecutionDispatchFailed` is a no-op (`=> current`), so the dispatch-failure data persisted via `SkillDefinitionExecutionDispatchFailedEvent` (execution ID, error message) is silently dropped from actor state on replay. The failure is only visible through the catalog's `StatusError` update. Either record the failure summary in state (e.g. `last_dispatch_error`) or delete the applier and event type if they're purely an audit log.

docs/canon/daily-command-pipeline.md
```diff
@@ -530,7 +530,7 @@ string failure_notification_provider_slug = 12;  // §C 旁路 proxy slug（入
 
 ### 12.3 单元测试 — SkillRunner actor
 
-文件：`test/Aevatar.GAgents.ChannelRuntime.Tests/SkillRunnerGAgentTests.cs`
+文件：`test/Aevatar.GAgents.ChannelRuntime.Tests/SkillDefinitionGAgentTests.cs`
```
eanzhao: **[codex]** severity=minor, category=arch

CLAUDE.md marks `docs/canon/` as the authoritative reference, but this edit only renames one test file while the same canonical document still describes the runtime as creating/executing `SkillRunnerGAgent` in the surrounding sections. Update the whole pipeline description to the `SkillDefinitionGAgent` + `SkillExecutionGAgent` split so the authoritative docs match the architecture change.

test/Aevatar.GAgents.ChannelRuntime.Tests/SkillRunnerStreamingReplySinkTests.cs
```diff
@@ -125,7 +125,7 @@ public async Task InitialPost_RejectedAsBotNotInChat_ViaHttp400Envelope_RetriesO
         // Production failures arrive through `NyxIdApiClient.SendAsync` as an HTTP-400 Nyx
         // envelope (`{"error": true, "status": 400, "body": "<raw json>"}`) — the same
         // wrapping shape pinned for the non-streaming path in
-        // `SkillRunnerGAgentTests.SendOutputAsync_ShouldRetryWithFallback_When_PrimaryRejectedAsBotNotInChat_ViaHttp400Envelope`.
+        // `SkillDefinitionGAgentTests.SendOutputAsync_ShouldRetryWithFallback_When_PrimaryRejectedAsBotNotInChat_ViaHttp400Envelope`.
```
eanzhao: **[kimi]** severity=minor, category=test

The comment cross-reference was updated to `SkillDefinitionGAgentTests`, but `SendOutputAsync` behavior (including the HTTP-400 envelope retry test) now lives in `SkillExecutionGAgentTests`. Update the comment to point to the correct test file so future maintainers find the coverage.

test/Aevatar.GAgents.ChannelRuntime.Tests/SkillRunnerToolFailureSafetyNetTests.cs
```diff
@@ -15,11 +15,11 @@ namespace Aevatar.GAgents.ChannelRuntime.Tests;
 ///
 ///   - <see cref="SkillRunnerToolFailureCounter"/>: state primitive
 ///   - <see cref="NyxIdProxyToolFailureCountingMiddleware"/>: classification + counting
-///   - <see cref="SkillRunnerGAgent.EnsureToolStatusAllowsCompletion"/>: failure policy
+///   - <see cref="SkillExecutionGAgent.EnsureToolStatusAllowsCompletion"/>: failure policy
 ///   - End-to-end wiring: middleware registered on the agent feeds the agent's counter
 ///
 /// We deliberately don't drive the full LLM loop in these tests — see the existing
-/// SkillRunnerGAgentTests pattern: ChatStreamAsync requires a live LLM provider, and the
+/// SkillDefinitionGAgentTests pattern: ChatStreamAsync requires a live LLM provider, and the
```
eanzhao: **[kimi]** severity=minor, category=test

The comment references `SkillDefinitionGAgentTests` for the `ChatStreamAsync` pattern, but `SkillDefinitionGAgent` is no longer an `AIGAgentBase` and has no LLM loop. The relevant coverage moved to `SkillExecutionGAgentTests`. Update the cross-reference.

agents/Aevatar.GAgents.Scheduled/ISkillDefinitionCommandPort.cs
```diff
@@ -0,0 +1,16 @@
+namespace Aevatar.GAgents.Scheduled;
```
eanzhao, line 1: **[mimo-v2.5-pro]** severity=nit, category=style

The old `ISkillRunnerCommandPort` had XML doc comments explaining actor lifecycle, catalog priming, and envelope dispatch responsibilities. `ISkillDefinitionCommandPort` drops all of them. Consider restoring at least a one-liner interface doc so callers understand the contract (especially the `InitializeAsync` dispatch-then-catalog-prime ordering guarantee tested in `HandleInitializeAsync_ShouldAwaitUpsertDispatchBeforeFiringExecutionUpdate`).

test/Aevatar.GAgents.ChannelRuntime.Tests/SkillExecutionGAgentTests.cs
```diff
@@ -964,17 +757,17 @@ private static ServiceProvider BuildServiceProvider(
         return services.BuildServiceProvider();
```
eanzhao, line 953: **[kimi]** severity=nit, category=test

`CreateInitializeCommand` returns `InitializeSkillDefinitionCommand`, which is never consumed by any test in this execution-agent test file. It is leftover from the pre-split `SkillRunnerGAgentTests`. Remove this dead helper to keep the test surface focused.

eanzhao:
## 我自己读完两个 fix commit (`1f86735a`, `3a2b46a0`) 之后的 review

按"该不该 block 这个 PR"分组。

### 真正的 bug —— 建议 merge 前修

**1. `HandleEnableAsync` 仍然会清掉执行历史** (`SkillDefinitionGAgent.cs:272`)

之前 4-model consensus 是对的。重新启用一个被禁用的 definition 时执行：

```csharp
await UpdateRegistryStatusAsync(
    SkillDefinitionDefaults.StatusRunning,
    preserveExecutionState: false,     // ← 清掉 LastRunAt, ErrorCount, LastError
    clearNextRunAt: false,
    ct: CancellationToken.None);
```

更糟糕的是它和 `ApplyEnabled` 里 bump `ConfigRevision`(line 537) 叠加 —— 禁用前在跑的那个 execution 会被 `IsStaleExecutionReport` 直接 drop 掉，**同时**目录里的 `last_run_at` 被清空。用户失去了"上次成功跑是什么时候"的观测能力。

**修复:** 改成 `preserveExecutionState: true`。如果确实希望 re-enable 时清掉错误以便用户重试，单独走一条只清 `error_count`/`last_error` 但保留 `last_run_at` 的路径。

---

**2. 并发漏洞: cron 在 completion 前就排掉了 + 不去重 in-flight execution** (`SkillDefinitionGAgent.cs:155, 208`)

```csharp
// line 155 —— 每次 trigger 都生成新的 execution_id (GUID), 不检查是否已有在跑的
var executionId = SkillDefinitionDefaults.GenerateExecutionId(Id);
// ...
// line 208 —— 下一次 cron dispatch 完就排了, 不等 execution 完成
await Scheduler.ScheduleNextRunAsync(now, CancellationToken.None);
```

如果 execution 时长 > cron 间隔，下一次 trigger 就会在前一次还没结束时触发。两个 execution actor 并行跑，回报时携带相同的 `ConfigRevision` —— stale-report guard 抓不住。catalog 状态最后写入者覆盖前面的。

老的 `SkillRunnerGAgent` 是单 actor 串行跑的，split 之后这个不变量丢了。

**修复:** 把 `Scheduler.ScheduleNextRunAsync` 移到 completion/failure handler 里再调用 (或者在 `HandleTriggerAsync` 里检查是否已有 in-flight execution，有就跳过/排队)。这一改之后并发问题自然消失。

---

**3. CI guard 没接进去** (`tools/ci/skill_runner_lifecycle_split_guard.sh`)

codex 之前的 #7。grep 验证过 —— `tools/ci/architecture_guards.sh` 完全没引用 `skill_runner_lifecycle_split_guard.sh`。脚本加了但永远不会被跑到。

**修复:** 在 `architecture_guards.sh` 里加一行调用，参考其他 split-guard 的接法。

---

**4. `HandleStartAsync` 没有任何直接测试** (`test/.../SkillExecutionGAgentTests.cs`)

glm-5.1 之前的 #6 也是对的。新加的 854 行测试只覆盖了 helper (`SendOutputAsync`, `TrySendFailureAsync`, `BuildExecutionMetadata`)。`SkillExecutionGAgentTests.cs:710` 那个 doc comment 自己写了:*"This avoids calling HandleStartAsync which would trigger…"* —— 所以是有意绕开。再加上 retry 路径和 completion 路径都没测，actor 的核心 lifecycle 是裸奔的。

definition 那边确实补了 `HandleExecutionCompletedAsync_WhenDefinitionDisabled_ShouldKeepCatalogStatusDisabled` 和 stale-revision guard 的测试，那个挺好的。但 execution 侧的 `HandleStart → execute → report` 这条 end-to-end 路径完全没有覆盖。

**修复:** 至少加一个测试: 用 stub LLM provider 跑 `HandleStartAsync` 的完整路径，断言 completion event 被持久化，并断言 `ReportSkillExecutionCompletedCommand` 被正确 dispatch。

### 多 model review 里我觉得没那么紧急的

**`IsStaleExecutionReport` 只看 revision 不看 execution_id** (codex #3) —— 严格说是真问题，但如果 #2 修了 (一个 definition 同一时间只有一个 in-flight execution)，光 revision 过滤就够用了。别过度设计，先修 #2。

**单一 completion 路径靠 `DispatchPort.DispatchAsync`** (codex #5) —— 取决于 `IActorDispatchPort.DispatchAsync` 是不是 at-least-once delivery。如果是 (大多数 actor framework 都是), 没有可靠性问题。值得确认一下契约，但没必要先加 retry 兜底。

**`ApplyExecutionDispatchFailed` 是 no-op reducer** (mimo #10) —— 其实没事。dispatch-failure event 是用来留 audit 记录的，catalog 更新走的是另一条 `UpdateRegistryExecutionErrorAsync` 路径 (line 211 那一带)。state 不需要跟踪它。可以加个短注释说明为什么是空的。

### 不是问题

测试 cross-ref 没更新 (kimi #12/#13)、XML doc 注释丢了 (mimo #14)、execution tests 里 `InitializeSkillDefinitionCommand` 没人用 (kimi #15)、canon doc 只改了一半 (codex #11) —— 都是真实存在但纯属洁癖类型的。值得做一次清理 sweep, 但不该挡 merge。

### 我的优先级

1. **必须改:** #1 (HandleEnable preserveExecutionState) 和 #2 (cron 排在 completion 之前的并发问题)。这两个都是相对老逻辑的用户可见 regression。
2. **应该改:** #3 (CI guard 接线) 和 #4 (`HandleStartAsync` smoke test)。
3. **后续 PR:** 文档 / 测试 cross-ref 清理、canon doc 更新。

两个 fix commit 干了实事 —— 完成态走 `SkillDefinitionGAgent.HandleExecutionCompletedAsync` 这条 routing 是对的, `ConfigRevision` 这套防 stale 的设计也很巧妙。但 cron 排程顺序和 enable 时的状态保留这两件事得在 land 之前修掉, 否则这个 PR 会引入用户可见的 regression。

Issue #462, opened 2026-04-27
Track remaining ownership/admission paths that still depend on read models or actor ID shape


louis4li:
## Context

Follow-up from issue #348 registry ownership review. The current registry command/query/admission port diff addresses GAgent registry ownership, but there are three adjacent ownership/admission paths that still rely on eventually consistent projection reads or actor ID shape. These should be tracked separately from the current registry ports patch.

### 1. Studio member bind command depends on StudioMember read model

`StudioMemberService.BindAsync` reads `IStudioMemberQueryPort` before sending the bind command. If a member was just created and the StudioMember read model lags, bind can fail with `STUDIO_MEMBER_NOT_FOUND` even though the authoritative member exists.

Path: `src/Aevatar.Studio.Application/Studio/Services/StudioMemberService.cs`

Expected direction: introduce an actor-owned command/admission contract for bind, or carry the strongly typed authoritative information returned by create into follow-up commands instead of depending on the projection read model.

### 2. Relay scope fallback uses projection for security-sensitive routing

When the relay JWT lacks a scope claim, `NyxIdRelayScopeResolver` looks up scope by API key through `IChannelBotRegistrationQueryByNyxIdentityPort`, which is projection-backed. Projection lag can make a newly provisioned bot callback return `401`, and stale/duplicate documents can affect tenant routing.

Path: `agents/Aevatar.GAgents.ChannelRuntime/NyxIdRelayScopeResolver.cs`

Expected direction: require relay tokens to carry a verifiable scope, or resolve through an authoritative ownership contract rather than a projection-backed lookup.

### 3. Relay conversation actor ID still encodes scope ownership

`BuildScopedRelayConversationActorId` hashes the scope into the conversation actor ID. Issue #348 calls scoped actor IDs a tactical workaround; actor IDs should remain opaque addresses, and scope ownership should come from actor-owned contract/read model/admission rather than ID format.

Path: `agents/Aevatar.GAgents.NyxidChat/NyxIdChatEndpoints.Relay.cs`

Expected direction: move scope ownership out of actor ID shape and into explicit ownership/admission semantics.

## Notes

These are architecture follow-ups outside the current registry command/query/admission port patch. They should not block the receipt semantics fix, but they should be handled before claiming #348-style ownership/admission cleanup is complete across all adjacent paths.

louis4li:
Closing this issue for now rather than continuing the current #506 direction.

Current status of the three original findings:

[...]

PR #506, opened 2026-04-29
Refactor issue 462 relay ownership admission paths

louis4li:
## Problem

Partially addresses #462.

Issue #462 tracks three adjacent ownership/admission problems:

- relay scope fallback used projection/read-model state for security-sensitive routing;
- relay conversation actor ids encoded scope ownership;
- StudioMember bind admission still depends on StudioMember read-model freshness.

This PR now intentionally covers only the two Relay items. The larger StudioMember binding lifecycle refactor is split out to #516 so it can get a separate design/review before implementation.

## Current PR Scope

- Require a verified canonical `scope_id` for NyxID relay admission instead of falling back to projection/read-model ownership inference.
- Carry the verified scope as typed relay activity context (`validated_scope_id`) so downstream handling can validate registration against the authoritative callback-token scope.
- Resolve NyxID bot registration with the verified scope instead of accepting an ambiguous API-key-only read-model match.
- Build relay conversation actor ids from relay identity + verified scope + canonical conversation key as an opaque hash partition. This prevents same-API-key cross-scope conversation state sharing while keeping ownership/admission decisions on typed verified scope and registration validation, not actor id parsing.
- Dispatch relay webhook turns through the standard actor dispatch port instead of calling the actor directly.
- Update focused tests and docs for the narrowed Relay admission/routing behavior.

## Explicitly Out of Scope

- StudioMember binding admission/lifecycle changes.
- Studio actor state/event/proto changes for binding runs.
- Projection continuation or binding-run actor implementation.
- Studio API/frontend binding workflow changes.

Those broader changes were intentionally reverted in `ff300e10` and are tracked separately in #516. The design document in this PR keeps the Studio direction as proposal context only; reviewers should not treat it as implemented behavior in PR #506.

## Architecture Notes

- Relay admission uses the callback token's verified scope claim as the authoritative ownership input.
- Relay actor ids remain opaque runtime addresses; verified scope is a hash partition input for isolation, not a parseable ownership fact.
- Relay downstream registration resolution validates that the API key belongs to the verified scope, avoiding ambiguous API-key-only routing.
- Projection remains a read-model materialization path in this PR; no business continuation is added to the projection runtime.
- Runtime/dispatch responsibilities stay separated for relay delivery: runtime handles lookup/topology and the dispatch port handles message delivery.

## Validation

- `dotnet test test/Aevatar.AI.Tests/Aevatar.AI.Tests.csproj --no-restore --filter "NyxIdChatEndpointsCoverageTests|NyxIdRelayAuthValidatorTests" --nologo` — 80 passed
- `dotnet test test/Aevatar.GAgents.ChannelRuntime.Tests/Aevatar.GAgents.ChannelRuntime.Tests.csproj --no-restore --filter "ChannelConversationTurnRunnerTests" --nologo` — 67 passed
- `bash tools/ci/test_stability_guards.sh` — passed
- `bash tools/ci/architecture_guards.sh` — passed
- `dotnet build aevatar.slnx --no-restore --nologo` — passed, 249 warnings, 0 errors
- `git diff --check` — passed

## Residual Risk

- `../NyxID` is not present in this workspace, so I could not locally verify the external NyxID callback-token scope-claim contract against the sibling repository. The implementation is based on repository-local relay contracts and tests.


louis4li:
Review follow-up pushed in `f3dd5ea5` and merge-updated in `9f680675`.

Addressed the requested architecture/code-review gaps:

- Relay conversation actor ids are now opaque hash-only addresses (`channel-conversation:relay:{hash}`), with the hash derived from relay identity + canonical conversation key; no plaintext canonical key or scope ownership is exposed in the actor id.
- Member bind frontend no longer fabricates legacy scope-binding status objects from the member binding view. It uses a typed `StudioMemberBindingView` and removes the old compatibility shim.
- Member bind UI now observes completion by polling the member read model for the accepted `bindingId` terminal run before refreshing the service catalog.
- StudioMember actor state now retains only the latest binding run, preventing unbounded growth of repeated run state with workflow payloads.
- Member bind frontend request bodies no longer send ignored `implementationKind` / `displayName` fields.

Validation after latest `origin/dev` merge:

- `pnpm tsc --noEmit`
- `pnpm jest src/shared/studio/api.test.ts src/pages/studio/components/bind/StudioMemberBindPanel.test.tsx src/pages/studio/index.test.tsx --runInBand`
- `dotnet test test/Aevatar.Studio.Tests/Aevatar.Studio.Tests.csproj --no-restore --filter "StudioMemberGAgentStateTests|StudioMemberBindingContinuationServiceTests|ActorDispatchStudioMemberCommandServiceTests" --nologo`
- `dotnet test test/Aevatar.AI.Tests/Aevatar.AI.Tests.csproj --no-restore --filter "NyxIdChatEndpointsCoverageTests|NyxIdRelayAuthValidatorTests" --nologo`
- `git diff --check`

louis4li:
Scope clarification after rollback:

PR #506 has been narrowed to the Relay ownership/admission fixes only. It no longer implements the StudioMember binding lifecycle/projection-continuation changes; those were reverted in `ff300e10` because the architecture impact is large enough to require a separate design and review.

The remaining StudioMember bind admission problem from #462 is now tracked separately in #516. Please review this PR as:

- in scope: Relay verified `scope_id`, typed `validated_scope_id`, verified-scope registration resolution, opaque relay actor id, dispatch-port delivery;
- out of scope: StudioMember binding run state/events, projection continuations, binding-run actors, Studio API/frontend workflow changes.

The PR description has been updated with the same scope note to avoid treating the design proposal as implemented behavior.

louis4li:
Follow-up cleanup: I trimmed the history design note in `a6623a68` from a detailed StudioMember implementation proposal down to a scoped Relay admission note.

The doc now only records:

- #462's three original findings;
- this PR's Relay-only implementation scope;
- explicit StudioMember out-of-scope boundaries;
- the architecture decisions for verified scope, opaque relay actor id, verified-scope registration matching, and dispatch-port delivery;
- #516 as the follow-up for StudioMember binding.

This should reduce review noise and avoid implying that the StudioMember async binding design is part of PR #506.

agents/Aevatar.GAgents.NyxidChat/NyxIdChatEndpoints.Relay.cs
```diff
@@ -183,67 +191,24 @@ private static long ResolveReplyTokenExpiresAtUnixMs(
...
-    private static string BuildScopedRelayConversationActorId(string? scopeId, string canonicalKey)
+    private static string BuildRelayConversationActorId(string relayIdentity, string canonicalKey)
     {
-        ArgumentException.ThrowIfNullOrWhiteSpace(scopeId);
+        ArgumentException.ThrowIfNullOrWhiteSpace(relayIdentity);
         ArgumentException.ThrowIfNullOrWhiteSpace(canonicalKey);
 
-        var scopeHash = Convert.ToHexString(SHA256.HashData(Encoding.UTF8.GetBytes(scopeId.Trim())))
+        var actorKey = $"{relayIdentity.Trim()}\n{canonicalKey.Trim()}";
+        var relayHash = Convert.ToHexString(SHA256.HashData(Encoding.UTF8.GetBytes(actorKey)))
```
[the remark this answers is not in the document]
louis4li: Addressed in `aeb467ba`: the relay conversation actor hash now includes `relayIdentity + validatedScopeId + canonicalKey`, so the same Nyx API key and same platform conversation cannot share one `ConversationGAgent` across scopes. The scope remains a typed `validated_scope_id` admission/registration fact; the actor id is still an opaque hash partition and is not parsed as ownership. Added a regression test for same Nyx key + same conversation + different scopes producing distinct actors, and updated the docs/PR description accordingly.

agents/Aevatar.GAgents.NyxidChat/NyxIdChatEndpoints.Relay.cs
```diff
@@ -67,12 +68,7 @@ private static async Task<IResult> HandleRelayWebhookAsync(
             }
 
             http.User = validation.Principal;
-            var scopeId = await ResolveRelayScopeIdAsync(
-                validation.ScopeId,
-                payload,
-                http.RequestServices,
-                logger,
-                ct);
+            var scopeId = ResolveRelayScopeId(validation.ScopeId);
```
eanzhao, line 71: **[codex]** severity=blocker, category=bug

This now makes `scope_id` mandatory for admission, but the current local NyxID source does not emit that claim in relay callback tokens. In `/Users/chronoai/Code/NyxID/backend/src/crypto/jwt.rs`, `RelayCallbackClaims` / `generate_relay_callback_token` only include `iss/aud/exp/iat/jti/token_type/api_key_id/message_id/platform/body_sha256`, so `validation.ScopeId` will be null for real callbacks and this endpoint will return `401` for all existing relay traffic. Please update and verify the NyxID callback-token contract first, and add a fixture/integration test using the real token shape before removing the fallback.

agents/Aevatar.GAgents.NyxidChat/NyxIdChatEndpoints.Relay.cs
```diff
@@ -183,67 +191,28 @@ private static long ResolveReplyTokenExpiresAtUnixMs(
...
 
-        var scopeHash = Convert.ToHexString(SHA256.HashData(Encoding.UTF8.GetBytes(scopeId.Trim())))
+        var actorKey = $"{relayIdentity.Trim()}\n{validatedScopeId.Trim()}\n{canonicalKey.Trim()}";
+        var relayHash = Convert.ToHexString(SHA256.HashData(Encoding.UTF8.GetBytes(actorKey)))
             .ToLowerInvariant();
-        return $"{ConversationGAgent.BuildActorId(canonicalKey)}:scope:{scopeHash}";
+        return $"channel-conversation:relay:{relayHash}";
     }
 
     private static string? NormalizeOptional(string? value)
```
eanzhao, line 218: **[v4-pro]** severity=minor, category=design

`BuildRelayConversationActorId` concatenates raw inputs with `\n` as delimiter before hashing. If any of `relayIdentity`, `validatedScopeId`, or `canonicalKey` contained an internal newline (unlikely in practice for UUID-like identifiers), two different input triplets could produce the same `actorKey` → hash collision. Consider using a length-prefixed or fixed-separator encoding for robustness.

_(Reviewer originally pointed to line 219; re-anchored to line 218 within diff hunk.)_

agents/Aevatar.GAgents.NyxidChat/NyxIdChatEndpoints.Relay.cs
```diff
@@ -122,33 +129,34 @@ private static async Task<IResult> HandleRelayWebhookAsync(
                 CorrelationId = activity.OutboundDelivery.CorrelationId,
             };
 
-            var actorId = BuildScopedRelayConversationActorId(scopeId, activity.Conversation.CanonicalKey);
-            var actor = await actorRuntime.CreateAsync<ConversationGAgent>(actorId, ct);
+            var actorId = BuildRelayConversationActorId(relayIdentity, scopeId, activity.Conversation.CanonicalKey);
+            var actor = await actorRuntime.GetAsync(actorId)
+                ?? await actorRuntime.CreateAsync<ConversationGAgent>(actorId, ct);
```
eanzhao, line 134: **[glm-5.1]** severity=major, category=design

**Breaking actor ID format with no migration path.** The format changes from `channel-conversation:{canonicalKey}:scope:{hash}` to `channel-conversation:relay:{hash}`. Any existing relay `ConversationGAgent` actors in production will become unreachable after this deploys — the new code generates different IDs and will never find the old actors, silently orphaning them.

This needs at least one of:
1. A documented migration/rollback plan (even if the answer is "relay conversations are short-lived and acceptable to lose").
2. A deployment-order note that existing in-flight relay turns will create new actors.
3. A startup compatibility layer that can resolve old-format IDs during a transition window.

The design doc (`2026-04-28-issue-462-ownership-admission-design.md`) acknowledges the isolation rationale but does not address state continuity for existing actors.

eanzhao, line 134: **[glm-5.1]** severity=minor, category=design

**TOCTOU in get-or-create.**
```csharp
var actor = await actorRuntime.GetAsync(actorId)
    ?? await actorRuntime.CreateAsync<ConversationGAgent>(actorId, ct);
```
Two concurrent webhook deliveries for the same relay identity+scope+conversation will both see `null` from `GetAsync` and both call `CreateAsync`. This is safe only if `CreateAsync` is idempotent (returns the existing actor on duplicate). If it isn't documented as such, consider a single `GetOrCreateAsync` primitive, or add a brief comment confirming the runtime guarantees idempotent creation.

agents/Aevatar.GAgents.NyxidChat/ChannelConversationTurnRunner.cs
```diff
@@ -622,9 +622,18 @@ private async Task<ConversationTurnResult> SendRelayTextFallbackAsync(
...
+        var validatedScopeId = NormalizeOptional(activity.TransportExtras?.ValidatedScopeId);
         if (!string.IsNullOrWhiteSpace(nyxAgentApiKeyId) &&
             _registrationQueryByNyxIdentityPort is not null)
         {
+            if (!string.IsNullOrWhiteSpace(validatedScopeId))
+            {
+                return await ResolveRegistrationByVerifiedScopeAsync(
+                    nyxAgentApiKeyId,
+                    validatedScopeId,
+                    ct);
```
eanzhao, line 634: **[glm-5.1]** severity=minor, category=design

**Client-side scope filtering after full list fetch.** `ResolveRegistrationByVerifiedScopeAsync` fetches all registrations for the API key via `ListByNyxAgentApiKeyIdAsync` and then filters in-memory. This is correct but could be pushed to the query layer (e.g., `GetByNyxAgentApiKeyIdAndScopeAsync`) to avoid transferring stale/unrelated registrations. Low priority since a single API key typically has few registrations, but worth noting if the data model grows.

test/Aevatar.AI.Tests/NyxIdChatEndpointsCoverageTests.cs
```diff
@@ -1993,6 +2115,7 @@ private static async Task<IResult> InvokeResultAsync(string methodName, params o
     {
         EnsureEndpointContextServices(args);
```
eanzhao, line 2116: **[glm-5.1]** severity=minor, category=test

**`NormalizeEndpointArguments` auto-inserts `ForwardingActorDispatchPort` silently.** Tests that forget to pass an explicit dispatch port get a forwarding stub that silently drops dispatches to non-existent actors. This masked a potential failure mode: the partition test (`ShouldPartitionRelayActorByVerifiedScope_WhenNyxKeyIsShared`) doesn't verify that envelopes actually reach the actors — it only checks actor IDs. Consider either (a) always requiring an explicit dispatch port in relay tests, or (b) having `ForwardingActorDispatchPort.DispatchAsync` throw/log when the target actor doesn't exist, so silent drops surface as test failures.

agents/Aevatar.GAgents.NyxidChat/NyxIdChatEndpoints.Relay.cs
```diff
@@ -82,6 +78,16 @@ private static async Task<IResult> HandleRelayWebhookAsync(
                 return Results.Unauthorized();
             }
 
+            var relayIdentity = NormalizeOptional(validation.RelayApiKeyId);
+            if (relayIdentity is null)
+            {
+                logger.LogWarning(
```
eanzhao, line 84: **[glm-5.1]** severity=nit, category=design

The `relayIdentity` variable holds `validation.RelayApiKeyId` (the Nyx API key ID from the JWT). The name `relayIdentity` is fine conceptually but could be confused with a relay *session* identity. A brief inline comment like `// relayApiKeyId serves as the relay identity for actor partitioning` would help future readers.

louis4li:
Closing this PR.

After review, the core premise of this branch is not valid for the current system boundary: NyxID callback tokens do not emit an Aevatar `scope_id`, and NyxID should not be coupled to Aevatar's internal scope ownership model. Because of that, the scope-claim based relay admission path in this PR would reject real relay traffic and should not be merged.

The only independently useful part is the dispatch-port cleanup, but that can be reintroduced later as a small standalone change if needed. The broader relay ownership/admission redesign should be revisited only after defining an Aevatar-owned authoritative admission contract for Nyx relay API key ownership.

PR #509, opened 2026-04-29
fix: deduplicate /daily auth card, pre-fill username, honor bot-owner LLM config

eanzhao:
## Summary

Three independent issues on the Lark `/daily` flow, each surfacing as a confusing UX problem for bot owners:

### 1. Duplicate "GitHub authorization required" block
`BuildDailyReportCredentialsCard` set both `content.Cards[0].Text` (card body) and `content.Text` (intended as non-card fallback) to the same auth content. Lark's form-mode composer (`LarkMessageComposer.BuildLeadingMarkdown`) concatenates `content.Text` with the first card body into a single rendered markdown element, so users saw the block twice. Card body is now the single source of truth; `content.Text` stays empty.

### 2. Submitted github_username dropped on auth re-prompt
`BuildDailyReportCredentialsCard` rebuilt the form with `preferredGithubUsername: null`, so the username the user typed (`/daily eanzhao`) was discarded and the field reverted to the `octocat` placeholder. `CreateDailyReportAgentAsync` now passes the resolved username into `BuildGitHubAuthorizationResponseAsync`, the auth-required JSON echoes it back as `github_username`, and the card threads it into the form's `default_value`.

### 3. SkillRunner / WorkflowAgent ignored bot owner's LLM route
Reported as `Upstream LLM route 'nyxid' rejected the request with HTTP 400 for model 'gpt-5.4'. Provider 'openai' not connected.` even though the bot owner had a working `chrono-llm` service in NyxID and a UserConfig with `defaultModel = gpt-5.5` / `preferredLlmRoute = /api/v1/proxy/s/chrono-llm`.

Root cause: `ChannelLlmReplyInboxRuntime` (nyxid-chat) reads UserConfig and pins `ModelOverride` / `NyxIdRoutePreference` / `MaxToolRoundsOverride` on outbound LLM metadata, but `SkillRunnerGAgent` and `WorkflowAgentGAgent` built their own metadata from `State.OutboundConfig` only. Without those keys, `NyxIdLLMProvider` fell through to compile-time defaults (`gpt-5.4` against `/api/v1/llm/gateway/v1/`), and the gateway routed `gpt-*` to OpenAI — which the bot owner hadn't connected.

Both agents now resolve `IUserConfigQueryPort` via `Services.GetService<>` and apply the same three metadata keys nyxid-chat applies. Empty fields and projection failures fall through to provider defaults silently, preserving existing test fixtures and no-projection hosts.

## NyxID-side verification

Used `nyxid` CLI on the affected account:

- `nyxid service list` → `api-github` shows `pending_auth`; `chrono-llm` shows `active`.
- `GET /api/v1/providers/my-tokens` → `{"tokens":[]}` (so the GitHub OAuth check correctly returns "not connected").
- `GET /api/v1/llm/status` → all providers `not_connected` (so any `gpt-*` model routed to the gateway 400s).
- `GET /api/user-config` → `defaultModel=gpt-5.5, preferredLlmRoute=/api/v1/proxy/s/chrono-llm`.

So the LLM was correctly configured at NyxID; the skill runner just wasn't reading it.

## Test plan
- [x] `dotnet test test/Aevatar.GAgents.ChannelRuntime.Tests/Aevatar.GAgents.ChannelRuntime.Tests.csproj` — 713 passing
- [x] New `BuildExecutionMetadata_ShouldApplyOwnerUserConfigOverrides_*` covers ScopeId-based lookup, missing port fallback, empty-fields fallback, and throwing-port fallback
- [x] New `FormatDailyReportToolReply_*` tests cover oauth_required + credentials_required + username pre-fill
- [x] Updated existing tool tests to assert `github_username` is echoed
- [x] Updated `NyxRelayAgentBuilderFlowTests.FormatToolResult_ShouldReturnCardForm_WhenCredentialsRequired` to require empty `Text` (the old assertion codified the duplicate-render bug)
- [ ] Manual sanity check on Lark with a real bot post-deploy


agents/Aevatar.GAgents.Scheduled/Aevatar.GAgents.Scheduled.csproj
```diff
@@ -24,6 +24,7 @@
     <ProjectReference Include="..\..\src\Aevatar.CQRS.Projection.Providers.InMemory\Aevatar.CQRS.Projection.Providers.InMemory.csproj" />
     <ProjectReference Include="..\..\src\Aevatar.Foundation.Abstractions\Aevatar.Foundation.Abstractions.csproj" />
     <ProjectReference Include="..\..\src\Aevatar.Foundation.Core\Aevatar.Foundation.Core.csproj" />
+    <ProjectReference Include="..\..\src\Aevatar.Studio.Application\Aevatar.Studio.Application.csproj" />
```
eanzhao: **[Consensus: 4 models]** severity=major, category=design

`Aevatar.GAgents.Scheduled` now project-references `Aevatar.Studio.Application` solely to consume `IUserConfigQueryPort`. All four models flag this as a cross-layer dependency from the agents layer into the full Studio application package. The runtime fallback via `Services.GetService<>()` mitigates the optionality concern but doesn't fix the compile-time direction. Recommendation: lift `IUserConfigQueryPort` (and any DTOs it touches) into a smaller abstractions package — `Aevatar.Studio.Application.Studio.Abstractions` already exists in path naming — so the agents layer can depend on the contract without pulling the application implementation.

<details>

- **codex** (major): `Aevatar.GAgents.Scheduled` now references the concrete `Aevatar.Studio.Application` project just to optionally query `IUserConfigQueryPort`. That couples the scheduled agent package to the Studio application layer and makes the fallback behavior still require the Studio assembly at compile/deploy time. Move `IUserConfigQueryPort`/`UserConfig` into an abstractions/contracts project already appropriate for this dependency direction, or introduce a narrow scheduled-agent abstraction that the host can bind to the Studio implementation.
- **v4-pro** (minor): New dependency on `Aevatar.Studio.Application` pulls the Studio Application layer into the GAgents.Scheduled project. If `IUserConfigQueryPort` is the only boundary needed, consider defining the port in a lighter-weight abstractions package (`Aevatar.Studio.Application.Abstractions` or similar) to avoid transitively depending on the full Application layer implementation.
- **glm-5.1** (minor): Adding a project reference to `Aevatar.Studio.Application` introduces a dependency from the agents layer to the Studio application layer. `IUserConfigQueryPort` lives under `Aevatar.Studio.Application.Studio.Abstractions` — if this interface were in its own abstractions package (or already in `Aevatar.Foundation.Abstractions`), the agents layer wouldn't need to reach into Studio's application package. The `Services.GetService<>()` fallback mitigates the risk at runtime, but the compile-time dependency direction is still worth cleaning up to avoid future coupling.
- **mimo-v2.5-pro** (major): Adding `Aevatar.Studio.Application` as a dependency from `Aevatar.GAgents.Scheduled` introduces a cross-layer dependency (Scheduled → Application). This is likely needed for `IUserConfigQueryPort`, but consider whether this violates the project's layering rules. A cleaner approach would be to define the interface in an abstractions package that both can reference, or use a mediator/event pattern to avoid the direct dependency.
</details>

agents/Aevatar.GAgents.Scheduled/SkillRunnerGAgent.cs
```diff
@@ -694,9 +694,52 @@ private IReadOnlyDictionary<string, string> BuildExecutionMetadata()
...
+        // skill runner falls through to NyxIdLLMProvider's compile-time defaults
+        // (`gpt-5.4` against `/api/v1/llm/gateway/v1/`), which the gateway routes to the
+        // OpenAI provider. Bot owners running their own LLM through a custom NyxID service
+        // (e.g. `chrono-llm` at `/api/v1/proxy/s/chrono-llm`) would see HTTP 400
+        // "Provider 'openai' not connected" because their gateway has no OpenAI grant.
+        await ApplyOwnerLlmConfigAsync(metadata, ct);
         return metadata;
     }
 
+    private async Task ApplyOwnerLlmConfigAsync(IDictionary<string, string> metadata, CancellationToken ct)
```
eanzhao: **[Consensus: 4 models]** severity=major, category=design

`ApplyOwnerLlmConfigAsync` is duplicated nearly verbatim between `SkillRunnerGAgent` (this file, L~700–745) and `WorkflowAgentGAgent` (L~210–256). The two copies differ only in the `"Skill runner"` vs `"Workflow agent"` log prefix. Future changes to scope lookup, the metadata-key list, or the swallow-everything `catch` will have to be made twice and are likely to drift. Extract into a shared helper — e.g. an `internal static OwnerLlmConfigApplier.ApplyAsync(metadata, scopeId, services, logger, actorLabel, ct)` — and have both agents call it.

<details>

- **v4-pro** (major): `ApplyOwnerLlmConfigAsync` is duplicated verbatim (except for the log template's actor-type name) between `SkillRunnerGAgent` and `WorkflowAgentGAgent`. Extract this into a shared helper or extension method (e.g., `OwnerLlmConfigApplier`) so future metadata-key additions only need one edit. The `Services?.GetService<IUserConfigQueryPort>()` service-locator call also propagates in two places; centralizing it would isolate that pattern.
- **k2p6** (major): `ApplyOwnerLlmConfigAsync` is duplicated almost verbatim in `WorkflowAgentGAgent.cs` (only the log-message prefix differs). Extract this to a shared helper—e.g. a static `LlmConfigApplier.ApplyAsync` or a common base-class method—so the two copies cannot drift on future changes.
- **glm-5.1** (major): `ApplyOwnerLlmConfigAsync` is duplicated character-for-character in `WorkflowAgentGAgent.cs` (~40 lines, same logic, same error handling, same log template). Extract this into a shared base class or a static utility (e.g. `OwnerLlmConfigResolver.ApplyAsync(metadata, scopeId, services, logger, id, ct)`) so the two agents don't drift apart on future metadata key additions.

If the agents already share a common base, put it there; otherwise a small `internal static` helper in the Scheduled project keeps the scope tight.
- **mimo-v2.5-pro** (major): The `ApplyOwnerLlmConfigAsync` method is duplicated identically in both `SkillRunnerGAgent` and `WorkflowAgentGAgent`. Extract this into a shared helper (e.g., a static utility method or an extension method on `IReadOnlyDictionary<string, string>`) to avoid the maintenance risk of future drift if the logic needs to change.
- **v4-pro** sibling at `WorkflowAgentGAgent.cs:219`: See `SkillRunnerGAgent` comment — this is the same `ApplyOwnerLlmConfigAsync` copy. The two differ only in `"Skill runner"` vs `"Workflow agent"` in the log template; inject the actor-type label instead of duplicating the whole method.
</details>

agents/Aevatar.GAgents.Scheduled/SkillRunnerGAgent.cs
```diff
@@ -694,9 +694,52 @@ private IReadOnlyDictionary<string, string> BuildExecutionMetadata()
...
+        if (string.IsNullOrWhiteSpace(State.ScopeId))
+            return;
+
+        var queryPort = Services?.GetService<Aevatar.Studio.Application.Studio.Abstractions.IUserConfigQueryPort>();
+        if (queryPort is null)
+            return;
+
+        try
+        {
+            var config = await queryPort.GetAsync(State.ScopeId, ct);
```
eanzhao: **[k2p6]** severity=major, category=bug

`queryPort.GetAsync` may return `null`; dereferencing `config.DefaultModel` will throw `NullReferenceException` that is only incidentally swallowed by the generic `catch (Exception)` block. Add an explicit `if (config is null) return;` guard immediately after the `await` before accessing members.

agents/Aevatar.GAgents.Scheduled/SkillRunnerGAgent.cs
```diff
@@ -694,9 +694,52 @@ private IReadOnlyDictionary<string, string> BuildExecutionMetadata()
...
         if (!string.IsNullOrWhiteSpace(State.ScopeId))
             metadata["scope_id"] = State.ScopeId;
+
+        // Apply the bot owner's UserConfig (DefaultModel + PreferredLlmRoute + MaxToolRounds)
+        // exactly like ChannelLlmReplyInboxRuntime does for nyxid-chat. Without this, the
+        // skill runner falls through to NyxIdLLMProvider's compile-time defaults
+        // (`gpt-5.4` against `/api/v1/llm/gateway/v1/`), which the gateway routes to the
+        // OpenAI provider. Bot owners running their own LLM through a custom NyxID service
+        // (e.g. `chrono-llm` at `/api/v1/proxy/s/chrono-llm`) would see HTTP 400
+        // "Provider 'openai' not connected" because their gateway has no OpenAI grant.
```
eanzhao: **[mimo-v2.5-pro]** severity=minor, category=design

Resolving `IUserConfigQueryPort` via `Services.GetService<>()` on every execution is functional but not idiomatic .NET dependency injection. Consider injecting it via the constructor (making it nullable/optional) so it's resolved once at activation time and is more explicit about the dependency surface.

> Same concern applies to `WorkflowAgentGAgent.cs:228` (mimo flagged it there too, identical body).

agents/Aevatar.GAgents.Scheduled/WorkflowAgentGAgent.cs
```diff
@@ -206,9 +206,51 @@ private IReadOnlyDictionary<string, string> BuildExecutionMetadata()
             metadata[ChannelMetadataKeys.LarkReceiveIdType] = State.LarkReceiveIdType;
         if (!string.IsNullOrWhiteSpace(State.NyxProviderSlug))
             metadata[ChannelMetadataKeys.LarkOutboundProxySlug] = State.NyxProviderSlug;
+
+        // Mirror SkillRunnerGAgent.ApplyOwnerLlmConfigAsync so workflow-backed agents (e.g.
+        // social_media) honor the bot owner's UserConfig (DefaultModel + PreferredLlmRoute +
+        // MaxToolRounds) the same way nyxid-chat does. Without this, the workflow's LLM steps
+        // fall through to NyxIdLLMProvider's compile-time default route + model and 400 when
+        // the bot owner pre-configured a custom NyxID service like `chrono-llm` instead of
+        // OpenAI.
```
eanzhao: **[codex]** severity=minor, category=test

The workflow-backed path is part of the reported regression, but the added tests only exercise `SkillRunnerGAgent.BuildExecutionMetadataAsync`. Add an equivalent `WorkflowAgentGAgent` metadata test covering scope lookup and the three override keys so this path cannot silently diverge from `SkillRunnerGAgent` again.

agents/Aevatar.GAgents.Scheduled/Aevatar.GAgents.Scheduled.csproj
eanzhao: Fixed in af5e4d3. The Scheduled package no longer references Studio.Application. Introduced `IOwnerLlmConfigSource` (narrow port) + `OwnerLlmConfig` record inside Scheduled itself, and added a small adapter `StudioUserConfigOwnerLlmConfigSource` in the Mainnet host that bridges Studio's `IUserConfigQueryPort` onto the narrow port. The host stays the only layer that compiles against both sides — agents see only the narrow contract.

agents/Aevatar.GAgents.Scheduled/SkillRunnerGAgent.cs
eanzhao: Fixed in af5e4d3. Extracted to a shared `OwnerLlmConfigApplier.ApplyAsync(metadata, scopeId, source, logger, actorLabel, actorId, ct)` helper in the Scheduled package; both `SkillRunnerGAgent` and `WorkflowAgentGAgent` now delegate. While here I also routed `ChannelConversationTurnRunner` (NyxidChat) through the same helper so the channel-bot LLM path can't drift either. Future metadata-key changes are now a one-site edit.

agents/Aevatar.GAgents.Scheduled/SkillRunnerGAgent.cs
eanzhao: Fixed in af5e4d3. The applier coalesces a possibly-null source result into `OwnerLlmConfig.Empty` before reading fields (`config = await source.GetForScopeAsync(...) ?? OwnerLlmConfig.Empty`), so a null config can no longer trigger an NRE that the catch-all would only incidentally swallow.

agents/Aevatar.GAgents.Scheduled/SkillRunnerGAgent.cs
eanzhao: Fixed in af5e4d3. Both agents now accept `IOwnerLlmConfigSource? ownerLlmConfigSource = null` via the constructor (mirroring the existing `NyxIdApiClient?` pattern). The previous `Services.GetService<>` call remains as a fallback only when the ctor param is null, so production DI binds it once at activation time and tests can inject a stub directly.

agents/Aevatar.GAgents.Scheduled/WorkflowAgentGAgent.cs
eanzhao: Fixed in af5e4d3. Added `HandleTriggerAsync_ShouldPinOwnerLlmConfigOverridesOnDispatchedMetadata` and `HandleTriggerAsync_ShouldOmitOwnerLlmOverrides_WhenSourceIsAbsent` to `WorkflowAgentGAgentTests`, mirroring the SkillRunner coverage. Also added matching tests for `ChannelConversationTurnRunner` since that path now goes through the same shared helper.

src/Aevatar.Mainnet.Host.Api/Hosting/StudioUserConfigOwnerLlmConfigSource.cs
```diff
@@ -0,0 +1,32 @@
...
+
+    public async Task<OwnerLlmConfig> GetForScopeAsync(string scopeId, CancellationToken ct = default)
+    {
+        var config = await _queryPort.GetAsync(scopeId, ct);
+        if (config is null)
+            return OwnerLlmConfig.Empty;
+
+        return new OwnerLlmConfig(
+            DefaultModel: config.DefaultModel,
+            PreferredLlmRoute: config.PreferredLlmRoute,
```
eanzhao: **[codex]** severity=minor, category=bug

`ProjectionUserConfigQueryPort.GetAsync` fills `PreferredLlmRoute` with `UserConfigLlmRouteDefaults.Gateway` when the user has no saved route, so this bridge will cause `OwnerLlmConfigApplier` to pin `NyxIdRoutePreference` for otherwise unconfigured owners instead of falling through to provider defaults as the new port contract says. Either return `null`/empty for the defaulted route here, or use a query shape that can distinguish an explicitly saved route from the projection default.

agents/Aevatar.GAgents.Scheduled/IOwnerLlmConfigSource.cs
```diff
@@ -0,0 +1,35 @@
+namespace Aevatar.GAgents.Scheduled;
```
eanzhao: **[glm-5.1]** severity=minor, category=design

`IOwnerLlmConfigSource` and `OwnerLlmConfigApplier` are consumed by `NyxidChat` (`ChannelConversationTurnRunner`), `SkillRunnerGAgent`, and `WorkflowAgentGAgent`. Housing them in the `Scheduled` package creates a dependency from NyxidChat → Scheduled that may surprise future maintainers. Consider moving both types to a shared abstractions or foundation package that both Scheduled and NyxidChat already reference, so the ownership boundary is clearer.

src/Aevatar.Mainnet.Host.Api/Hosting/StudioUserConfigOwnerLlmConfigSource.cs
eanzhao: Fixed in e958e7d. Hardened `StudioUserConfigOwnerLlmConfigSource` to run `config.PreferredLlmRoute` through `UserConfigLlmRoute.Normalize` and pass null when it collapses to the gateway sentinel. Today the sentinel is empty-string so the applier's null-or-whitespace guard already handled this case, but the explicit normalization makes the contract future-proof against any change to `UserConfigLlmRouteDefaults.Gateway`. Added 10 tests in `StudioUserConfigOwnerLlmConfigSourceTests` covering empty / whitespace / `auto` / `gateway` / invalid-URI / explicit-route / bare-slug expansion / null-config paths.

agents/Aevatar.GAgents.Scheduled/IOwnerLlmConfigSource.cs
eanzhao: Fixed in e958e7d. Lifted both types out of `Aevatar.GAgents.Scheduled` and into the shared AI layer: `IOwnerLlmConfigSource` + `OwnerLlmConfig` now live in `Aevatar.AI.Abstractions.LLMProviders` (pure contract) and `OwnerLlmConfigApplier` lives in `Aevatar.AI.Core.LLMProviders` (where the `ILogger` dep belongs). Both Scheduled and NyxidChat already reference these projects, so the move is a wash on dependency surface but eliminates the surprising NyxidChat → Scheduled coupling. All consumers updated; behavior unchanged.

eanzhao:
## Multi-model fix-check on `e958e7d`

Re-ran the same 5 reviewers (codex, deepseek-v4-pro, kimi-k2p6, glm-5.1, mimo-v2.5-pro) against the latest commit and asked each to classify whether prior inline comments are addressed.

| # | comment | verdict | N/M | summary |
|---|---|---|---|---|
| 1 | [#3159046881](https://github.com/aevatarAI/aevatar/pull/509#discussion_r3159046881) `Aevatar.GAgents.Scheduled.csproj:27` | ✅ resolved | 5/5 | The scheduled agent code no longer consumes Studio.Application types directly; it depends on IOwnerLlmConfigSource from Aevatar.AI.Abstractions and OwnerLlmConfigApplier from Aevatar.AI.Core. |
| 2 | [#3159046953](https://github.com/aevatarAI/aevatar/pull/509#discussion_r3159046953) `SkillRunnerGAgent.cs:709` | ✅ resolved | 5/5 | The duplicated ApplyOwnerLlmConfigAsync logic was extracted into the shared OwnerLlmConfigApplier.ApplyAsync helper. |
| 3 | [#3159047022](https://github.com/aevatarAI/aevatar/pull/509#discussion_r3159047022) `SkillRunnerGAgent.cs:720` | ✅ resolved | 5/5 | The direct queryPort.GetAsync dereference path was removed from the agents. |
| 4 | [#3159047120](https://github.com/aevatarAI/aevatar/pull/509#discussion_r3159047120) `SkillRunnerGAgent.cs:704` | ✅ resolved | 4/5 · codex→partial | SkillRunnerGAgent and WorkflowAgentGAgent now accept IOwnerLlmConfigSource via constructor injection as the primary path; Services.GetService is kept only as a… |
| 5 | [#3159047179](https://github.com/aevatarAI/aevatar/pull/509#discussion_r3159047179) `WorkflowAgentGAgent.cs:215` | ✅ resolved | 5/5 | WorkflowAgentGAgentTests now includes HandleTriggerAsync_ShouldPinOwnerLlmConfigOverridesOnDispatchedMetadata, which verifies scope lookup and all three override metadata keys. |
| 6 | [#3159289613](https://github.com/aevatarAI/aevatar/pull/509#discussion_r3159289613) `StudioUserConfigOwnerLlmConfigSource.cs:29` | ✅ resolved | 5/5 | StudioUserConfigOwnerLlmConfigSource now normalizes config.PreferredLlmRoute through UserConfigLlmRoute.Normalize before returning it. |
| 7 | [#3159289743](https://github.com/aevatarAI/aevatar/pull/509#discussion_r3159289743) `IOwnerLlmConfigSource.cs:1` | ✅ resolved | 5/5 | IOwnerLlmConfigSource was moved out of the Scheduled package into Aevatar.AI.Abstractions, and OwnerLlmConfigApplier was moved into Aevatar.AI.Core. |

**Totals:** 7 resolved

### Dissents worth surfacing

- **#3159047120** (SkillRunnerGAgent.cs:704) — `codex` (high) → **partial**: SkillRunnerGAgent and WorkflowAgentGAgent now accept an optional IOwnerLlmConfigSource through constructors, which addresses the explicit dependency surface. However, both still fall back to Services?.GetService<IOwnerLlmConfigSource>() inside metadata construction when the constructor field is null, so the per-execution service-locator pattern remains as a fallback.

---

_Generated by `/opencode-pr-fix-check`. This bot does not auto-resolve threads — close them manually if you agree with the verdicts._

eanzhao:
## Multi-model fix-check on `e958e7d`

Re-ran the same 5 reviewers (codex, deepseek-v4-pro, kimi-k2p6, glm-5.1, mimo-v2.5-pro) against the latest commit and asked each to classify whether prior inline comments are addressed.

| # | comment | verdict | N/M | summary |
|---|---|---|---|---|
| 1 | [#3159046881](https://github.com/aevatarAI/aevatar/pull/509#discussion_r3159046881) `Aevatar.GAgents.Scheduled.csproj:27` | ✅ resolved | 5/5 | The scheduled agent code no longer consumes Studio.Application types directly; it depends on IOwnerLlmConfigSource from Aevatar.AI.Abstractions and OwnerLlmConfigApplier from Aevatar.AI.Core. |
| 2 | [#3159046953](https://github.com/aevatarAI/aevatar/pull/509#discussion_r3159046953) `SkillRunnerGAgent.cs:709` | ✅ resolved | 5/5 | The duplicated ApplyOwnerLlmConfigAsync logic was extracted into the shared OwnerLlmConfigApplier.ApplyAsync helper. |
| 3 | [#3159047022](https://github.com/aevatarAI/aevatar/pull/509#discussion_r3159047022) `SkillRunnerGAgent.cs:720` | ✅ resolved | 5/5 | The direct queryPort.GetAsync dereference path was removed from the agents. |
| 4 | [#3159047120](https://github.com/aevatarAI/aevatar/pull/509#discussion_r3159047120) `SkillRunnerGAgent.cs:704` | ✅ resolved | 4/5 · codex→partial | SkillRunnerGAgent and WorkflowAgentGAgent now accept IOwnerLlmConfigSource via constructor injection as the primary path; Services.GetService is kept only as a… |
| 5 | [#3159047179](https://github.com/aevatarAI/aevatar/pull/509#discussion_r3159047179) `WorkflowAgentGAgent.cs:215` | ✅ resolved | 5/5 | WorkflowAgentGAgentTests now includes HandleTriggerAsync_ShouldPinOwnerLlmConfigOverridesOnDispatchedMetadata, which verifies scope lookup and all three override metadata keys. |
| 6 | [#3159289613](https://github.com/aevatarAI/aevatar/pull/509#discussion_r3159289613) `StudioUserConfigOwnerLlmConfigSource.cs:29` | ✅ resolved | 5/5 | StudioUserConfigOwnerLlmConfigSource now normalizes config.PreferredLlmRoute through UserConfigLlmRoute.Normalize before returning it. |
| 7 | [#3159289743](https://github.com/aevatarAI/aevatar/pull/509#discussion_r3159289743) `IOwnerLlmConfigSource.cs:1` | ✅ resolved | 5/5 | IOwnerLlmConfigSource was moved out of the Scheduled package into Aevatar.AI.Abstractions, and OwnerLlmConfigApplier was moved into Aevatar.AI.Core. |

**Totals:** 7 resolved

### Dissents worth surfacing

- **#3159047120** (SkillRunnerGAgent.cs:704) — `codex` (high) → **partial**: SkillRunnerGAgent and WorkflowAgentGAgent now accept an optional IOwnerLlmConfigSource through constructors, which addresses the explicit dependency surface. However, both still fall back to Services?.GetService<IOwnerLlmConfigSource>() inside metadata construction when the constructor field is null, so the per-execution service-locator pattern remains as a fallback.

---

_Generated by `/opencode-pr-fix-check`. This bot does not auto-resolve threads — close them manually if you agree with the verdicts._

agents/Aevatar.GAgents.Scheduled/SkillRunnerGAgent.cs
eanzhao: Pure ctor injection in 2e432c8. Verified that `LocalActorRuntime.CreateAgentInstance` uses `ActivatorUtilities.CreateInstance(_services, agentType)`, so DI fills the optional ctor param at activation time and the per-execution `Services.GetService<>` fallback was dead code in production. Dropped it from `SkillRunnerGAgent`, `WorkflowAgentGAgent`, and `ChannelConversationTurnRunner` (which also gained a ctor param + field, replacing its `_services.GetService<>()` per-turn lookup). The test factory `CreateRunner` now pulls the source out of the test-supplied service provider once and threads it through the ctor. 719 + 59 tests still pass.

eanzhao:
## Multi-model fix-check on `2e432c8` — all clear ✅

Re-ran the same 5 reviewers (codex, deepseek-v4-pro, kimi-k2p6, glm-5.1, mimo-v2.5-pro) against the latest commit. **All 7 prior inline comments are now unanimously resolved (5/5).** Codex's earlier dissent on the DI-pattern thread (#4) is gone — the `Services?.GetService<>()` fallback was removed and `IOwnerLlmConfigSource` is now constructor-injected at all three consumers.

| # | comment | verdict | N/M | summary |
|---|---|---|---|---|
| 1 | [#3159046881](https://github.com/aevatarAI/aevatar/pull/509#discussion_r3159046881) `Aevatar.GAgents.Scheduled.csproj:27` | ✅ resolved | 5/5 | The direct Scheduled -> Studio.Application dependency is no longer present in the current diff. |
| 2 | [#3159046953](https://github.com/aevatarAI/aevatar/pull/509#discussion_r3159046953) `SkillRunnerGAgent.cs:709` | ✅ resolved | 5/5 | The duplicated per-agent `ApplyOwnerLlmConfigAsync` logic has been extracted into `OwnerLlmConfigApplier.ApplyAsync` in `src/Aevatar.AI.Core/LLMProviders/OwnerLlmConfigApplier.cs`. |
| 3 | [#3159047022](https://github.com/aevatarAI/aevatar/pull/509#discussion_r3159047022) `SkillRunnerGAgent.cs:720` | ✅ resolved | 5/5 | The null case is explicitly handled in the new bridge: `StudioUserConfigOwnerLlmConfigSource.GetForScopeAsync` returns `OwnerLlmConfig.Empty` when `_queryPort.GetAsync(scopeId, ct)` returns null. |
| 4 | [#3159047120](https://github.com/aevatarAI/aevatar/pull/509#discussion_r3159047120) `SkillRunnerGAgent.cs:704` | ✅ resolved | 5/5 | The per-execution `Services.GetService<>()` pattern was removed from the production code path. |
| 5 | [#3159047179](https://github.com/aevatarAI/aevatar/pull/509#discussion_r3159047179) `WorkflowAgentGAgent.cs:215` | ✅ resolved | 5/5 | `WorkflowAgentGAgentTests` now includes `HandleTriggerAsync_ShouldPinOwnerLlmConfigOverridesOnDispatchedMetadata`, which verifies scope lookup and all three metadata override keys. |
| 6 | [#3159289613](https://github.com/aevatarAI/aevatar/pull/509#discussion_r3159289613) `StudioUserConfigOwnerLlmConfigSource.cs:29` | ✅ resolved | 5/5 | `StudioUserConfigOwnerLlmConfigSource` now normalizes `config.PreferredLlmRoute` through `UserConfigLlmRoute.Normalize` and then collapses empty/default values to null. |
| 7 | [#3159289743](https://github.com/aevatarAI/aevatar/pull/509#discussion_r3159289743) `IOwnerLlmConfigSource.cs:1` | ✅ resolved | 5/5 | `IOwnerLlmConfigSource` was moved out of `Aevatar.GAgents.Scheduled` into `src/Aevatar.AI.Abstractions/LLMProviders`, and `OwnerLlmConfigApplier` lives in `Aevatar.AI.Core`. |

**Totals:** 7 resolved · 0 partial · 0 unresolved · 0 cannot_determine

---

_Generated by `/opencode-pr-fix-check`. Threads are not auto-resolved — close them manually if you agree._

PR #510, opened 2026-04-29
fix: connect api-lark-bot proxy service during provisioning + projection OCC self-heal

### LINKED ISSUE #508: fix: Lark bot provisioning missing api-lark-bot proxy service + projection OCC self-heal
## Problem

When migrating a Lark bot to a new NyxID account via `POST /api/channels/registrations`, two issues cause the bot to be partially broken:

### 1. Missing `api-lark-bot` NyxID proxy service connection

`NyxLarkProvisioningService.ProvisionAsync` creates:
- Relay API key (`CreateRelayApiKeyAsync`)
- Channel bot (`RegisterChannelBotAsync`)
- Conversation route (`CreateDefaultRouteAsync`)
- Local registration mirror (`RegisterLocalMirrorAsync`)

But does **not** connect the `api-lark-bot` NyxID proxy service with the Lark app credentials (`app_id` + `app_secret`). This service is required for:
- Typing reaction (keyboard emoji on inbound message)
- Done reaction (swap after reply completes)
- Any direct Lark API calls proxied through NyxID

Without it, all reaction/emoji interactions fail silently (logged at Debug level with `NoPermissionToReact` or proxy 404).

### 2. Projection OCC blocks new registration visibility

The `channel-bot-registration` projection scope actor (`ProjectionMaterializationScopeGAgent`) hits an `EventStoreOptimisticConcurrencyException` on startup when the persisted version in Garnet is stale from a previous pod incarnation. The current behavior:
- `HandleObservedEnvelopeAsync` catches OCC → `ShouldPropagate` returns true → re-throws
- `RuntimeActorGrain.HandleEnvelopeAsyncCore` logs "retry exhausted" and drops the envelope
- The projection scope is permanently stuck until pod restart clears the stale Garnet state

This means new registrations committed to the store actor are invisible to the read model (and thus to `ResolveRegistrationAsync` in the turn runner) until a pod restart.

### Fix 1: Add `api-lark-bot` service connection to provisioning

After `RegisterChannelBotAsync` and `CreateDefaultRouteAsync`, add a step in `NyxLarkProvisioningService.ProvisionAsync`:

```csharp
await ConnectLarkBotProxyServiceAsync(
    request.AccessToken,
    request.AppId,
    request.AppSecret,
    ct);
```

Implementation:
- Use `NyxIdApiClient.CreateServiceAsync` with:
  ```json
  {
    "service_slug": "api-lark-bot",
    "credential": "{\"app_id\":\"...\",\"app_secret\":\"...\"}"
  }
  ```
- **Idempotent**: If the service already exists (409 Conflict or already connected), skip silently — this covers the re-registration case where the user already has `api-lark-bot` connected
- **Best-effort**: If this step fails, log warning but do NOT rollback the bot/route/mirror — the core relay path still works, only reactions are degraded
- Add to rollback chain only if subsequent steps fail

Edge cases:
- **Fresh NyxID account**: No `api-lark-bot` exists → create it
- **Re-registration**: `api-lark-bot` exists with old credentials → update credentials or skip if same app
- **Cross-app migration**: `api-lark-bot` exists with different app credentials → update with new credentials

### Fix 2: Projection OCC self-heal

In `ProjectionScopeGAgentBase.HandleObservedEnvelopeAsync`, when `EventStoreOptimisticConcurrencyException` is caught:

1. Reload the actor's event-sourced state from the store (re-hydrate from Garnet)
2. Retry `DispatchObservationAsync` once with the refreshed version
3. If retry succeeds → continue normally
4. If retry fails → fall through to existing failure propagation

This avoids the permanent stuck state without requiring a pod restart. The pattern is consistent with the existing event-sourced recovery model in the codebase.

## Files to change

- `agents/channels/Aevatar.GAgents.Channel.NyxIdRelay/NyxLarkProvisioningService.cs` — add service connection step
- `agents/channels/Aevatar.GAgents.Channel.NyxIdRelay/NyxTelegramProvisioningService.cs` — same pattern for Telegram if applicable
- `src/Aevatar.CQRS.Projection.Core/Orchestration/ProjectionScopeGAgentBase.cs` — OCC retry
- `src/Aevatar.AI.ToolProviders.NyxId/NyxIdApiClient.cs` — may need `GetServiceBySlugAsync` if not exists
- Tests for both fixes

## Verification

1. `POST /api/channels/registrations` with Lark credentials → verify `api-lark-bot` service exists via `nyxid proxy discover`
2. Send message to Lark bot → verify typing reaction appears → verify DONE reaction after reply
3. Simulate projection OCC by writing stale version → verify self-heal without pod restart

eanzhao:
## Summary

Fixes two issues that caused Lark bot migration to a new NyxID account to be partially broken:

### 1. Missing `api-lark-bot` proxy service connection

`NyxLarkProvisioningService.ProvisionAsync` now calls `TryConnectLarkBotProxyServiceAsync` after bot+route creation. This connects the `api-lark-bot` NyxID proxy service with the Lark app credentials (`app_id` + `app_secret`), enabling:
- Typing reaction (keyboard emoji on inbound message)
- Done reaction (swap after reply completes)
- Any direct Lark API calls proxied through NyxID

Best-effort: 409 conflict (service already exists) or any other error is logged as warning but does not fail provisioning. Handles fresh accounts, re-registration, and cross-app migration.

### 2. Projection OCC self-heal

`ProjectionScopeGAgentBase.HandleObservedEnvelopeAsync` now attempts a single self-heal retry when `EventStoreOptimisticConcurrencyException` is caught:
1. Replays state from the event store (`EventSourcing.ReplayAsync`) to resync `_currentVersion`
2. Retries `DispatchObservationAsync` once
3. On success: continues normally (no throw)
4. On failure: falls through to existing throw behavior

This eliminates the permanent stuck state where the channel-bot-registration projection was blocked until pod restart.

## Test plan
- [x] 706/706 channel runtime tests pass (0 failures)
- [x] CI architecture guards pass
- [x] Build succeeds (0 errors)
- [x] Happy-path provisioning test updated to verify `api-lark-bot` service creation
- [x] Rollback test updated to account for new service call ordering

Closes #508

### DIFF
--- a/agents/channels/Aevatar.GAgents.Channel.NyxIdRelay/NyxLarkProvisioningService.cs
+++ b/agents/channels/Aevatar.GAgents.Channel.NyxIdRelay/NyxLarkProvisioningService.cs
@@ -266,12 +266,6 @@ public sealed class NyxLarkProvisioningService : INyxLarkProvisioningService, IN
                 confirmedBot.Id,
                 confirmedApiKey.Id,
                 ct);
-
-            // Note: proxy service connection (api-lark-bot) is skipped during repair because
-            // the repair request does not carry Lark app credentials (AppId/AppSecret).
-            // The service was connected during the original ProvisionAsync call and is
-            // reusable across registrations.
-
             await RegisterLocalMirrorAsync(
                 registrationId,
                 nyxProviderSlug,
@@ -406,10 +400,10 @@ public sealed class NyxLarkProvisioningService : INyxLarkProvisioningService, IN
         try
         {
             var credential = JsonSerializer.Serialize(new { app_id = appId, app_secret = appSecret });
-            var body = JsonSerializer.Serialize(new { service_slug = DefaultNyxProviderSlug, credential });
+            var body = JsonSerializer.Serialize(new { service_slug = DefaultNyxProviderSlug, credential, label = $"Lark App {appId}" });
             await _nyxClient.CreateServiceAsync(accessToken, body, ct);
         }
-        catch (Exception ex)
+        catch (Exception ex) when (ex is not OperationCanceledException)
         {
             // Best-effort: 409 conflict (service already exists) or any other error is
             // non-fatal. The core relay path works without this; only typing reactions
--- a/agents/channels/Aevatar.GAgents.Channel.NyxIdRelay/NyxLarkProvisioningService.cs
+++ b/agents/channels/Aevatar.GAgents.Channel.NyxIdRelay/NyxLarkProvisioningService.cs
@@ -172,6 +172,10 @@ public sealed class NyxLarkProvisioningService : INyxLarkProvisioningService, IN
                 ct);
             routeId = await CreateDefaultRouteAsync(request.AccessToken, channelBotId, apiKeyId, ct);
 
+            // Best-effort: connect the api-lark-bot NyxID proxy service so typing
+            // reactions can call the Lark API. Intentionally NOT in the rollback chain
+            // because the service is reusable across registrations; a 409 on re-provision
+            // is the expected idempotent case, not an orphan to clean up.
             await TryConnectLarkBotProxyServiceAsync(
                 request.AccessToken,
                 request.AppId.Trim(),
--- a/agents/channels/Aevatar.GAgents.Channel.NyxIdRelay/NyxLarkProvisioningService.cs
+++ b/agents/channels/Aevatar.GAgents.Channel.NyxIdRelay/NyxLarkProvisioningService.cs
@@ -172,6 +172,12 @@ public sealed class NyxLarkProvisioningService : INyxLarkProvisioningService, IN
                 ct);
             routeId = await CreateDefaultRouteAsync(request.AccessToken, channelBotId, apiKeyId, ct);
 
+            await TryConnectLarkBotProxyServiceAsync(
+                request.AccessToken,
+                request.AppId.Trim(),
+                request.AppSecret.Trim(),
+                ct);
+
             var webhookUrl = $"{nyxBaseUrl}/api/v1/webhooks/channel/lark/{Uri.EscapeDataString(channelBotId)}";
             await RegisterLocalMirrorAsync(
                 registrationId,
@@ -260,6 +266,12 @@ public sealed class NyxLarkProvisioningService : INyxLarkProvisioningService, IN
                 confirmedBot.Id,
                 confirmedApiKey.Id,
                 ct);
+
+            // Note: proxy service connection (api-lark-bot) is skipped during repair because
+            // the repair request does not carry Lark app credentials (AppId/AppSecret).
+            // The service was connected during the original ProvisionAsync call and is
+            // reusable across registrations.
+
             await RegisterLocalMirrorAsync(
                 registrationId,
                 nyxProviderSlug,
@@ -383,6 +395,30 @@ public sealed class NyxLarkProvisioningService : INyxLarkProvisioningService, IN
             ct);
 
         return NyxApiResponseHelper.ExtractRequiredId(response, "channel_route_id");
+    }
+
+    private async Task TryConnectLarkBotProxyServiceAsync(
+        string accessToken,
+        string appId,
+        string appSecret,
+        CancellationToken ct)
+    {
+        try
+        {
+            var credential = JsonSerializer.Serialize(new { app_id = appId, app_secret = appSecret });
+            var body = JsonSerializer.Serialize(new { service_slug = DefaultNyxProviderSlug, credential });
+            await _nyxClient.CreateServiceAsync(accessToken, body, ct);
+        }
+        catch (Exception ex)
+        {
+            // Best-effort: 409 conflict (service already exists) or any other error is
+            // non-fatal. The core relay path works without this; only typing reactions
+            // are degraded when the proxy service is not connected.
+            _logger.LogWarning(
+                ex,
+                "Best-effort api-lark-bot proxy service connection failed (non-fatal). appId={AppId}",
+                appId);
+        }
     }
 
     private async Task RegisterLocalMirrorAsync(
--- a/src/Aevatar.CQRS.Projection.Core/Orchestration/ProjectionScopeGAgentBase.cs
+++ b/src/Aevatar.CQRS.Projection.Core/Orchestration/ProjectionScopeGAgentBase.cs
@@ -1,5 +1,4 @@
 using Aevatar.Foundation.Abstractions.Attributes;
-using Aevatar.Foundation.Abstractions.Persistence;
 using Aevatar.Foundation.Abstractions.Streaming;
 using Aevatar.Foundation.Core;
 using Aevatar.Foundation.Core.EventSourcing;
@@ -127,37 +126,24 @@ public abstract class ProjectionScopeGAgentBase<TContext>
         {
             if (ProjectionObservationFailurePolicy.ShouldPropagate(ex))
             {
-                if (ex is EventStoreOptimisticConcurrencyException)
+                if (ProjectionObservationFailurePolicy.ContainsOcc(ex))
                 {
-                    try
-                    {
-                        _logger.LogInformation(
-                            "Projection scope attempting OCC self-heal via state replay. actorId={ActorId}",
-                            Id);
-                        var replayedState = await EventSourcing!.ReplayAsync(Id, CancellationToken.None);
-                        if (replayedState is not null)
-                            State = replayedState;
-                        await DispatchObservationAsync(envelope, CancellationToken.None);
-                        _logger.LogInformation(
-                            "Projection scope OCC self-heal succeeded. actorId={ActorId}",
-                            Id);
-                        return;
-                    }
-                    catch (Exception retryEx)
-                    {
-                        _logger.LogWarning(
-                            retryEx,
-                            "Projection scope OCC self-heal retry also failed. actorId={ActorId}",
-                            Id);
-                    }
+                    EventSourcing?.DiscardPendingEvents();
+                    _logger.LogWarning(
+                        ex,
+                        "Projection scope hit OCC; stale pending events discarded. Self-heal will occur on next grain activation. actorId={ActorId} projectionKind={ProjectionKind}",
+                        Id,
+                        State.ProjectionKind);
                 }
-
-                _logger.LogWarning(
-                    ex,
-                    "Projection scope observation handling hit a retryable failure. actorId={ActorId} projectionKind={ProjectionKind} sessionId={SessionId}",
-                    Id,
-                    State.ProjectionKind,
-                    State.SessionId);
+                else
+                {
+                    _logger.LogWarning(
+                        ex,
+                        "Projection scope observation handling hit a retryable failure. actorId={ActorId} projectionKind={ProjectionKind} sessionId={SessionId}",
+                        Id,
+                        State.ProjectionKind,
+                        State.SessionId);
+                }
                 throw;
             }
 
--- a/src/Aevatar.CQRS.Projection.Core/Orchestration/ProjectionScopeGAgentBase.cs
+++ b/src/Aevatar.CQRS.Projection.Core/Orchestration/ProjectionScopeGAgentBase.cs
@@ -1,4 +1,5 @@
 using Aevatar.Foundation.Abstractions.Attributes;
+using Aevatar.Foundation.Abstractions.Persistence;
 using Aevatar.Foundation.Abstractions.Streaming;
 using Aevatar.Foundation.Core;
 using Aevatar.Foundation.Core.EventSourcing;
@@ -126,6 +127,31 @@ public abstract class ProjectionScopeGAgentBase<TContext>
         {
             if (ProjectionObservationFailurePolicy.ShouldPropagate(ex))
             {
+                if (ex is EventStoreOptimisticConcurrencyException)
+                {
+                    try
+                    {
+                        _logger.LogInformation(
+                            "Projection scope attempting OCC self-heal via state replay. actorId={ActorId}",
+                            Id);
+                        var replayedState = await EventSourcing!.ReplayAsync(Id, CancellationToken.None);
+                        if (replayedState is not null)
+                            State = replayedState;
+                        await DispatchObservationAsync(envelope, CancellationToken.None);
+                        _logger.LogInformation(
+                            "Projection scope OCC self-heal succeeded. actorId={ActorId}",
+                            Id);
+                        return;
+                    }
+                    catch (Exception retryEx)
+                    {
+                        _logger.LogWarning(
+                            retryEx,
+                            "Projection scope OCC self-heal retry also failed. actorId={ActorId}",
+                            Id);
+                    }
+                }
+
                 _logger.LogWarning(
                     ex,
                     "Projection scope observation handling hit a retryable failure. actorId={ActorId} projectionKind={ProjectionKind} sessionId={SessionId}",
--- a/src/Aevatar.CQRS.Projection.Core/Orchestration/ProjectionScopeGAgentBase.cs
+++ b/src/Aevatar.CQRS.Projection.Core/Orchestration/ProjectionScopeGAgentBase.cs
@@ -126,24 +126,17 @@ public abstract class ProjectionScopeGAgentBase<TContext>
         {
             if (ProjectionObservationFailurePolicy.ShouldPropagate(ex))
             {
+                // ShouldPropagate currently only returns true for OCC (direct or
+                // wrapped). Discard stale pending events so the grain can deactivate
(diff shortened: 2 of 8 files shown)


agents/channels/Aevatar.GAgents.Channel.NyxIdRelay/NyxLarkProvisioningService.cs
```diff
@@ -385,6 +397,30 @@ private async Task<string> CreateDefaultRouteAsync(
         return NyxApiResponseHelper.ExtractRequiredId(response, "channel_route_id");
     }
 
+    private async Task TryConnectLarkBotProxyServiceAsync(
+        string accessToken,
+        string appId,
+        string appSecret,
+        CancellationToken ct)
+    {
```
eanzhao, line 403: **[Consensus: 2 models — codex + glm-5.1]** severity=blocker, category=bug

The `/api/v1/keys` payload built here is missing the `label` field, which NyxID's `CreateKeyRequest` declares as required (`backend/src/handlers/keys.rs:150`, `label: String`). On a fresh account this request will be rejected; the warning is then swallowed and the typing/done reactions silently stop working — exactly the scenario this PR is supposed to fix.

Codex also flags a second concern: `service_slug = api-lark-bot` with no override policy. NyxID's auto-disambiguation kicks in on slug collision and creates `api-lark-bot-2`, while the local mirror still references `api-lark-bot` — leaving stale entries on each re-registration / cross-app migration.

Suggested fix:
1. Add `label = $"Lark App {appId}"` (or similar) to the body.
2. Decide on slug strategy: pass an explicit `slug` override and on 409 conflict, either rotate the existing service's credential or surface the conflict instead of silently disambiguating.
3. Compare against `NyxIdServicesTool.cs:106-117` which already includes `label`.

<details>

- **codex**: The payload is not a valid or sufficient `/api/v1/keys` create request. NyxID's `CreateKeyRequest` requires `label`, so fresh provisioning is rejected before creating the UserService; also, without a `slug` override, an existing `api-lark-bot` connection is auto-disambiguated (for example `api-lark-bot-2`) while the local mirror still uses `api-lark-bot`. Include `label`, pass/use the same `nyxProviderSlug` that registration stores, and handle existing-slug conflicts by updating/rotating that service's credential instead of silently ignoring them.
- **glm-5.1** (anchored at line 416): The request body omits `label` which `NyxIdServicesTool.CreateServiceAsync` always includes (`label = args.Str("label") ?? slug` at `NyxIdServicesTool.cs:110`). If the NyxID API requires or uses `label` for display/filtering, this service will have a missing label.
</details>

src/Aevatar.CQRS.Projection.Core/Orchestration/ProjectionScopeGAgentBase.cs
```diff
@@ -126,6 +127,31 @@ public async Task HandleObservedEnvelopeAsync(EventEnvelope envelope)
         {
             if (ProjectionObservationFailurePolicy.ShouldPropagate(ex))
             {
+                if (ex is EventStoreOptimisticConcurrencyException)
```
eanzhao: **[Consensus: 2 models — v4-pro + glm-5.1]** severity=major, category=test

The self-heal logic (replay → assign State → retry dispatch → fallthrough on second failure) is a significant behavioral change but has zero unit tests. The PR's test updates only cover provisioning HTTP request count.

Add tests covering at minimum:
1. `DispatchObservationAsync` throws `EventStoreOptimisticConcurrencyException` → self-heal succeeds → method returns normally.
2. Replay returns null → State stays untouched → retry dispatch behavior.
3. Retry dispatch also throws → original behavior (warning + throw via existing fallthrough).

<details>

- **v4-pro**: The OCC self-heal logic (replay + retry) has no test coverage. Without tests, the self-heal path—including the null-state guard, the retry-once policy, and the fallthrough-to-throw behavior—can silently regress.
- **glm-5.1** (anchored at line 127): The OCC self-heal logic is a significant behavioral change yet has zero test coverage. This is especially important because `EventSourcing.ReplayAsync` is called with `CancellationToken.None` and the version resync behavior depends on `EventSourcingBehavior._currentVersion` being updated as a side effect.
</details>

eanzhao: **[codex]** severity=major, category=bug

_Verified independently against `ProjectionObservationFailurePolicy.cs:7-21`._

`ProjectionObservationFailurePolicy.ShouldPropagate(ex)` recursively unwraps `ProjectionDispatchAggregateException.Failures`, `AggregateException.InnerExceptions`, and chained `InnerException` to detect retryable failures (including OCC). But the new self-heal branch only triggers on a top-level `ex is EventStoreOptimisticConcurrencyException`. A wrapped OCC (which is the more common form when the failure originates inside an aggregate dispatch) passes `ShouldPropagate` but skips self-heal and falls through to the throw — defeating the fix in production.

Use the same recursive detection as `ShouldPropagate`, or change `ProjectionObservationFailurePolicy` to return the matched root cause (e.g. `TryGetRetryableInner(ex, out var inner)`) and switch on that here.

src/Aevatar.CQRS.Projection.Core/Orchestration/ProjectionScopeGAgentBase.cs
```diff
@@ -126,6 +127,31 @@ public async Task HandleObservedEnvelopeAsync(EventEnvelope envelope)
...
+                {
+                    try
+                    {
+                        _logger.LogInformation(
```
eanzhao: **[k2p6]** severity=major, category=design

_Verified against `CLAUDE.md` architecture rules: 「正常路径禁止 replay：query path 和 projection path 不依赖 event replay/rebuild/backfill；replay 只属于后台修复/迁移/灾难恢复」._

Calling `EventSourcing.ReplayAsync` inside `HandleObservedEnvelopeAsync` adds event replay to the normal projection observation path. The project's stated rule (CLAUDE.md section "权威状态 / ReadModel / Projection") is that replay belongs only to background repair, migration, or disaster recovery — not online observation/read paths.

This isn't just a style preference — `_currentVersion` desync from upstream events is a runtime symptom; the right fix is usually to drop & restart the projection observation session or to have the projection pipeline coordinate version resync, not to bake replay into the dispatch hot path. Consider:
- Dropping the in-line replay; let the projection runtime / aggregate orchestrator rebuild the observation session via the standard pipeline.
- If a fast in-process recovery is genuinely needed, encapsulate it behind a documented "recovery seam" abstraction so it isn't entangled with normal observation.

src/Aevatar.CQRS.Projection.Core/Orchestration/ProjectionScopeGAgentBase.cs
```diff
@@ -126,6 +127,31 @@ public async Task HandleObservedEnvelopeAsync(EventEnvelope envelope)
...
+                            "Projection scope attempting OCC self-heal via state replay. actorId={ActorId}",
+                            Id);
+                        var replayedState = await EventSourcing!.ReplayAsync(Id, CancellationToken.None);
```
eanzhao: **[codex]** severity=major, category=bug

_Verified against `EventSourcingBehavior.cs:154-177` and the `_pending` field at `EventSourcingBehavior.cs:30`._

`ReplayAsync` rebuilds `_currentVersion` (line 175) but never clears the `_pending` list — that list is only ever pruned by `RemoveCommittedPendingPrefix` after a successful `ConfirmEventsAsync`. If `ConfirmEventsAsync` failed with OCC, `_pending` still holds the failed events. Self-heal then replays state, retries dispatch, and the next `ConfirmEventsAsync` call will append the stale failed events together with the new retry events.

At minimum, clear or recreate the pending queue before the retry. Cleaner: have `ReplayAsync` (or a new explicit `ResetPendingAsync`) drop pending state when called in a recovery scenario, since callers always want "forget what was uncommitted."

src/Aevatar.CQRS.Projection.Core/Orchestration/ProjectionScopeGAgentBase.cs
```diff
@@ -126,6 +127,31 @@ public async Task HandleObservedEnvelopeAsync(EventEnvelope envelope)
...
+                        if (replayedState is not null)
+                            State = replayedState;
```
eanzhao: **[glm-5.1]** severity=major, category=bug

_Verified: `[AllEventHandler]` decoration at line 106 means `StateGuard.WriteScope` is active, so the setter won't throw. But `OnStateChangedAsync` (`GAgentBase.TState.cs:49`) is only invoked from `ActivateAsync` and `PersistDomainEventsAsync` — direct property assignment skips it._

`State = replayedState` mutates `_state` but does not invoke `OnStateChangedAsync`. Subscribers that depend on this hook (downstream projections, telemetry, cache invalidation, etc.) will miss the replay-induced state change and stay out of sync.

Either (a) call `OnStateChangedAsync` explicitly after the replay assignment, or (b) route the recovery through a dedicated method that mirrors the activation path so all downstream consumers observe the rebuild consistently.

agents/channels/Aevatar.GAgents.Channel.NyxIdRelay/NyxLarkProvisioningService.cs
```diff
@@ -385,6 +397,30 @@ private async Task<string> CreateDefaultRouteAsync(
         return NyxApiResponseHelper.ExtractRequiredId(response, "channel_route_id");
```
eanzhao, line 395: **[glm-5.1]** severity=minor, category=design

When `ProvisionAsync` rolls back due to local mirror registration failure, the newly created `/api/v1/keys` proxy service is not deleted (`DELETE` requests in the rollback test only cover route, channel-bot, and api-key). On a fresh NyxID account, this leaves an orphaned `api-lark-bot` UserService entry — which then collides with the next `ProvisionAsync` attempt and triggers slug auto-disambiguation (see comment #1).

Either:
1. Add a best-effort `DELETE /api/v1/keys/{id}` to the rollback path (use the id from the create response).
2. Or document explicitly that this leak is intentional because the service is designed to be reused across registrations, and pair it with a slug-conflict resolution policy in `TryConnectLarkBotProxyServiceAsync`.

agents/channels/Aevatar.GAgents.Channel.NyxIdRelay/NyxLarkProvisioningService.cs
```diff
@@ -385,6 +397,30 @@ private async Task<string> CreateDefaultRouteAsync(
...
+        string accessToken,
+        string appId,
+        string appSecret,
+        CancellationToken ct)
+    {
+        try
+        {
+            var credential = JsonSerializer.Serialize(new { app_id = appId, app_secret = appSecret });
+            var body = JsonSerializer.Serialize(new { service_slug = DefaultNyxProviderSlug, credential });
+            await _nyxClient.CreateServiceAsync(accessToken, body, ct);
```
eanzhao, line 408: **[v4-pro]** severity=minor, category=bug

`catch (Exception ex)` will swallow an `OperationCanceledException` raised when `ct` is cancelled mid-`CreateServiceAsync`, turning intentional shutdown / cancellation into a logged warning. Use a filter:

```csharp
catch (Exception ex) when (ex is not OperationCanceledException)
{
    _logger.LogWarning(ex, "Best-effort api-lark-bot proxy service connection failed (non-fatal). appId={AppId}", appId);
}
```

Or catch `OperationCanceledException` explicitly and re-throw.

src/Aevatar.CQRS.Projection.Core/Orchestration/ProjectionScopeGAgentBase.cs
```diff
@@ -126,6 +127,31 @@ public async Task HandleObservedEnvelopeAsync(EventEnvelope envelope)
...
             {
+                if (ex is EventStoreOptimisticConcurrencyException)
+                {
+                    try
+                    {
+                        _logger.LogInformation(
+                            "Projection scope attempting OCC self-heal via state replay. actorId={ActorId}",
+                            Id);
+                        var replayedState = await EventSourcing!.ReplayAsync(Id, CancellationToken.None);
+                        if (replayedState is not null)
```
eanzhao: **[mimo-v2.5-pro]** severity=minor, category=bug

Both `ReplayAsync` and the retry `DispatchObservationAsync` are called with `CancellationToken.None`. `HandleObservedEnvelopeAsync(EventEnvelope envelope)` has no `CancellationToken` parameter, but the actor likely exposes a lifetime / shutdown token (typical Orleans patterns). Using `None` here means a graceful actor shutdown during a long replay can't preempt the self-heal.

eanzhao:
## Fix-check round 2 (5-model consensus, head `020c1e68`)

Re-ran the panel against the latest commit. Stats: ✅ 8 resolved · ⚠️ 1 partial · ❌ 1 unresolved.

The doc comment justifying no-rollback for the proxy service was accepted by 4/5 reviewers as a valid intentional design decision (the original review explicitly listed "document the leak as intentional" as one of two acceptable resolutions). [That thread](https://github.com/aevatarAI/aevatar/pull/510#discussion_r3158954532) is effectively closed.

Two items still need attention before merge:

### ❌ [Test coverage for the new OCC recovery path](https://github.com/aevatarAI/aevatar/pull/510#discussion_r3158954207) — 3/5 unresolved

The original replay-and-retry tests are correctly N/A now, but the *new* path (`ContainsOcc(ex) → DiscardPendingEvents() → rethrow → next-activation replay`) has no unit tests. Concrete asks:

- A test where `DispatchObservationAsync` throws a wrapped OCC (e.g. inside `ProjectionDispatchAggregateException` or `AggregateException`) and asserts:
  1. `EventSourcing.DiscardPendingEvents` is invoked, AND
  2. The original exception is rethrown so the grain deactivates.
- A test where a non-OCC retryable exception flows through and `DiscardPendingEvents` is **not** invoked (regression guard for the `else` branch).

### ⚠️ [Slug-collision policy still missing](https://github.com/aevatarAI/aevatar/pull/510#discussion_r3158954166) — 3/5 partial, 2/5 resolved

`label` is fixed. The remaining concern: `service_slug = DefaultNyxProviderSlug` is sent with no conflict-resolution policy. The new doc comment covers the *rollback* case (409 = idempotent reuse) but doesn't address NyxID's auto-disambiguation behavior, where a slug collision can return a *different* slug (e.g. `api-lark-bot-2`) rather than a 409 — leaving the local mirror referencing `api-lark-bot` while the server has `api-lark-bot-2`.

Either:
1. On 409, look up the existing service id and rotate its credential via `PUT /api/v1/keys/{id}` — keeps the slug stable.
2. Pass an explicit unique slug per registration (e.g. include `appId` or `channelBotId`) and persist it in the local mirror.

A reasonable middle-ground if you want minimal churn: capture and persist the *actual* slug NyxID returns from the create call (the response's `slug` field, accounting for disambiguation) into the local mirror, instead of assuming `DefaultNyxProviderSlug`.

---

<sub>🤖 5-model panel: deepseek-v4-pro, kimi-k2p6, glm-5.1, mimo-v2.5-pro, codex. Verdict basis: diff `bba24c3c..020c1e68`. (My earlier draft listing 3 unresolved items is superseded by this.)</sub>

eanzhao:
## Fix-check round 3 — all clear (5-model consensus, head `45c36c18`)

Re-ran the panel against `45c36c18`. **All 10 inline comments now resolved by majority** (vs 7/1/2 in round 1, 8/1/1 in round 2).

The new test commit (`45c36c18`) closes the last unresolved item — [test coverage for the OCC recovery path](https://github.com/aevatarAI/aevatar/pull/510#discussion_r3158954207) — flipping it from 3/5 unresolved to 5/5 resolved. The added tests cover exactly the asks:

- ✅ `ContainsOcc` policy: direct OCC, wrapped OCC, non-OCC false case
- ✅ `HandleObservedEnvelopeAsync`: direct OCC triggers `DiscardPendingEvents`, wrapped OCC also triggers it, non-OCC failures do **not** invoke it (regression guard for the `else` branch)

Nice and tight. From a review-coverage standpoint this PR is ready to merge.

### Minor non-blocking note

2/5 reviewers (glm-5.1, codex) continue to flag the [slug-collision strategy](https://github.com/aevatarAI/aevatar/pull/510#discussion_r3158954166) as partial — concerned that NyxID can auto-disambiguate (`api-lark-bot` → `api-lark-bot-2`) without returning 409, leaving the local mirror referencing the wrong slug. The majority (3/5) accept the current design, but if you want to belt-and-suspenders this later, persisting the actual `slug` returned by `CreateServiceAsync` into the local mirror is a small change that would close it. Not a merge blocker.

---

<sub>🤖 5-model panel: deepseek-v4-pro, kimi-k2p6, glm-5.1, mimo-v2.5-pro, codex. Verdict basis: diff `bba24c3c..45c36c18`. Round trip: 10 inline comments → 9 actionable findings → 7→8→10 resolved across 3 commits.</sub>

Issue #514, opened 2026-04-29
Studio script member lifecycle Build returns to workflow editor


potter-sun:
## Summary
When a Studio member is bound to a Script implementation, clicking the top lifecycle `Build` button from the `Bind` step can return to the workflow editor instead of the Script build surface.

## Reproduction
1. Open Studio with a bound Script member selected, for example `memberId=script-member&step=bind`.
2. Use a member whose display name, published service id, and script id are not the same, for example `draft-test`, `service-script-alpha`, and `script-alpha`.
3. Click the top lifecycle `Build` button.

## Actual Behavior
Studio navigates back to the workflow build flow or keeps the wrong build identity, so the selected Script member is not shown in the Script build panel.

## Expected Behavior
Studio should resolve the selected member implementation kind and return to the Script build surface with the authoritative `scriptId` selected.

## Impact
Users lose context after binding a Script member and can accidentally continue editing or inspecting the wrong build surface.

## Acceptance Criteria
- Clicking `Build` from `Bind` for a Script member opens the Script build surface.
- The selected script id comes from the member implementation reference or published revision, not from display name or published service id fallback.
- Workflow members continue to return to the workflow editor.
- A regression test covers the case where display name, service id, and script id differ.

Issue #511, opened 2026-04-29
Evaluate scope-scoped service catalog API for Studio Bind endpoints


potter-sun:
## Problem

After binding a Studio Script member, the Bind page can land on a state where the Endpoint selector is empty and both `Send smoke test` and `Continue to Invoke` remain disabled.

Observed request:

```http
GET /api/services?tenantId=296c7d52-d2a6-4291-b1fb-38c5e7d0c0a2&appId=default&namespace=default
```

This returns `403` in the Studio environment. Because the Bind page currently derives endpoint options from `ServiceCatalogSnapshot.endpoints` returned by this global `/api/services` list call, the page cannot populate the Endpoint selector even after Bind succeeds.

## Current frontend data flow

- Studio calls `servicesApi.listServices({ tenantId: scopeId, appId: "default", namespace: "default" })`.
- That maps to `GET /api/services?...`.
- Bind receives the returned `ServiceCatalogSnapshot[]`.
- Endpoint options are read from `selectedService.endpoints`.
- Revisions/supporting details already use scope-scoped APIs such as `/api/scopes/{scopeId}/services/{serviceId}/revisions`.

This means Bind mixes two authorization/data boundaries:

- service catalog/current endpoint list: global `/api/services?...`
- revisions/details: scope-scoped `/api/scopes/{scopeId}/services/...`

## Why this needs backend evaluation

Studio users are operating inside a scope/team context. It may be correct that they do not have access to the global `/api/services?...` surface. If so, Studio should not rely on that API to show current published contract endpoints.

## Proposed backend direction

Please evaluate adding a scope-scoped service catalog query, for example:

```http
GET /api/scopes/{scopeId}/services
GET /api/scopes/{scopeId}/services/{serviceId}
```

The response should provide the current service catalog/read-model view needed by Studio Bind/Invoke, including at least:

- `serviceId`
- `displayName`
- serving revision ids/status as appropriate
- `endpoints[]` with `endpointId`, `displayName`, `kind`, `requestTypeUrl`, `responseTypeUrl`, `description`

Then frontend Studio can switch from global `servicesApi.listServices(...)` to a scope-scoped API such as `scopeRuntimeApi.listServices(scopeId)`.

## Alternative / short-term fallback

[...]

PR #518, opened 2026-04-29
fix: stabilize Studio script lifecycle binding flow

### LINKED ISSUE #507: Verify Studio Script save observation after backend read-model activation
## Problem

#496 fixed the backend read-model activation gap for Script save observation.

We need to re-run the Studio frontend flow against updated `dev` and confirm the Script Build UI now moves from `save accepted / pending` to `catalog-applied`, then unlocks Bind.

## Scope

- Re-test Studio Script Build after #496.
- Confirm frontend polling of `observeSaveScript` reaches `applied`.
- Confirm `GET /api/scopes/{scopeId}/scripts?includeSource=true` includes the saved script after save.
- Confirm `Continue to Bind` unlocks only after catalog-applied.
- Confirm Bind uses `bindScopeScript`.
- Confirm Invoke uses the bound service endpoint contract, not draft-run/chat assumptions.
- Capture any remaining UX or timing problems as follow-up issues.

## Acceptance Criteria

- Saving a new Script revision shows pending/waiting state first.
- `save-observation` eventually returns `applied` with non-null `currentScript`.
- Refetch catalog returns the saved script.
- Frontend status changes to `Catalog applied / Ready to bind`.
- `Continue to Bind` becomes enabled.
- Binding succeeds and selects/refetches the bound service.
- Invoke page shows the correct endpoint contract.
- No frontend workaround treats pending as applied.

## References

- Backend fix: #496
- Original backend issue: #488
- Script lifecycle issue: #442
- Frontend PR: #443

potter-sun:
## Summary

- Keep Script save/bind revision semantics separate from service revision identity, so user-entered Script revisions such as `draft-1` are not sent as top-level binding revision ids.
- Send the Script service id when binding a Script member, preventing refresh-time binds from falling back to the default service key.
- Improve the Bind page layout for endpoint-discovery-pending states and allow the lifecycle to continue to Invoke while #511 is unresolved, without presenting the temporary endpoint as a real discovered contract.

## Issue links

Related to #507
Refs #511
Refs #514

## Validation

- `npm --prefix apps/aevatar-console-web test -- --runTestsByPath src/shared/studio/api.test.ts src/pages/studio/index.test.tsx src/modules/studio/scripts/ScriptsWorkbenchPage.test.tsx --runInBand` (112 tests passed)
- `npm --prefix apps/aevatar-console-web test -- --runTestsByPath src/pages/studio/components/bind/StudioMemberBindPanel.test.tsx --runInBand` (6 tests passed)
- `git diff --check`

## Notes

This PR intentionally does not close #507 because final end-to-end verification still depends on the backend endpoint/catalog follow-up tracked in #511.

### DIFF
--- a/apps/aevatar-console-web/src/modules/studio/scripts/ScriptsWorkbenchPage.tsx
+++ b/apps/aevatar-console-web/src/modules/studio/scripts/ScriptsWorkbenchPage.tsx
@@ -287,6 +287,10 @@ function normalizeStudioId(value: string, fallbackPrefix: string): string {
     .slice(0, 14)}`;
 }
 
+function trimOptional(value: string | null | undefined): string {
+  return String(value || '').trim();
+}
+
 function createStarterPackage(): ScriptPackage {
   return createSingleSourcePackage(STARTER_SOURCE);
 }
@@ -1757,10 +1761,16 @@ const ScriptsWorkbenchPage: React.FC<ScriptsWorkbenchPageProps> = ({
     }
 
     const scriptId = normalizeStudioId(scopeScript.scriptId || selectedDraft.scriptId, 'script');
-    const scriptRevision = normalizeStudioId(
+    const scriptRevision = trimOptional(
       scopeScript.activeRevision || selectedDraft.revision,
-      'rev',
     );
+    if (!scriptRevision) {
+      setNotice({
+        type: 'warning',
+        message: 'Save the current script into the scope before binding it.',
+      });
+      return;
+    }
 
     setBindPending(true);
     setNotice(null);
@@ -2371,8 +2381,15 @@ const ScriptsWorkbenchPage: React.FC<ScriptsWorkbenchPageProps> = ({
       (selectedDraft.scopeDetail?.script?.activeRevision || selectedDraft.baseRevision),
   );
   const scopeBindingScript = selectedDraft?.scopeDetail?.script || null;
+  const scopeBindingRevision = trimOptional(
+    scopeBindingScript?.activeRevision || selectedDraft?.revision,
+  );
   const canBindScope = Boolean(
-    selectedDraft && scopeBacked && scopeBindingScript && !bindPending,
+    selectedDraft &&
+      scopeBacked &&
+      scopeBindingScript &&
+      scopeBindingRevision &&
+      !bindPending,
   );
   const scopeSelectionId = selectedDraft?.scopeDetail?.script?.scriptId || '';
   const hasScopeChanges = isScopeDetailDirty(selectedDraft);
@@ -3253,7 +3270,7 @@ const ScriptsWorkbenchPage: React.FC<ScriptsWorkbenchPageProps> = ({
             <div className="console-scripts-eyebrow">Script</div>
             <div className="console-scripts-detail-copy">
               {scopeBindingScript?.scriptId || selectedDraft?.scriptId || '-'} ·{' '}
-              {scopeBindingScript?.activeRevision || selectedDraft?.revision || '-'}
+              {scopeBindingRevision || 'save required'}
             </div>
           </div>
           <div style={{ marginTop: 16 }}>
--- a/apps/aevatar-console-web/src/modules/studio/scripts/ScriptsWorkbenchPage.tsx
+++ b/apps/aevatar-console-web/src/modules/studio/scripts/ScriptsWorkbenchPage.tsx
@@ -248,10 +248,6 @@ function createDraftKey(prefix: string): string {
   return `${prefix}_${Date.now().toString(36)}_${draftCounter.toString(36)}`;
 }
 
-function createRevisionSeed(): string {
-  return `draft-${new Date().toISOString().replace(/[-:.TZ]/g, '').slice(0, 14)}`;
-}
-
 function buildScopePageHref(
   path: string,
   scopeId: string,
@@ -289,13 +285,6 @@ function normalizeStudioId(value: string, fallbackPrefix: string): string {
     .toISOString()
     .replace(/[-:.TZ]/g, '')
     .slice(0, 14)}`;
-}
-
-function buildScopeScriptBindingRevisionId(
-  scriptId: string,
-  scriptRevision: string,
-): string {
-  return normalizeStudioId(`${scriptId}-${scriptRevision}`, 'rev');
 }
 
 function createStarterPackage(): ScriptPackage {
@@ -369,7 +358,7 @@ function createDraft(index: number, seed: Partial<ScriptDraft> = {}): ScriptDraf
   return {
     key: seed.key || createDraftKey('script_draft'),
     scriptId: seed.scriptId || `script-${index}`,
-    revision: seed.revision || createRevisionSeed(),
+    revision: seed.revision || '',
     baseRevision: seed.baseRevision || '',
     reason: seed.reason || '',
     input: seed.input || '',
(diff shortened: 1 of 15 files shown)


apps/aevatar-console-web/src/modules/studio/scripts/ScriptsWorkbenchPage.tsx
```diff
@@ -1772,20 +1761,16 @@ const ScriptsWorkbenchPage: React.FC<ScriptsWorkbenchPageProps> = ({
       scopeScript.activeRevision || selectedDraft.revision,
```
eanzhao, line 1765: **[codex]** severity=major, category=bug

`scriptRevision` still falls back through `normalizeStudioId(..., 'rev')` when neither `scopeScript.activeRevision` nor `selectedDraft.revision` exists. Because `canBind` below no longer requires a saved revision, the bind flow can now send a synthetic revision id for an unsaved/local draft. That can bind a non-existent script revision. Keep `scriptRevision` as `scopeScript.activeRevision || selectedDraft.revision`, block bind when it is empty, and surface a save-required message instead of generating a fallback revision.

apps/aevatar-console-web/src/shared/studio/api.test.ts
```diff
@@ -740,6 +740,7 @@ describe('studioApi host-session requests', () => {
     const result = await studioApi.bindScopeScript({
       scopeId: 'scope-1',
       displayName: 'script-1',
+      serviceId: 'script-1',
```
eanzhao, line 743: **[mimo-v2.5-pro]** severity=major, category=test

This test still passes `revisionId: 'rev-1'` in the input and asserts it appears in the JSON body, but production code in `api.ts` no longer sends `revisionId` (it was removed from the `compactObject` call). The test will either fail or mask a regression by asserting stale behavior.

Fix: remove `revisionId` from the input object and from the `JSON.parse` assertion to match production code.

apps/aevatar-console-web/src/pages/studio/components/bind/StudioMemberBindPanel.tsx
```diff
@@ -915,15 +1065,15 @@ const StudioMemberBindPanel: React.FC<StudioMemberBindPanelProps> = ({
                 <Button
                   block
                   icon={<LinkOutlined />}
-                  disabled={!selectedService || !selectedEndpoint}
+                  disabled={!selectedService}
```
eanzhao: **[mimo-v2.5-pro]** severity=minor, category=bug

The `Continue to Invoke` button is only disabled when `!selectedService`, but `invokeEndpointId` falls back to `'chat'` when `selectedService` exists but has no endpoints. If `selectedEndpoint` is undefined and the service has no endpoints, `onContinueToInvoke` is called with `('chat', 'chat')` — the endpoint id becomes `'chat'` rather than the real service's endpoint id.

This is likely intentional as a graceful degradation, but the naming is misleading: `invokeEndpointId` suggests a real discovered endpoint. Consider renaming to `fallbackEndpointId` or adding a comment to clarify this is a temporary placeholder until #511 resolves.

apps/aevatar-console-web/src/pages/studio/index.tsx
```diff
@@ -2604,6 +2632,34 @@ const StudioPage: React.FC = () => {
     () => studioMembersQuery.data?.members ?? [],
     [studioMembersQuery.data?.members],
   );
+  useEffect(() => {
```
eanzhao, line 2635: **[mimo-v2.5-pro]** severity=minor, category=bug

This `useEffect` reads `routeState.memberId` and `studioScopeMembers` but also writes `selectedScriptId`. If `studioScopeMembers` changes on the next render (e.g. query refetch) after `selectedScriptId` is set, the effect re-runs and may re-trigger, clearing `selectedWorkflowId` and `setTemplateWorkflow` again.

Consider adding a guard like `if (selectedScriptId === scriptId) return;` after computing `scriptId`, or narrowing the dependency array so the effect only fires when the route member actually changes.

apps/aevatar-console-web/src/pages/studio/index.tsx
```diff
@@ -6371,13 +6430,25 @@ const StudioPage: React.FC = () => {
...
-        : null,
-    [invokeTargetServiceId, runtimeConsoleServices],
+    () => {
+      if (!invokeTargetServiceId) {
+        return null;
+      }
+
+      const matchedService =
+        runtimeConsoleServices.find(
+          (service) => service.serviceId === invokeTargetServiceId,
```
eanzhao, line 6442: **[mimo-v2.5-pro]** severity=minor, category=design

When `matchedService` is null, `createEndpointDiscoveryPendingInvokeService` is called with `currentMemberLabel` as the label. If `currentMemberLabel` is empty/null, the fallback chain `(label || serviceId)` inside the factory is fine — but the broader concern is that this silently fabricates a synthetic service with a hardcoded `'chat'` endpoint. This is intentional per the PR notes, but consider adding a runtime log/warning so that if the backend never publishes real endpoints, the synthetic service doesn't silently mask a real problem.

apps/aevatar-console-web/src/pages/studio/components/bind/StudioMemberBindPanel.tsx
```diff
@@ -915,15 +1065,15 @@ const StudioMemberBindPanel: React.FC<StudioMemberBindPanelProps> = ({
                 <Button
```
eanzhao, line 1066: **[glm-5.1]** severity=major, category=bug

The `disabled` check was relaxed from `!selectedService || !selectedEndpoint` to `!selectedService`. This correctly unblocks invoke when the service has zero endpoints (the `invokeEndpointId` falls back to `'chat'`), but it **also** unblocks invoke when the service *does* have endpoints and the user simply hasn't selected one yet. In that case `invokeEndpointId` resolves to `''` (empty string), which is passed straight through to `onContinueToInvoke`.

Suggested fix:
```tsx
disabled={!selectedService || (hasEndpointOptions && !selectedEndpoint)}
```
This preserves the new fallback-to-chat path when `!hasEndpointOptions` while still requiring an explicit endpoint selection when one is available.

apps/aevatar-console-web/src/pages/studio/index.tsx
```diff
@@ -2604,6 +2632,34 @@ const StudioPage: React.FC = () => {
...
+
+    const routeMember = studioScopeMembers.find(
+      (member) => trimOptional(member.memberId) === routeMemberId,
+    );
+    if (
+      normalizeStudioMemberBindingImplementationKind(
+        routeMember?.implementationKind,
+      ) !== 'script'
+    ) {
+      return;
```
eanzhao, line 2649: **[k2p6]** severity=major, category=bug

_(Re-anchored from k2p6's reported line 2623 — the actual inverted ternary lives at this line.)_

The `scriptId` ternary is inverted: when `routeMember.publishedServiceId` exists (truthy), `scriptId` is assigned `''` and the effect bails out. This means URL navigation to an already-bound script member never auto-selects it in the workbench.

**Fix:** Prefer the published service id (or display name) instead of discarding it:
```ts
const scriptId = trimOptional(routeMember?.publishedServiceId)
  || trimOptional(routeMember?.displayName);
```

apps/aevatar-console-web/src/shared/studio/models.ts
```diff
@@ -503,6 +503,7 @@ export const getStudioDefaultRouteTargetCurrentRevision =
 export interface StudioScopeScriptBindingInput {
   readonly scopeId: string;
   readonly displayName?: string | null;
+  readonly serviceId?: string | null;
   readonly scriptId: string;
   readonly scriptRevision: string;
```
eanzhao, line 508: **[glm-5.1]** severity=minor, category=design

`revisionId` is still declared in `StudioScopeScriptBindingInput` but no caller passes a meaningful value anymore (one passes `null`, the others omit it entirely), and `api.ts` no longer includes it in the request body. Per project convention (*"把不需要的直接删除, 无需考虑兼容性"*), this field should be removed from the interface to avoid misleading future callers.

Issue #500, opened 2026-04-28
Actor model evolution patterns — lazy state migration vs projection-driven splits


eanzhao:
## Background

Discussion seeded by #498 (Adopt AgentKind + Kind Registry for runtime actor
identity). #498 covers the **identity layer** — kind ≠ CLR type. This issue
covers the **state / event layer**: when a refactor changes business model
shape (split, merge, re-key, schema upgrade), what is the prescribed pattern,
and what infrastructure does each pattern need?

The trigger was a proposal to ship a generic
`IActorMigration { Migrate(old) → new }` interface that runs lazily on
`OnActivateAsync`. That pattern is **partially correct** — it is the right
answer for narrow within-actor upgrades and the wrong answer for cross-actor
splits / merges / re-keying. We need the matrix written down before we start
adding interfaces.

## Decision matrix

| Evolution type | Example | Correct mechanism | New abstraction needed? |
|---|---|---|---|
| Cross-assembly CLR rename | `ChannelRuntime.X` → `Scheduled.X` | `[LegacyAgentKind]` | No (covered by #498) |
| Proto field add / remove | `reserved 8,10,11,12` on a state proto | proto3 + `[LegacyProtoFullName]` | No |
| Within-actor logic upgrade / state recompute | derived field algorithm changes | lazy `IActorStateMigration` | **Yes — narrow, deferred until first real case** |
| Within-actor field hygiene | trim, case-fold, ID normalization | same | **same** |
| Actor split (one → many) | `SkillRunner` → `SkillDefinition` + `SkillExecution` | projection-pipeline + bootstrap-from-projection + retire | **Yes — bootstrap port (separate issue)** |
| Actor merge (many → one) | re-aggregation | same | same |
| Actor identity re-keying | `skill-runner-{user}-{name}` → `skill-definition-{team}-{name}` | `IActorRedirectSpec` | **Yes — separate issue** |
| **Event semantic change** | `UserUpdated` semantics shift over time | new TypeUrl, never mutate existing event | No — doctrine only |
| Hard schema break (no compat path) | rare — disaster scenarios | offline tooling + dual-write + cutover | No (out of band) |

## Where lazy on-activation migration is right (narrow, real)

`IActorStateMigration` is the right tool when the migration is **fully within
one actor's boundary**:

- The actor's persisted state proto needs a field recomputed (e.g., a derived
  field's algorithm changed).
- One-time hygiene: ID format normalization, case folding, trimming.
[...]

Issue #498, opened 2026-04-28
Adopt AgentKind + Kind Registry for runtime actor identity


eanzhao:
## Background

PR #495 introduces spec-driven startup cleanup for retired runtime actor types
in the deleted `Aevatar.GAgents.ChannelRuntime` assembly. PR #497 will split
`SkillRunnerGAgent` into two actors and would otherwise reproduce the same
activation-failure mode #495 fixes — but for live business actors, not retired
ones.

The original framing of this issue treated this as a governance problem
(classify every refactor as `drop/rebuild` / `preserve/migrate` /
`dormant/delete`; gate via CI guard). On reflection that is the symptom layer.
The root cause is that `RuntimeActorGrainState.AgentTypeName` persists a CLR
fully-qualified type name, and `RuntimeActorGrain.OnActivateAsync` resolves it
via `Type.GetType` + `AppDomain.GetAssemblies()`. Persisted authoritative state
references a runtime-incidental detail, so every rename / move / split becomes
a destructive migration concern.

The architecturally correct fix is to replace CLR-name identity with a stable
business **kind token**. Once that's in, most of the original governance
proposal collapses into a no-op.

## Diagnosis (preserved from original framing)

1. **PR #495 only handles registered retired actor specs.**
   It does not detect future deleted / renamed actor runtime types. A persisted
   `RuntimeActorGrainState.AgentTypeName` only gets cleaned if an
   `IRetiredActorSpec` declares the relevant actor ids and retired type tokens.

2. **`LegacyClrTypeName` is a payload codec compatibility layer, not an
   identity-migration tool.** It is consumed by
   `ProtobufContractCompatibility.IsCompatibleClrTypeName` for state /
   event / command deserialization. It does not rewrite the runtime actor
   implementation CLR type stored in `RuntimeActorGrainState.AgentTypeName`.
   If the stored `AgentTypeName` points to a deleted implementation type,
   activation fails in `RuntimeActorGrain.ResolveAgentType` before payload
   compatibility matters.

3. **The 3-classification governance scheme is a workaround for the underlying
   brittleness, not a fix.** `preserve/migrate` requires destructive writes
   on production state for what is fundamentally a naming change.
   `dormant/delete` relies on "expected to be unreachable", which contradicts
   the very reachability paths (callbacks, streams, projections, query-time
   activation) that motivated #495 in the first place.

## Decision: replace CLR-name identity with stable `AgentKind` tokens

[...]

PR #535, opened 2026-04-30
refactor(runtime): adopt AgentKind + Kind Registry for actor identity (#498 Phase 1)

eanzhao:
## Summary

Phase 1 of issue #498: replaces CLR-name actor identity with stable business
`AgentKind` tokens, additively, without mutating any persisted state. Unblocks
PR #497 (`SkillRunner` split) which can now ship as a kind-alias registration
instead of a destructive runtime-type rewrite.

The original framing of #498 treated this as a governance problem
(classify every removed actor as `drop/rebuild` / `preserve/migrate` /
`dormant/delete`; CI gate per PR). On reflection that was the symptom layer.
The root cause is `RuntimeActorGrainState.AgentTypeName` persisting a CLR
fully-qualified type name and `RuntimeActorGrain.OnActivateAsync` resolving
it via `Type.GetType` + `AppDomain.GetAssemblies()`. Authoritative state
referenced a runtime-incidental implementation detail, so every rename / move /
split became a destructive migration concern. Replacing the identity layer
collapses most of the original governance proposal to a no-op.

## What's in

| Area | What changed |
|---|---|
| Identity envelope | New `RuntimeActorIdentity` Orleans-serialized record (`Kind`, `StateSchemaVersion`, `LegacyClrTypeName`); embedded in `RuntimeActorGrainState` at `[Id(7)]`. |
| Registry | `IAgentKindRegistry` returns opaque `AgentImplementation(Func<IAgent>, Type stateContract, Metadata)`. CLR `Type` of the agent class is **not** on the contract surface — leaves the door open for scripted / workflow / out-of-process implementations behind the same kind. |
| Attributes | `[GAgent("module.entity")]` (single-valued, declares kind), `[LegacyAgentKind("...")]` (multi-valued, claims old kinds during identity-only refactors). Existing `[LegacyClrTypeName]` doubles for CLR-name aliasing. |
| Default impl | `AgentKindRegistry` + `AgentKindRegistryBuilder` (assembly scan + explicit register); `ReflectionLegacyAgentClrTypeResolver` encapsulates the previous reflection probe so the grain itself stays free of `Type.GetType`. |
| Grain wiring | `RuntimeActorGrain.OnActivateAsync` resolves identity via the registry (kind first, then legacy CLR name with lazy-tag write-back). `AgentTypeName` is preserved untouched until Phase 3 hard-deprecation so mixed-version pods stay compatible. New `InitializeAgentByKindAsync(kind)` + `GetAgentKindAsync()` on `IRuntimeActorGrain`; legacy `InitializeAgentAsync(typeName)` retained and re-routed through the registry. |
| DI | `services.AddAevatarAgentKindRegistry(builder => …)` accumulates contributions from each module; both Orleans and in-memory runtime extensions wire it by default. |
| ADRs | `docs/adr/0019-stable-agent-kind-identity.md` (supersedes the original 3-classification framing). `docs/adr/0020-actor-state-version-placement.md` (companion: `state_schema_version` lives on the runtime envelope, not on business state protos — pre-records #500's consumer contract). |
| CI guard | `tools/ci/agent_kind_naming_guard.sh` enforces `^[a-z0-9]+(\.[a-z0-9-]+)+$` and rejects `-v\d+` tails on `[GAgent]` / `[LegacyAgentKind]`. Wired into `architecture_guards.sh`. |
| Tests | Format validation, registry registration / resolution (direct, legacy alias, CLR-name reverse lookup), DI wiring (in-memory + Orleans), `RuntimeActorIdentity` Orleans round-trip, legacy-row deserialization without `Identity`. |

## What's deliberately not in

- **Decoration of every existing agent class.** Phase 1 keeps the
  transitional reflection fallback (encapsulated behind
  `ILegacyAgentClrTypeResolver`) so un-decorated classes continue to
  activate. Modules opportunistically decorate as they're touched (PR #497
  is the first; existing classes follow in Phase 2).
- **Spec migration in PR #495.** The retired-actor specs continue to key
  on CLR-name tokens for now; their migration to `AgentKind` tokens lands
  alongside or after Phase 2.
- **`Type.GetType` removal from the in-memory `LocalActorRuntime`.** Issue
  scope was the Orleans grain; the in-memory path is for tests + dev and
  can mirror the architecture in a follow-up.
- **The `IActorStateMigration<TState>` interface** (issue #500). The
  `state_schema_version` field is reserved on the envelope so #500's
  consumer contract reads from a known location; the migration interface
  itself lands with the first concrete migration case.

## Why now

PR #497 is blocked on this because deleting `SkillRunnerGAgent` without an
identity layer would break activation for every live skill-runner row in
production: `Type.GetType("Aevatar.GAgents.Scheduled.SkillRunnerGAgent")`
would return null and every envelope would be dropped (`Dropping envelope
for actor {ActorId}: initialization failed`). After this PR ships, PR #497
adds `[GAgent("scheduled.skill-definition")] [LegacyAgentKind("scheduled.skill-runner")]`
on `SkillDefinitionGAgent` and the existing rows lazy-tag onto the new
implementation with zero state mutation.

## Phased rollout (recap)

- **Phase 1 (this PR):** identity envelope + registry + attributes + grain
  wiring + ADRs + CI guard. Existing un-decorated classes keep activating
  via the transitional reflection fallback.
- **Phase 2:** modules opportunistically add `[GAgent]` / `[LegacyAgentKind]`
  as they're touched; #495 specs migrate from CLR-name tokens to
  `AgentKind` tokens.
- **Phase 3:** telemetry confirms no recent reads of `AgentTypeName` as
  primary identity (recommended grace ≥ 4 weeks). Mark `AgentTypeName` as
  `reserved`; remove default registration of `ILegacyAgentClrTypeResolver`.

## Test plan

- [x] `dotnet test test/Aevatar.Foundation.Core.Tests/...` — 205/205 pass (incl. 30 new TypeSystem tests)
- [x] `dotnet test test/Aevatar.Foundation.Runtime.Hosting.Tests/...` — 159/174 pass, 15 pre-existing skips
- [x] `dotnet test test/Aevatar.Architecture.Tests/...` — 105/106 pass, 1 pre-existing skip
- [x] `dotnet test test/Aevatar.GAgents.ChannelRuntime.Tests/...` — 767/767 pass
- [x] `dotnet test test/Aevatar.Hosting.Tests/...` — 59/59 pass
- [x] `dotnet test test/Aevatar.Integration.Tests/...` — 286/290 pass, 4 pre-existing skips
- [x] `bash tools/ci/agent_kind_naming_guard.sh` — ok
- [x] `bash tools/docs/lint.sh` — ok
- [x] `bash tools/ci/architecture_guards.sh` — ok up to the playground asset drift step (pre-existing pnpm/tsc tooling gap on `dev`, unrelated to this change)

## Related

- Parent issue: #498
- Companion: #500 (consumer of `state_schema_version` placement)
- Unblocks: #497 (`SkillRunner` lifecycle split)
- Builds on: #495 (retired-actor cleanup landed earlier)


src/Aevatar.Foundation.Runtime.Implementations.Orleans/Grains/RuntimeActorGrain.cs
```diff
@@ -128,8 +151,28 @@ public async Task<bool> InitializeAgentAsync(string agentTypeName)
...
+
+        if (_agent != null)
+            return string.Equals(_activeKind, kind, StringComparison.Ordinal);
+
+        var bound = await BindAgentByKindAsync(kind);
+        if (!bound)
+            return false;
+
+        _state.State.AgentId = this.GetPrimaryKeyString();
+        _state.State.Identity = new RuntimeActorIdentity { Kind = kind };
```
eanzhao: **[Consensus: 2 models — mimo, codex]** severity=major, category=bug

Two reviewers flag that `InitializeAgentByKindAsync` writes `Identity.Kind` but does NOT populate the legacy `AgentTypeName` field. During the Phase 1/2 mixed-version rollout this is a real risk: an older pod that picks up such a row sees no `AgentTypeName`, fails its legacy `Type.GetType` lookup, and drops envelopes. Either keep writing a compatible CLR name from the resolved registration, gate the kind-only path behind a feature flag until Phase 3, or document and verify (via deployment ordering) that no legacy pods can observe these rows. At minimum: document the deprecation signal so callers of `GetAgentTypeNameAsync()` know it returns empty for kind-initialized actors.

<details>

- **mimo**: `InitializeAgentByKindAsync` sets `Identity` but does not clear `AgentTypeName`. If a grain was previously initialized via `InitializeAgentAsync(typeName)` and then re-initialized by kind, both fields are populated. This is intentional per the ADR (mixed-version pods), but `IsInitializedAsync()` and `GetAgentTypeNameAsync()` will continue returning stale `AgentTypeName` data. Consider whether callers of `GetAgentTypeNameAsync()` need a deprecation signal or doc note.
- **codex**: `InitializeAgentByKindAsync` persists only `Identity.Kind` and leaves legacy `AgentTypeName` empty. During the Phase 1/2 mixed-version rollout, an old pod that activates this row will see no `AgentTypeName`, treat the actor as uninitialized, and drop envelopes; keep writing the compatible CLR name until Phase 3, or do not expose/use this initializer until legacy pods are gone.
</details>

src/Aevatar.Foundation.Runtime.Implementations.Orleans/Grains/RuntimeActorGrain.cs
```diff
@@ -314,58 +367,160 @@ public async Task PurgeAsync()
...
     }
 
-    private IAgent CreateAgentInstance(Type agentType)
+    /// <summary>
+    /// Strip the assembly-qualifier from a CLR type token so the registry
+    /// can match against <see cref="Type.FullName"/>. Tests + bootstrap
+    /// helpers historically pass <see cref="Type.AssemblyQualifiedName"/>.
+    /// </summary>
+    private static bool TryNormalizeClrTypeName(string clrTypeName, out string normalized)
     {
```
eanzhao: **[Consensus: 2 models — gemini, v4-pro]** severity=major, category=bug

Two reviewers flag that `TryNormalizeClrTypeName` uses `clrTypeName.IndexOf(',')` to strip the assembly qualifier, which mishandles generic assembly-qualified names like `Namespace.Class`1[[T, Assembly, ...]]`. The first comma occurs inside the generic-parameter brackets, producing a truncated/garbage normalized name that won't match `Type.FullName` keys in the registry. Today the `[GAgent]` agents are non-generic so this is latent, but the doc comment claims this path accepts `Type.AssemblyQualifiedName`. Fix: bracket/angle-bracket-aware split, or roundtrip through `Type.GetType(name)?.FullName`.

<details>

- **gemini**: The logic `clrTypeName.IndexOf(',')` is incorrect for generic types with multiple parameters (e.g., `MyAgent<T1, T2>, MyAssembly`). It will split at the first comma inside the generic brackets, resulting in a truncated type name like `MyAgent<T1`. This will cause `TryResolveKindByClrTypeName` to fail for such types. Consider using a logic that skips commas inside balanced brackets or angle brackets.
- **v4-pro**: `TryNormalizeClrTypeName` strips assembly qualifiers by naively splitting on the first comma (`IndexOf(',')`). For assembly-qualified names of generic types (e.g. `Namespace.Class`1[[T, Assembly, ...]], ...`), the comma inside the generic parameter brackets appears before the assembly separator, producing an incorrect normalized name that won't match the registry's `Type.FullName` key. In practice agent classes are non-generic, but the doc comment says this path handles `Type.AssemblyQualifiedName` from tests/bootstrap helpers. Use a bracket-aware split or `Type.GetType(name).FullName` to extract the clean full name.
</details>

src/Aevatar.Foundation.Runtime.Implementations.Orleans/Grains/RuntimeActorIdentity.cs
```diff
@@ -0,0 +1,37 @@
...
+/// stay pure domain artifacts and do not carry a version field — see ADR
+/// 0020 for rationale and the consumer contract from issue #500.
+/// </para>
+/// <para>
+/// <see cref="LegacyClrTypeName"/> is populated only during the Phase 1/2
+/// transition window; once Phase 3 lands and CLR-name identity is removed,
+/// this field becomes <c>reserved</c>.
+/// </para>
+/// </remarks>
+[GenerateSerializer]
```
eanzhao: **[Consensus: 2 models — kimi, codex]** severity=major, category=arch

Two reviewers flag that `RuntimeActorIdentity` is a new persisted state object introduced with Orleans `[GenerateSerializer]` rather than a Protobuf contract, which violates the CLAUDE.md rule: 新增状态/事件/持久化载荷必须先定义 `.proto`. Counterargument: the parent `RuntimeActorGrainState` is already Orleans-serialized and switching it to protobuf is a separate migration. But the explicit rule is *先定义 `.proto`, 再接入实现*, so a new persisted record on the envelope (`StateSchemaVersion`, mixed-version compatibility-critical) is precisely where the Protobuf-first mandate matters. Either define a `.proto` for the identity record (storing protobuf bytes via a typed submessage), or capture this exception explicitly in ADR 0019 / 0020 with a follow-up issue to migrate the entire envelope.

<details>

- **kimi**: `RuntimeActorIdentity` is a new persisted state object introduced without a `.proto` contract, using Orleans `[GenerateSerializer]` instead. CLAUDE.md binding rule: *新增状态对象、事件对象、持久化载荷时，先定义 `.proto` 契约并生成类型，再接入实现；禁止先写临时序列化结构、后补 Protobuf。* Even though `RuntimeActorGrainState` already uses Orleans serialization, new state additions must start with Protobuf. Suggest defining a `.proto` for `RuntimeActorIdentity` (and ultimately migrating `RuntimeActorGrainState`) so the runtime envelope follows the repository-wide serialization mandate.
- **codex**: This adds a new persisted actor identity object using Orleans `[GenerateSerializer]` instead of a protobuf contract, which violates `CLAUDE.md: 统一 Protobuf / 新增状态对象先定义 .proto`. Define `RuntimeActorIdentity` as a protobuf message or store a protobuf-generated typed submessage/bytes in the runtime state envelope before persisting it.
</details>

src/Aevatar.Foundation.Core/TypeSystem/AgentKindRegistry.cs
```diff
@@ -0,0 +1,205 @@
...
+
+        var implClrName = registration.ImplementationType.FullName
+            ?? throw new InvalidOperationException(
+                $"Agent class '{registration.ImplementationType}' has no FullName; cannot register CLR-name lookup.");
+
+        TryAddClrTypeNameLookup(implClrName, registration.Kind);
+        foreach (var legacyClrName in registration.LegacyClrTypeNames)
+            TryAddClrTypeNameLookup(legacyClrName, registration.Kind);
+
+        foreach (var legacyKind in registration.LegacyKinds)
```
eanzhao, line 106: **[codex]** severity=major, category=bug

Legacy kind aliases are not checked against primary kinds, so if one registration has primary kind `a.old` and another claims `[LegacyAgentKind("a.old")]`, `Resolve("a.old")` returns the old primary because `_byKind` is checked first. Reject any legacy alias that collides with an existing primary kind, and reject primary registrations that collide with existing legacy aliases.

src/Aevatar.Foundation.Core/TypeSystem/AgentKindRegistry.cs
```diff
@@ -0,0 +1,205 @@
...
+            StateContractType: StateContractType,
+            Metadata: new AgentImplementationMetadata(
+                Kind: Kind,
+                ImplementationClrTypeName: ImplementationType.FullName ?? ImplementationType.Name,
+                LegacyKinds: LegacyKinds,
+                LegacyClrTypeNames: LegacyClrTypeNames));
+
+    private IAgent CreateInstance(IServiceProvider services)
+    {
+        var instance = ActivatorUtilities.CreateInstance(services, ImplementationType);
```
eanzhao: **[codex]** severity=major, category=di

`AgentImplementation.Factory` resolves agent instances from the singleton registry's captured root `IServiceProvider` (registered at `AgentKindRegistryServiceCollectionExtensions.cs:39`), not the grain activation provider that the old `RuntimeActorGrain` path used. This creates captive scoped dependencies and bypasses Orleans activation scoping; make the factory accept an `IServiceProvider` and call it with `RuntimeActorGrain.ServiceProvider`, or otherwise create an activation scope per bind.

docs/adr/0019-stable-agent-kind-identity.md
```diff
@@ -0,0 +1,172 @@
...
+
+### Attribute surface
+
+- `[GAgent("scheduled.skill-runner")]` — declares the primary kind on an
+  agent class. Single-valued.
+- `[LegacyAgentKind("scheduled.skill-runner")]` — claims a previously-used
+  kind on a new class so persisted state pointing at the old kind resolves
+  to the new implementation without state mutation. Multi-valued.
+- `[LegacyClrTypeName("...")]` (existing, in
+  `Aevatar.Foundation.Abstractions.Compatibility`) — already used by
```
eanzhao, line 109: **[mimo]** severity=minor, category=other

The ADR states the CI guard enforces `^[a-z0-9]+(\.[a-z0-9-]+)+$` but the actual code (both `AgentKindToken.FormatPattern` and `agent_kind_naming_guard.sh` `KIND_REGEX`) uses `^[a-z0-9]+(\.[a-z0-9]+(-[a-z0-9]+)*)+$`. The documented regex would accept `module.-entity` (leading hyphen in non-prefix segment) while the real regex correctly rejects it. Update the ADR to match the actual pattern.

src/Aevatar.Foundation.Core/TypeSystem/AgentKindRegistry.cs
```diff
@@ -0,0 +1,205 @@
...
+
+        foreach (var registration in registrations)
+            Add(registration);
+    }
+
+    public AgentImplementation Resolve(string kind)
+    {
+        ArgumentException.ThrowIfNullOrWhiteSpace(kind);
+
+        if (_byKind.TryGetValue(kind, out var direct))
```
eanzhao: **[gemini]** severity=minor, category=di

The registry captures and uses the `IServiceProvider` from its own construction (typically the root/silo provider). When `RuntimeActorGrain` resolves an agent, it would be safer to pass the grain's own `ServiceProvider` to the factory to ensure that if any grain-scoped services are ever needed via constructor injection, they are available. Currently, `ToImplementation` closes over the registry's provider.

src/Aevatar.Foundation.Core/TypeSystem/AgentKindRegistry.cs
```diff
@@ -0,0 +1,205 @@
...
+    public bool TryGetKind(AgentImplementation implementation, out string kind)
+    {
+        ArgumentNullException.ThrowIfNull(implementation);
+        kind = implementation.Metadata.Kind;
+        return _byKind.ContainsKey(kind);
+    }
+
+    private void Add(AgentRegistration registration)
+    {
+        if (_byKind.TryGetValue(registration.Kind, out var existing))
```
eanzhao: **[v4-pro]** severity=minor, category=design

When adding a `LegacyAgentKind` alias, the `Add` method does not check whether that legacy kind already exists as a *primary* kind in `_byKind`. If two types declare `[GAgent("test.foo")]` and `[GAgent("test.bar")] [LegacyAgentKind("test.foo")]`, `test.foo` resolves to the first type (primary lookup wins in `Resolve`), but `_legacyKindToKind` silently redirects to the second. Consider rejecting this ambiguous configuration at registration time so the conflict is surfaced at startup rather than silently routing to the wrong implementation.

src/Aevatar.Foundation.Core/TypeSystem/AgentKindRegistryBuilder.cs
```diff
@@ -0,0 +1,112 @@
...
+                return;
+
+            throw new InvalidOperationException(
+                $"Duplicate agent kind '{registration.Kind}': '{existing.ImplementationType.FullName}' " +
+                $"and '{registration.ImplementationType.FullName}' both claim it.");
+        }
+
+        _byKind[registration.Kind] = registration;
+    }
+
```
eanzhao, line 99: **[mimo]** severity=minor, category=design

`Build()` returns `_byKind.Values` which is a live `Dictionary.ValueCollection` view, not a snapshot. The doc says "intentionally not reusable" but nothing enforces it. Consider returning `_byKind.Values.ToList()` so mutations to the builder after `Build()` cannot affect the already-constructed `AgentKindRegistry`.

src/Aevatar.Foundation.Core/TypeSystem/AgentKindRegistryServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+    private static AgentKindRegistryBuilder ResolveSharedBuilder(IServiceCollection services)
+    {
+        var existing = services.FirstOrDefault(static descriptor =>
+            descriptor.ServiceType == typeof(AgentKindRegistryBuilder));
+
+        if (existing?.ImplementationInstance is AgentKindRegistryBuilder shared)
+            return shared;
+
+        var builder = new AgentKindRegistryBuilder();
+        services.AddSingleton(builder);
```
eanzhao, line 38: **[v4-pro]** severity=minor, category=di

`ResolveSharedBuilder` calls `services.AddSingleton<IAgentKindRegistry>(sp => ...)` on every invocation instead of `TryAddSingleton`. Each `AddAevatarAgentKindRegistry(...)` call re-registers the factory, relying on "last-wins" DI semantics. The shared builder makes this functionally correct, but it's fragile — if `IServiceCollection` order changes or anything calls `Replace` / `RemoveAll` on `IAgentKindRegistry` between calls, registrations could be lost. Consider registering the factory once with `TryAddSingleton` and using a separate mechanism for `configure` contributions.

src/Aevatar.Foundation.Runtime.Implementations.Orleans/Grains/RuntimeActorGrain.cs
```diff
@@ -118,8 +141,8 @@ public async Task<bool> InitializeAgentAsync(string agentTypeName)
         if (_agent != null)
             return string.Equals(_state.State.AgentTypeName, agentTypeName, StringComparison.Ordinal);
 
-        var initialized = await InitializeAgentInternalAsync(agentTypeName);
-        if (!initialized)
+        var bound = await BindAgentByLegacyClrTypeAsync(agentTypeName);
+        if (!bound)
             return false;
```
eanzhao, line 164: **[glm-5.1]** severity=minor, category=perf

`InitializeAgentAsync` → `BindAgentByLegacyClrTypeAsync` causes **three sequential `WriteStateAsync` calls** for a single activation:

1. `BindAgentAsync` writes (line ~484)
2. `LazyTagIdentityAsync` writes (line ~497)
3. `InitializeAgentAsync` sets `AgentId` + `AgentTypeName` and writes again (lines 153-155)

Each write is a persistent-store round-trip. Consider having `BindAgentByLegacyClrTypeAsync` return the resolved identity info without writing, then let the caller (`InitializeAgentAsync` / `ResumeFromPersistedIdentityAsync`) perform a single consolidated write after all mutations to `_state.State` are complete. The same pattern (2 writes) exists in `InitializeAgentByKindAsync` — consolidating would halve I/O on the kind path too.

src/Aevatar.Foundation.Runtime.Implementations.Orleans/Grains/RuntimeActorGrain.cs
```diff
@@ -128,8 +151,28 @@ public async Task<bool> InitializeAgentAsync(string agentTypeName)
         return true;
     }
 
+    public async Task<bool> InitializeAgentByKindAsync(string kind)
+    {
```
eanzhao, line 176: **[kimi]** severity=minor, category=test

New kind-based grain activation paths (`InitializeAgentByKindAsync`, `ResumeFromPersistedIdentityAsync` kind-first branch, and `GetAgentKindAsync`) are not exercised by any grain-level integration test. The added tests cover registry resolution, DI wiring, and Orleans serialization round-trip, but `RuntimeActorGrain` itself is a critical infrastructure boundary and should have at least one integration test verifying that `InitializeAgentByKindAsync` binds the agent, writes `Identity`, and round-trips through reactivation.

eanzhao, line 176: **[gemini]** severity=minor, category=arch

When `InitializeAgentByKindAsync` is called, it only sets `Identity`. Legacy properties like `AgentTypeName` remain null. While correct for the new model, ensure that existing telemetry or external monitoring that might still be reading `AgentTypeName` via `GetAgentTypeNameAsync` (e.g. in some dashboards or logs) is prepared for empty values for kind-initialized actors.

src/Aevatar.Foundation.Runtime.Implementations.Orleans/Grains/RuntimeActorGrain.cs
```diff
@@ -138,18 +181,24 @@ private async Task HandleEnvelopeAsyncCore(byte[] envelopeBytes, bool propagateF
...
-                if (!initialized || _agent == null)
+                if (!string.IsNullOrWhiteSpace(_state.State.AgentTypeName)
+                    || !string.IsNullOrWhiteSpace(_state.State.Identity?.Kind))
                 {
-                    _logger.LogWarning("Dropping envelope for actor {ActorId}: initialization failed", this.GetPrimaryKeyString());
-                    return;
+                    _logger.LogWarning(
+                        "Dropping envelope for actor {ActorId}: initialization failed",
+                        this.GetPrimaryKeyString());
                 }
```
eanzhao, line 237: **[glm-5.1]** severity=minor, category=arch

`HandleEnvelopeAsyncCore` now unconditionally calls `ResumeFromPersistedIdentityAsync` when `_agent == null`, even though `OnActivateAsync` already called it. If the registry couldn't resolve the identity at activation time (e.g. missing module registration), **every inbound envelope** will re-attempt the full resolution path (registry lookup, possible reflection scan, possibly another `WriteStateAsync`). Consider adding a lightweight `_activationAttempted` flag or a short-circuit after the first failed resolution so that persistent misconfigurations don't amplify into per-envelope I/O.

src/Aevatar.Foundation.Runtime.Implementations.Orleans/Grains/RuntimeActorGrain.cs
```diff
@@ -314,58 +367,160 @@ public async Task PurgeAsync()
...
+        var registry = ServiceProvider?.GetService<IAgentKindRegistry>();
+        var legacyResolver = ServiceProvider?.GetService<ILegacyAgentClrTypeResolver>();
+
+        // Prefer kind resolution: if a registered class lists this CLR full
+        // name (current Type.FullName or [LegacyClrTypeName]), bind by kind
+        // and lazy-tag Identity.Kind so subsequent activations skip this lane.
+        if (registry != null && TryNormalizeClrTypeName(clrTypeName, out var normalizedClrName) &&
+            registry.TryResolveKindByClrTypeName(normalizedClrName, out var resolvedKind))
+        {
+            var implementation = registry.Resolve(resolvedKind);
```
eanzhao: **[mimo]** severity=minor, category=bug

`registry.Resolve(resolvedKind)` can throw `UnknownAgentKindException` but is not wrapped in try-catch here, unlike the same call in `BindAgentByKindAsync` (lines 389–402) which does catch it. Although `TryResolveKindByClrTypeName` returning a kind that `Resolve` can't find is extremely unlikely (registry is immutable after construction), the inconsistency means an unexpected failure here would propagate as an unhandled exception and drop the envelope. Wrap in the same try-catch pattern for consistency.

src/Aevatar.Foundation.Runtime.Implementations.Orleans/Grains/RuntimeActorGrain.cs
```diff
@@ -314,58 +367,160 @@ public async Task PurgeAsync()
...
+            if (await BindAgentAsync(implementation, ct))
+            {
+                await LazyTagIdentityAsync(resolvedKind, clrTypeName);
```
eanzhao: **[mimo]** severity=minor, category=perf

`LazyTagIdentityAsync` calls `_state.WriteStateAsync()` inside `BindAgentByLegacyClrTypeAsync`. When called from `InitializeAgentAsync`, there's a second `_state.WriteStateAsync()` at line 150. Two persistent-state writes per initialization for the lazy-tag path. Consider deferring the lazy-tag write or batching — e.g., set `Identity` in-memory and let the caller's `WriteStateAsync()` persist it.

test/Aevatar.Foundation.Core.Tests/TypeSystem/AgentKindRegistryTests.cs
```diff
@@ -0,0 +1,163 @@
+using Aevatar.Foundation.Abstractions;
```
eanzhao, line 1: **[mimo]** severity=minor, category=test

No test covers the `ILegacyAgentClrTypeResolver` fallback path (un-decorated class → `ReflectionLegacyAgentClrTypeResolver.TryResolve` → agent activates with `Kind = ""`). The existing tests cover registry resolution and token validation but the transitional reflection fallback — the main backward-compat safety net — is untested. Add a test with a non-`[GAgent]` `IAgent` class to verify the fallback resolves correctly and `GetAgentKindAsync()` returns empty.

tools/ci/agent_kind_naming_guard.sh
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env bash
+
+# Issue #498 acceptance: every [GAgent("...")] / [LegacyAgentKind("...")]
+# token must match ^[a-z0-9]+(\.[a-z0-9-]+)+$ and must NOT end with -v\d+.
```
eanzhao: **[mimo]** severity=minor, category=other

Comment says `^[a-z0-9]+(\.[a-z0-9-]+)+$` but `KIND_REGEX` on line 15 is `^[a-z0-9]+(\.[a-z0-9]+(-[a-z0-9]+)*)+$`. The comment pattern would allow leading hyphens in segments (e.g. `module.-entity`); the actual code correctly rejects them. Sync the comment with the variable.

tools/ci/agent_kind_naming_guard.sh
```diff
@@ -0,0 +1,59 @@
...
+  local reason="$4"
+  echo "::error file=${file},line=${line}::Agent kind '${kind}' invalid: ${reason}"
+  violations=$((violations + 1))
+}
+
+# Files that may declare kinds: any C# under src/, agents/, tools/ excluding tests.
+SEARCH_PATHS=("src" "agents" "tools/Aevatar.Tools.Cli")
+
+# Walk every [GAgent("...")] and [LegacyAgentKind("...")] occurrence.
+attribute_pattern='\[(GAgent|LegacyAgentKind)\("([^"]+)"\)\]'
```
eanzhao: **[kimi]** severity=minor, category=bug

The `rg --no-heading -n` invocation outputs one line per matching line, not per match. If a source line contains both `[GAgent("a.b")]` and `[LegacyAgentKind("c.d")]`, `rg` emits the line once and the `sed` extraction (greedy `.*`) only validates the last matched kind token, silently missing the first. Fix: use `rg -o --no-heading -n` so each match is emitted separately, or adjust the loop to parse all occurrences on the line.

src/Aevatar.Foundation.Abstractions/TypeSystem/AgentKindToken.cs
```diff
@@ -0,0 +1,53 @@
...
+        ArgumentException.ThrowIfNullOrWhiteSpace(kind, parameterName);
+
+        if (VersionedTail.IsMatch(kind))
+            throw new ArgumentException(
+                $"Agent kind '{kind}' is forbidden: kinds are never versioned. " +
+                "Schema evolution within a kind goes through proto3 field rules or state-version migration, not through kind rename.",
+                parameterName);
+
+        if (!Format.IsMatch(kind))
+            throw new ArgumentException(
```
eanzhao, line 49: **[gemini]** severity=nit, category=style

Including the raw regex `FormatPattern` in the exception message can be confusing for end users due to double backslashes and technical syntax. Consider using the provided example '(e.g. "scheduled.skill-runner")' as the primary guidance in the message.

src/Aevatar.Foundation.Core/TypeSystem/AgentKindRegistry.cs
```diff
@@ -0,0 +1,205 @@
...
+
+        _services = services;
+        _byKind = new Dictionary<string, AgentRegistration>(StringComparer.Ordinal);
+        _legacyKindToKind = new Dictionary<string, string>(StringComparer.Ordinal);
+        _clrTypeNameToKind = new Dictionary<string, string>(StringComparer.Ordinal);
+
+        foreach (var registration in registrations)
+            Add(registration);
+    }
+
```
eanzhao, line 45: **[glm-5.1]** severity=nit, category=design

`Resolve(string)` calls `registration.ToImplementation(_services)` which creates a new `AgentImplementation` record + `Func<IAgent>` closure on **every activation**. Since the registry is read-only after construction, consider caching the `AgentImplementation` per registration at build time (the `IServiceProvider` reference is stable). Minor allocation saving on the activation hot path.

src/Aevatar.Foundation.Core/TypeSystem/AgentKindRegistry.cs
```diff
@@ -0,0 +1,205 @@
...
+            .ToArray();
+
+        var stateContract = ResolveStateContract(agentType);
+
+        return new AgentRegistration(
+            Kind: gAgent.Kind,
+            ImplementationType: agentType,
+            StateContractType: stateContract,
+            LegacyKinds: legacyKinds,
+            LegacyClrTypeNames: legacyClrTypeNames);
```
eanzhao, line 222: **[gemini]** severity=nit, category=perf

The loop `for (var current = agentType; current != null; current = current.BaseType)` is redundant because `current.GetInterfaces()` already returns all interfaces in the entire inheritance hierarchy. You can simply call `agentType.GetInterfaces()` once and search for `IAgent<>`.

src/Aevatar.Foundation.Runtime.Implementations.Orleans/Grains/RuntimeActorGrain.cs
```diff
@@ -128,8 +151,28 @@ public async Task<bool> InitializeAgentAsync(string agentTypeName)
...
+        if (_agent != null)
+            return string.Equals(_activeKind, kind, StringComparison.Ordinal);
+
+        var bound = await BindAgentByKindAsync(kind);
+        if (!bound)
+            return false;
+
+        _state.State.AgentId = this.GetPrimaryKeyString();
+        _state.State.Identity = new RuntimeActorIdentity { Kind = kind };
+        await _state.WriteStateAsync();
```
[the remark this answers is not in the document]
eanzhao: Fixed in d8245144: `InitializeAgentByKindAsync` now backfills `AgentTypeName` from `implementation.Metadata.ImplementationClrTypeName`, so older runtime pods can still activate kind-initialized rows through their legacy `Type.GetType` lookup during the Phase 1/2 window.

src/Aevatar.Foundation.Runtime.Implementations.Orleans/Grains/RuntimeActorGrain.cs
```diff
@@ -314,58 +367,160 @@ public async Task PurgeAsync()
...
+                kind,
+                this.GetPrimaryKeyString());
+            return false;
+        }
+
+        var bound = await BindAgentAsync(implementation, ct);
+        if (!bound)
+            return false;
+
+        _activeKind = kind;
```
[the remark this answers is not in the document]
eanzhao: Fixed in d8245144: `BindAgentByKindAsync` now records `implementation.Metadata.Kind` (canonical) as `_activeKind`, and `InitializeAgentByKindAsync` persists that into `Identity.Kind` instead of the caller's input. Idempotency on repeat calls is handled by `KindResolvesToActiveImplementation`, which resolves the input through the registry to compare against the canonical.

src/Aevatar.Foundation.Runtime.Implementations.Orleans/Grains/RuntimeActorGrain.cs
eanzhao: Fixed in d8245144: `InitializeAgentByKindAsync` now writes both `Identity.Kind` (canonical) and `AgentTypeName` (from `implementation.Metadata.ImplementationClrTypeName`). Older pods that only read `AgentTypeName` continue to activate the row through the legacy CLR-name path until Phase 3 deprecation.

src/Aevatar.Foundation.Runtime.Implementations.Orleans/Grains/RuntimeActorGrain.cs
eanzhao: Fixed in d8245144: `TryNormalizeClrTypeName` is now bracket-depth-aware — commas inside `[...]` are skipped so generic AQNs like `Foo\`1[[T, ParamAsm]], OuterAsm` produce `Foo\`1[[T, ParamAsm]]` (matchable against `Type.FullName`). Tests cover non-generic, single-generic, multi-generic, and nested-generic AQNs.

src/Aevatar.Foundation.Runtime.Implementations.Orleans/Grains/RuntimeActorIdentity.cs
eanzhao: Documented in ADR 0019 (commit d8245144): the Orleans-serialization choice is now an explicit waiver. `RuntimeActorIdentity` lives inside the existing Orleans-serialized `RuntimeActorGrainState` envelope; switching it alone to Protobuf would require dual-encoding or migrating the whole envelope. Phase 1 keeps the envelope homogeneous; full Protobuf migration of `RuntimeActorGrainState` is a follow-up that must land before any further state field is added.

src/Aevatar.Foundation.Core/TypeSystem/AgentKindRegistry.cs
eanzhao: Fixed in d8245144: registry build is now two-pass. Pass 1 collects all primary kinds; pass 2 validates legacy aliases against the full primary set and throws `InvalidOperationException` when a `[LegacyAgentKind]` collides with an existing `[GAgent]` token. New test `Build_RejectsLegacyAliasCollidingWithExistingPrimaryKind` covers this.

src/Aevatar.Foundation.Core/TypeSystem/AgentKindRegistry.cs
eanzhao: Fixed in d8245144: `AgentImplementation.Factory` now takes `Func<IServiceProvider, IAgent>` and the grain passes its own `ServiceProvider` at activation time, so grain-scoped dependencies bind through the grain's container rather than the silo root. Test `Factory_UsesCallerSuppliedServiceProvider_NotRegistryCaptureTime` verifies this directly.

docs/adr/0019-stable-agent-kind-identity.md
eanzhao: Verified in d8245144: ADR 0019 line 85 already shows the strict regex `^[a-z0-9]+(\.[a-z0-9]+(-[a-z0-9]+)*)+$`. The CI guard's *header comment* (`agent_kind_naming_guard.sh:4`) was the loose form — synced to match `KIND_REGEX` in this commit.

src/Aevatar.Foundation.Core/TypeSystem/AgentKindRegistry.cs
eanzhao: Fixed in d8245144: `AgentImplementation.Factory` is now `Func<IServiceProvider, IAgent>`; the registry no longer captures a provider, and the grain passes its activation-time provider at call time. `ToImplementation` is replaced by `BuildImplementation` (no provider parameter) to make the no-capture stance explicit.

src/Aevatar.Foundation.Core/TypeSystem/AgentKindRegistry.cs
eanzhao: Fixed in d8245144: see `Build_RejectsLegacyAliasCollidingWithExistingPrimaryKind` — registration now rejects `[GAgent(\"a.bar\")] [LegacyAgentKind(\"a.foo\")]` when another type already owns `a.foo` as its primary.

src/Aevatar.Foundation.Core/TypeSystem/AgentKindRegistryBuilder.cs
eanzhao: Fixed in d8245144: `AgentKindRegistryBuilder.Build()` now returns `_byKind.Values.ToList()` (defensive snapshot). Test `Build_RejectsBuilderSnapshotMutationsAfterBuild` asserts post-Build mutations don't leak.

src/Aevatar.Foundation.Core/TypeSystem/AgentKindRegistryServiceCollectionExtensions.cs
eanzhao: Fixed in d8245144: `AddAevatarAgentKindRegistry` now uses `TryAddSingleton` for the registry factory. The shared builder still accumulates contributions, but the factory itself only registers once, so DI ordering is stable across module extension calls.

src/Aevatar.Foundation.Runtime.Implementations.Orleans/Grains/RuntimeActorGrain.cs
eanzhao: Fixed in d8245144: `BindAgentByLegacyClrTypeAsync` no longer writes state itself. It returns a `LegacyBindOutcome` enum and mutates `Identity` in-memory only; the caller (`ResumeFromPersistedIdentityAsync` / `InitializeAgentAsync`) owns a single `WriteStateAsync` covering both legacy fields and the lazy-tagged `Identity`. Three writes → one.

src/Aevatar.Foundation.Runtime.Implementations.Orleans/Grains/RuntimeActorGrain.cs
eanzhao: Added in d8245144: `RuntimeActorIdentityResolutionTests` exercises the extracted activation-resolution helpers directly (kind matching across legacy aliases, generic-AQN normalization). Combined with the existing `AgentKindIdentityIntegrationTests` (DI wiring + Orleans round-trip), this covers the activation logic without an Orleans test cluster, which is structurally hard to set up at the grain level. Full grain-cluster smoke is deferred to PR #497's mixed-version test.

eanzhao: Fixed in d8245144: `InitializeAgentByKindAsync` now also writes `AgentTypeName` from `implementation.Metadata.ImplementationClrTypeName`, so `GetAgentTypeNameAsync` keeps returning a meaningful value for telemetry / dashboards. Phase 3 hard-deprecation removes the field; until then, kind-initialized actors expose both.

src/Aevatar.Foundation.Runtime.Implementations.Orleans/Grains/RuntimeActorGrain.cs
eanzhao: Fixed in d8245144: `_identityResolutionAttempted` flag is set in `OnActivateAsync` after the first attempt; `HandleEnvelopeAsyncCore` short-circuits when the flag is true. Persistent misconfigurations now drop envelopes with a single log line each, no per-envelope registry probe / reflection scan / state write. `PurgeAsync` resets the flag so a re-bound actor can be re-resolved.

src/Aevatar.Foundation.Runtime.Implementations.Orleans/Grains/RuntimeActorGrain.cs
eanzhao: Fixed in d8245144: `registry.Resolve` inside `BindAgentByLegacyClrTypeAsync` is now wrapped in the same try/catch as `BindAgentByKindAsync`. The defensive symmetry guards against future registry bugs (registry returns a kind that `Resolve` then can't find — an internal-corruption case).

src/Aevatar.Foundation.Runtime.Implementations.Orleans/Grains/RuntimeActorGrain.cs
eanzhao: Fixed in d8245144: `LazyTagIdentityAsync` is renamed to `ApplyLazyIdentityTagInMemory` and no longer writes; `Identity` is mutated in-memory and the caller persists once. Resume path: 1 write (or 0 when there's nothing to lazy-tag). Initialize path: 1 write covering both legacy fields and the lazy-tag.

test/Aevatar.Foundation.Core.Tests/TypeSystem/AgentKindRegistryTests.cs
eanzhao: Added in d8245144: `ReflectionLegacyAgentClrTypeResolverTests` covers the un-decorated fallback path: full-name + AQN resolution, missing-type rejection, non-IAgent rejection, empty-input rejection, and constructor-failure surface.

tools/ci/agent_kind_naming_guard.sh
eanzhao: Fixed in d8245144: `agent_kind_naming_guard.sh` header comment now reads `^[a-z0-9]+(\.[a-z0-9]+(-[a-z0-9]+)*)+$` to match `KIND_REGEX` exactly.

tools/ci/agent_kind_naming_guard.sh
eanzhao: Fixed in d8245144: `agent_kind_naming_guard.sh` now uses `rg -o --no-heading -n`. Each match is emitted as its own line, so `[GAgent(\"a.b\")]` and `[LegacyAgentKind(\"c.d\")]` on the same source line are validated independently.

src/Aevatar.Foundation.Abstractions/TypeSystem/AgentKindToken.cs
eanzhao: Fixed in d8245144: `AgentKindToken.Validate` error message replaces the raw regex with a plain-language format description plus examples (`'scheduled.skill-runner'`, `'channels.bot-registration'`).

src/Aevatar.Foundation.Core/TypeSystem/AgentKindRegistry.cs
eanzhao: Fixed in d8245144: `AgentKindRegistry` now pre-builds `AgentImplementation` per registration at construction time and stores `Dictionary<string, AgentImplementation> _implByKind`. `Resolve` does a single dictionary lookup with no allocations on the activation hot path.

src/Aevatar.Foundation.Core/TypeSystem/AgentKindRegistry.cs
eanzhao: Fixed in d8245144: `ResolveStateContract` now does one `agentType.GetInterfaces()` scan; the inheritance-chain loop was redundant (`Type.GetInterfaces()` already returns the full set across the hierarchy).

src/Aevatar.Foundation.Runtime.Implementations.Orleans/Grains/RuntimeActorIdentity.cs
eanzhao: Re-fixed in 7b148d27 — went the proto route instead of the waiver.

`RuntimeActorIdentity` is now a real Protobuf message defined at `src/Aevatar.Foundation.Abstractions/runtime_actor_identity.proto`:
```proto
message RuntimeActorIdentity {
  string kind = 1;
  int32 state_schema_version = 2;
  string legacy_clr_type_name = 3;
}
```

The Orleans `[GenerateSerializer]` POCO is deleted; the proto-generated class lives in `Aevatar.Foundation.Abstractions.Runtime` and is serialized by Orleans through `AddProtobufSerializer` (already wired in `AddAevatarFoundationRuntimeOrleans`). `RuntimeActorGrainState.Identity` references the proto type directly; round-trip tests pass with the existing Orleans serializer.

The runtime envelope (`RuntimeActorGrainState` itself) stays Orleans-serialized — that's the runtime-owned outer record (CLR type names, parent/child topology, opaque agent-state bytes), not a domain state, and the existing fields predate this PR. ADR 0019 updated to reflect the actual decision (proto-typed sub-message embedded in the Orleans envelope), the earlier waiver text is gone.

src/Aevatar.Foundation.Runtime.Implementations.Orleans/Grains/RuntimeActorGrain.cs
eanzhao: Added in 7b148d27: `AgentKindGrainActivationIntegrationTests` boots a real Orleans silo (same harness as `OrleansDirectDispatchFailurePropagationTests`) and exercises `RuntimeActorGrain` end-to-end. Five tests cover:

- `InitializeAgentByKindAsync` binds via the registry, persists `Identity.Kind` (canonical), mirrors `AgentTypeName` for mixed-version pods.
- `InitializeAgentByKindAsync` with a `[LegacyAgentKind]` alias persists the canonical kind, not the deprecated token (P2 verification at the grain level).
- `ResumeFromPersistedIdentityAsync` reactivates correctly when a second grain reference picks up the persisted row.
- `InitializeAgentAsync` (legacy CLR-name path) lazy-tags `Identity.Kind` while preserving `AgentTypeName`.
- Unknown kind returns false and leaves the grain uninitialized.

Test setup is in-memory persistence + in-memory streams so the silo bootstraps in ~1 second per test. All five pass on the latest commit.

eanzhao:
## Multi-model fix-check (PR #535)

Re-ran `/opencode-pr-fix-check` after commits `d8245144` + `7b148d27` were pushed in response to the 22 inline comments from the [earlier multi-model review](https://github.com/aevatarAI/aevatar/pull/535#discussion_r3165789715). Six independent reviewers (`v4-pro`, `glm-5.1`, `mimo`, `kimi`, `gemini`, `codex`) classified each comment against the post-fix diff.

### Verdict

| Status | Count |
|---|---|
| ✅ Resolved | 22 / 22 |
| ⚠️ Partial | 0 |
| ❌ Unresolved | 0 |
| ❓ Cannot determine | 0 |

19 comments unanimous (6/6); 3 had a single dissenter that I fact-checked against the current files and found incorrect.

### Highlights of fixes verified

- **Mixed-version compatibility (#1)** — `AgentTypeName` now mirrored when initializing by kind, so older pods can still resolve activation.
- **Generic-type AQN (#2)** — `TryNormalizeClrTypeName` now bracket-aware (handles `Class\`1[[T, Asm, ...]], Asm`).
- **Protobuf-first identity (#3)** — `RuntimeActorIdentity` migrated to `runtime_actor_identity.proto`.
- **Alias collision guard (#4, #8)** — registry now rejects legacy alias that collides with another primary kind.
- **DI scoping (#5, #7)** — `IServiceProvider` now passed at resolve-time, not captured at construction.
- **Write consolidation (#11, #16)** — multiple `WriteStateAsync` calls collapsed into a single per-activation write.
- **Resolver fallback test coverage (#17)** — `ReflectionLegacyAgentClrTypeResolverTests.cs` covers full-name, AQN, unknown, non-agent, and empty-input cases.
- **CI guard accuracy (#18, #19)** — header comment now matches `KIND_REGEX`; `rg` pattern handles multi-attribute lines.

### Two outstanding gaps the panel missed (manual review)

While the fix-check panel reached "resolved" on all 22, two real issues remain in the fix commits that the panel did not flag:

1. **P2 — test does not prove agent re-binding** (`test/Aevatar.Foundation.Runtime.Hosting.Tests/AgentKindGrainActivationIntegrationTests.cs:87-111`). `ResumeFromPersistedIdentity_ReactivatesByKindOnSecondGrainLookup` only asserts `IsInitializedAsync()` and `GetAgentKindAsync()`, both of which return from persisted `_state.State` alone. If `ResumeFromPersistedIdentityAsync` set `Identity` but failed to bind `_agent`, the test would still pass. Suggest adding an assertion that exercises the live agent (e.g. `GetDescriptionAsync()` returning the fixture description) to prove `_agent` was re-bound.

2. **P3 — ADR 0020 contradicts the implementation** (`docs/adr/0020-actor-state-version-placement.md:41-49`). The ADR still shows `RuntimeActorIdentity` as a `[GenerateSerializer]` C# class with `[Id]` fields, but the fix moved it to `runtime_actor_identity.proto`. ADR 0019 documents the Protobuf migration; ADR 0020 should be updated to match (either reference the `.proto` directly or remove the stale C# block).

### Next steps

LGTM on the original 22 — the implementation work is solid. The two gaps above are minor (a test strengthening + a doc sync) and can be addressed in a follow-up commit before merge.

---
*Generated via `/opencode-pr-fix-check` — artifacts at `/tmp/opencode-pr-fix-check/aevatarAI-aevatar/535/`*

eanzhao:
Fixed both gaps in `10b60fef`:

**P2 (test strengthening)** — `ResumeFromPersistedIdentity_ReactivatesByKindOnSecondGrainLookup` now also asserts `GetDescriptionAsync()` returns `"IntegrationFixtureCanonicalAgent"`. The grain forwards `GetDescriptionAsync` to the live `_agent` instance (returns `"Uninitialized:..."` fallback when `_agent == null`), so a regression where `Identity` is set but `_agent` isn't re-bound now actually fails the test.

**P3 (ADR 0020 doc sync)** — Updated the embedded code block to show the Protobuf message instead of the deleted Orleans `[GenerateSerializer]` class, with a note about the `AddProtobufSerializer` wiring. ADR 0019 already documents the full migration.

PR #540, opened 2026-04-30
[codex] Expose NyxID spec catalog readiness

eanzhao:
## Summary

- Surface `NyxIdSpecCatalog` status, including initial refresh, in-progress state, last refresh failure kind, and successful refresh timestamp.
- Add a critical Mainnet readiness contributor named `nyxid-catalog`; it fails readiness for missing/rejected spec token or an empty initial catalog, while transient NyxID/network fetch failures stay ready and wait for the next refresh.
- Preserve the last valid spec catalog when a later refresh yields no operations, document that stale-catalog soft-failure contract, and serialize refresh status updates behind a single lock with overlapping refresh ticks skipped.
- Record bounded `spec_catalog_lookup_miss_total{reason=unknown_operation}` metrics while keeping attempted operation ids in logs.
- Document `AEVATAR_Aevatar__NyxId__SpecFetchToken` / `Aevatar__NyxId__SpecFetchToken` injection, deployment smoke check, and stale catalog behavior in the Mainnet host README.

## Impact

Missing production token wiring and invalid tokens are visible through readiness, warning logs, and health details. Specialized NyxID tools still work without the catalog; generic discovery/execution paths expose catalog misses without unbounded metric cardinality.

Transient NyxID or network outages during startup no longer make the host not-ready; the health component reports `lastRefreshError` / `lastRefreshFailureKind` and the background refresh updates the catalog after recovery. If a later refresh returns an empty spec after a valid catalog was loaded, the host keeps serving the last catalog and exposes `lastRefreshFailureKind=EmptySpec`.

This repo does not contain the production Helm/Secret manifest, so the real `SpecFetchToken` value still has to be injected by the deployment platform.

## Validation

- `dotnet test test/Aevatar.AI.Tests/Aevatar.AI.Tests.csproj --filter NyxIdSpecCatalogTests --nologo`
- `dotnet test test/Aevatar.Hosting.Tests/Aevatar.Hosting.Tests.csproj --filter MainnetNyxIdCatalogHealthTests --nologo`

Refs #538


docs/adr/0002-mainnet-architecture.md
```diff
@@ -1616,6 +1616,7 @@ flowchart LR
 | `AEVATAR_ActorRuntime__KafkaTopicName` | Kafka Topic 名称 |
 | `AEVATAR_ActorRuntime__KafkaConsumerGroup` | Kafka 消费组 |
 | `AEVATAR_Projection__Graph__Providers__Neo4j__Password` | Mainnet Host 连接 Neo4j 的密码 |
+| `AEVATAR_Aevatar__NyxId__SpecFetchToken` | NyxID OpenAPI spec catalog 拉取 token；缺失时 `/health/ready` 的 `nyxid-catalog` 组件不通过 |
```
eanzhao: **[codex]** severity=major, category=arch

This edits an existing ADR, which violates `CLAUDE.md: docs/adr immutable` (`docs/adr/` may only be superseded by new decisions). Move this deployment env-var note to an allowed mutable location such as the Mainnet README/canon docs, or add a new superseding ADR if this is an architecture decision.

src/Aevatar.AI.ToolProviders.NyxId/NyxIdSpecCatalog.cs
```diff
@@ -28,8 +30,9 @@ public NyxIdSpecCatalog(
...
         {
             // NyxID's /api/v1/docs/openapi.json is human-only; without a token
-            // every fetch returns 401. Skip the timer to avoid 30-min noise.
-            _logger.LogInformation(
+            // every fetch returns 401. Skip the timer and surface this through
+            // readiness so production does not run with an invisible empty catalog.
+            _logger.LogWarning(
                 "NyxIdSpecCatalog: SpecFetchToken not configured; skipping background refresh, catalog will remain empty");
             return;
         }
```
eanzhao, line 45: **[gemini]** severity=major, category=bug

The `dueTime` for the refresh timer is set to 30 minutes, and the `InitialFetchAsync` call at line 38 has a short 5-second timeout with no retries. If the first fetch fails (e.g., transient network issue during startup), the `nyxid-catalog` health check (now marked `Critical` in `MainnetHostBuilderExtensions.cs`) will remain `Unhealthy`, causing the service to stay in a not-ready state for 30 minutes until the timer fires. Suggest using `TimeSpan.Zero` for the timer's `dueTime` and moving the initial fetch logic into `RefreshAsync` to leverage its existing retry logic and longer timeout.

src/Aevatar.AI.ToolProviders.NyxId/NyxIdToolProviderMetrics.cs
```diff
@@ -0,0 +1,25 @@
...
+
+    public static void RecordSpecCatalogLookupMiss(string? operationId)
+    {
+        var normalizedOperationId = string.IsNullOrWhiteSpace(operationId)
+            ? "unknown"
+            : operationId.Trim();
+
+        SpecCatalogLookupMisses.Add(
+            1,
+            new KeyValuePair<string, object?>("operation_id", normalizedOperationId));
```
eanzhao: **[codex]** severity=major, category=perf

`operation_id` is copied directly from user input on lookup misses, so an authenticated caller can create unbounded metric label cardinality by sending random operation IDs. Keep the attempted operation ID in logs, but make the metric label bounded, for example `reason=empty|unknown|invalid` or remove the label from the counter.

src/Aevatar.Mainnet.Host.Api/Hosting/MainnetHostBuilderExtensions.cs
```diff
@@ -139,4 +141,64 @@ public static WebApplication MapAevatarMainnetHost(this WebApplication app)
...
+                }
+
+                if (!status.SpecFetchTokenConfigured)
+                {
+                    return ValueTask.FromResult(AevatarHealthContributorResult.Unhealthy(
+                        "NyxID spec fetch token is missing; generic capability discovery is unavailable.",
+                        details));
+                }
+
+                if (status.OperationCount == 0)
```
eanzhao: **[kimi]** severity=major, category=design

The critical health probe returns `Unhealthy` when `status.OperationCount == 0`, but `NyxIdSpecCatalog` performs its initial fetch via fire-and-forget (`_ = InitialFetchAsync()`). During startup, readiness probes will see `OperationCount == 0` and return 503 before the first fetch completes, causing unnecessary transient failures. Distinguish "still loading" from "genuinely empty" by adding an `InitialFetchAttempted` flag to `NyxIdSpecCatalogStatus` and only returning unhealthy after the first fetch attempt has finished (successfully or not).

src/Aevatar.AI.ToolProviders.NyxId/NyxIdSpecCatalog.cs
```diff
@@ -12,6 +12,8 @@ public sealed class NyxIdSpecCatalog : IDisposable
     private readonly Timer? _refreshTimer;
 
     private OperationCard[] _catalog = [];
```
eanzhao, line 18: **[gemini]** severity=minor, category=concurrency

On 32-bit platforms, `Volatile.Read`/`Write` on a `long` field (`_lastSuccessfulRefreshUnixTimeSeconds`) is not atomic and can result in torn reads. While likely running on 64-bit in production, using `Interlocked.Read` and `Interlocked.Exchange` is safer for 64-bit values shared across threads.

src/Aevatar.AI.ToolProviders.NyxId/NyxIdSpecCatalog.cs
```diff
@@ -128,13 +148,24 @@ private async Task FetchAndUpdateAsync(CancellationToken ct)
 
         if (cards.Length == 0)
         {
+            Volatile.Write(ref _lastRefreshError, "Spec yielded no operations.");
```
eanzhao: **[kimi]** severity=minor, category=bug

When `cards.Length == 0`, `_lastRefreshError` is set but `_lastSuccessfulRefreshUnixTimeSeconds` is not reset. If a prior refresh succeeded, `GetStatus()` will show both `OperationCount > 0` (old catalog) and `LastRefreshError != null`, which is inconsistent. Consider clearing the old catalog or at least documenting that an empty spec response is treated as a soft error that preserves the previous catalog.

src/Aevatar.AI.ToolProviders.NyxId/NyxIdSpecCatalog.cs
```diff
@@ -128,13 +148,24 @@ private async Task FetchAndUpdateAsync(CancellationToken ct)
...
             _logger.LogWarning("NyxIdSpecCatalog: spec yielded no operations, skipping update");
             return;
         }
 
         Volatile.Write(ref _catalog, cards);
```
eanzhao: **[v4-pro]** severity=minor, category=concurrency

`_catalog` is written before `_lastSuccessfulRefreshUnixTimeSeconds` in `FetchAndUpdateAsync` (lines 156–158), but `GetStatus()` (line 49) reads `_lastSuccessfulRefreshUnixTimeSeconds` first, then `Operations.Length`. A concurrent reader can observe `OperationCount > 0` alongside a still-null `LastSuccessfulRefreshUtc`. The health check is unaffected (it only gates on `OperationCount == 0`), but consumers of `GetStatus()` for observability may see a transient inconsistent snapshot. Reorder the writes so `_lastSuccessfulRefreshUnixTimeSeconds` is set before `_catalog`, or snapshot both fields under a single lock/interlocked operation.

test/Aevatar.Hosting.Tests/MainnetNyxIdCatalogHealthTests.cs
```diff
@@ -0,0 +1,38 @@
...
+using Aevatar.Mainnet.Host.Api.Hosting;
+using FluentAssertions;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Aevatar.Hosting.Tests;
+
+public sealed class MainnetNyxIdCatalogHealthTests
+{
+    [Fact]
+    public async Task NyxIdCatalogHealthContributor_WhenSpecFetchTokenMissing_ShouldBeUnhealthy()
```
eanzhao: **[mimo-v2.5-pro]** severity=minor, category=test

Only the unhealthy path (token missing) is tested. Add a test for the happy path: `BaseUrlConfigured=true`, `SpecFetchTokenConfigured=true`, `OperationCount>0` → returns `Healthy` with the "loaded N operations" message. This validates the final branch in `CreateNyxIdCatalogHealthContributor`.

src/Aevatar.Mainnet.Host.Api/Hosting/MainnetHostBuilderExtensions.cs
```diff
@@ -139,4 +141,64 @@ public static WebApplication MapAevatarMainnetHost(this WebApplication app)
...
         return app;
     }
+
+    internal static AevatarHealthContributorRegistration CreateNyxIdCatalogHealthContributor() =>
+        new()
+        {
+            Name = "nyxid-catalog",
+            Category = "dependency",
+            Critical = true,
+            ProbeAsync = static (serviceProvider, _) =>
```
eanzhao: **[gemini]** severity=nit, category=arch

The health probe uses `static` but then captures `BuildNyxIdCatalogHealthDetails`. While the closure is fine, consider if this contributor should be its own class in `Infrastructure` or `ToolProviders` rather than an internal method in the Host layer to keep the composition root cleaner, following the 'Core Mandates' on layer separation.

eanzhao:
## 🤖 Multi-model fix-check (`/opencode-pr-fix-check`)

Reviewers: kimi, gemini, codex (3/6 — others hung in CLI: v4-pro, glm-5.1, mimo-v2.5-pro)

**⚠️ partial: 1 | ✅ resolved: 9**

### ⚠️ Partially addressed

- [`src/Aevatar.AI.ToolProviders.NyxId/NyxIdSpecCatalog.cs:151`](https://github.com/aevatarAI/aevatar/pull/540#discussion_r3166605131) (2/3 agree)
  - **kimi**: The author added `NyxIdSpecCatalogRefreshFailureKind.EmptySpec` and a log message stating 'preserving existing catalog', but the old catalog is still preserved while `_lastRefreshError` is set. There is no code comment or design note documenting this intentional inconsistency for consumers of `GetStatus()`.
  - **gemini**: The implementation now logs that it preserves the existing catalog on empty responses, and the state remains consistent under a lock even when partial refresh failures occur.
  - **codex**: The code now intentionally preserves the existing catalog on an empty spec and logs 'preserving existing catalog', which addresses the likely operational intent. However GetStatus can still report OperationCount > 0 with LastRefreshError/EmptySpec from a later failed refresh, and the preservation semantics are not explicitly documented outside the log/status behavior.

---
<sub>Generated by `/opencode-pr-fix-check`. Each model independently classified every prior inline comment as resolved/partial/unresolved/cannot-determine; the verdict above is the majority view (conservative tiebreak).</sub>

PR #541, opened 2026-04-30
fix(channel-identity): hardcode public OAuth callback URL, detect redirect-URI drift

eanzhao:
## Summary

Production incident on `aismart-app-mainnet` 2026-04-30: every `/init`'s authorize URL was unreachable. Pod log confirmed DCR succeeded, but the registered `redirect_uri` was `http://+:8080/api/oauth/nyxid-callback` — Kestrel's wildcard listen address from `ASPNETCORE_URLS`, not a browser-routable URL.

## Root cause

`NyxIdRedirectUriResolver` had a priority chain that read `ASPNETCORE_URLS` / `IConfiguration[ServerUrls]` as fallbacks. In K8s those are pod-internal listen addresses, never valid OAuth callback targets. The resolver conflated "what does Kestrel bind to" with "what's the public callback URL" — different concepts in any non-trivial deployment.

## Resolver fix

- Mirror `NyxIdAuthorityResolver`: **hardcode the production default** (`https://aevatar-console-backend-api.aevatar.ai`). Production is zero-config — no env var to set.
- Optional `AEVATAR_OAUTH_REDIRECT_BASE_URL` env-var override for staging / dev / test.
- **Drop `ASPNETCORE_URLS` / `IConfiguration[ServerUrls]` from the priority chain entirely** — the prod incident proves they're never safe.
- Defense-in-depth: a misconfigured override pointing at a wildcard host (`+`, `*`, `0.0.0.0`, `[::]`) is rejected + warning logged + fall back to default.

## Drift detection (heals existing prod state)

The existing prod cluster's actor state already has a `client_id` registered at NyxID with the wrong `redirect_uri`. Just deploying the resolver fix wouldn't heal it — `HandleEnsureProvisioned` short-circuits when authority + client_id match, so the actor would never re-register.

Fix:
- Persist `redirect_uri` on `AevatarOAuthClientState` + `AevatarOAuthClientProvisionedEvent` + `AevatarOAuthClientDocument`.
- `HandleEnsureProvisioned` re-runs DCR when stored `redirect_uri` drifts from the resolver's current output. Empty stored value (legacy event-store snapshots that predate the field) is treated as match-anything to avoid spurious re-DCR on first redeploy of clusters that were already correct.
- Bootstrap pre-flight applies the same comparison so the actor's command actually fires when drift is detected.
- `/api/oauth/aevatar-client/status` now exposes `redirect_uri_registered` / `redirect_uri_resolved` / `redirect_uri_drifted` for ops triage.

## Cleanup

`IConfiguration` removed from `NyxIdRemoteCapabilityBroker` and `AevatarOAuthClientBootstrapService` ctors — no longer needed.

## Test plan

- [x] 7 new `NyxIdRedirectUriResolverTests` (default, override priority, override trim, 4 wildcard variants rejected, empty override, ASPNETCORE_URLS isolation).
- [x] 2 new `HandleEnsureProvisioned` tests: drift triggers re-DCR (different client_id, state.redirect_uri updated), matching redirect_uri does not.
- [x] 818 ChannelRuntime tests green.
- [ ] After deploy: bootstrap log shows `redirect URI drifted` → DCR re-runs → new client_id, correct redirect_uri at NyxID.
- [ ] `/init` from Lark returns an authorize URL pointing at `https://aevatar-console-backend-api.aevatar.ai/api/oauth/nyxid-callback` — link is clickable + reachable.
- [ ] `GET /api/oauth/aevatar-client/status` reports `redirect_uri_drifted=false` after the heal.

## Operator action

**None.** Bootstrap detects drift on cold boot, re-DCRs against NyxID with the corrected callback, projection materializes the new `client_id`, next `/init` works.

### DIFF
--- a/agents/Aevatar.GAgents.Channel.Identity.Abstractions/IAevatarOAuthClientProvider.cs
+++ b/agents/Aevatar.GAgents.Channel.Identity.Abstractions/IAevatarOAuthClientProvider.cs
@@ -46,7 +46,8 @@ public sealed record AevatarOAuthClientSnapshot(
     DateTimeOffset? BrokerCapabilityObservedAt,
     string? PreviousHmacKid = null,
     byte[]? PreviousHmacKey = null,
-    DateTimeOffset? PreviousHmacDemotedAt = null);
+    DateTimeOffset? PreviousHmacDemotedAt = null,
+    string? RedirectUri = null);
 
 /// <summary>
 /// Thrown when an OAuth flow tries to use the cluster client before the
--- a/agents/Aevatar.GAgents.Channel.Identity/Broker/NyxIdRemoteCapabilityBroker.cs
+++ b/agents/Aevatar.GAgents.Channel.Identity/Broker/NyxIdRemoteCapabilityBroker.cs
@@ -3,7 +3,6 @@ using System.Text.Json;
 using System.Text.Json.Serialization;
 using Aevatar.GAgents.Channel.Abstractions;
 using Aevatar.GAgents.Channel.Identity.Abstractions;
-using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 
@@ -51,7 +50,6 @@ public sealed class NyxIdRemoteCapabilityBroker : INyxIdCapabilityBroker, INyxId
     private readonly StateTokenCodec _stateTokenCodec;
     private readonly IExternalIdentityBindingQueryPort _queryPort;
     private readonly TimeProvider _timeProvider;
-    private readonly IConfiguration? _configuration;
     private readonly ILogger<NyxIdRemoteCapabilityBroker> _logger;
 
     public NyxIdRemoteCapabilityBroker(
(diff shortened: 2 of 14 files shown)


eanzhao:
P0: this does not heal the existing production client registered with `http://+:8080`.

The drift path depends on `snapshot.RedirectUri` / `State.RedirectUri`, but all existing prod `AevatarOAuthClientProvisionedEvent` records predate the new `redirect_uri` field. After deployment those actors/readmodels will surface an empty stored redirect URI. Both pre-flight and actor-side checks explicitly treat empty as "match anything":

- bootstrap: `RedirectUriDrifted(cached.RedirectUri, redirectUri)` returns false when stored is empty, so a cached `client_id` + same authority returns before dispatching EnsureProvisioned.
- actor: `redirectUriDrifted` also requires `!string.IsNullOrEmpty(State.RedirectUri)`, so even if a command reaches the actor, legacy state will not re-DCR.

That means the exact aismart-app-mainnet state described in the PR body keeps the old NyxID client whose registered callback is `http://+:8080/api/oauth/nyxid-callback`; `/init` will continue to generate an authorize flow backed by the unusable client. To actually repair prod, legacy empty `redirect_uri` must be treated as drift/unknown for this recovery path, or there must be another actor-owned forced re-DCR path keyed by the resolved public redirect URI.

eanzhao:
已在 e4ef02e6 修复前面指出的问题，并一起改了 Lark bot 未绑定提示：

- legacy `redirect_uri` 为空现在会被当作 unknown/drift，bootstrap 和 actor 侧都会触发一次 re-DCR，避免继续复用注册了 `http://+:8080/api/oauth/nyxid-callback` 的旧 NyxID client。
- `/api/oauth/status` 的 `redirect_uri_drifted` 对 legacy 空值也会返回 true，状态诊断和修复语义一致。
- Lark 私聊未绑定时直接发送 NyxID 绑定卡片，不再要求用户手动输入 `/init`；群聊场景只提示去私聊点击绑定卡片，不暴露登录 URL。

验证：
- `bash tools/ci/projection_state_version_guard.sh`
- `bash tools/ci/projection_state_mirror_current_state_guard.sh`

Issue #547, opened 2026-04-30
Make Studio team command ACK semantics honest


louis4li:
## Problem

StudioTeamService.UpdateAsync and ArchiveAsync dispatch commands and immediately read the eventually consistent team read model to return a post-state snapshot. That can return stale data or incorrectly surface not-found after a successful dispatch.

## Why it matters

ACK semantics should only promise stages that are actually guaranteed. Returning a post-state requires an explicit observed-version/readmodel readiness contract; otherwise the endpoint should return an accepted receipt.

## Acceptance criteria

- Team update/archive endpoints return an honest accepted receipt, or implement a clear observed-version/readmodel readiness contract.
- No endpoint implies read-model freshness after a weak dispatch ACK.
- Tests cover stale readmodel after dispatch.

## Source

Follow-up from StudioMember async bind PR review; out of scope for the binding-run PR.

Issue #543, opened 2026-04-30
Fix NyxID OAuth binding projection activation


louis4li:
## Problem

NyxID OAuth callback can create/update ExternalIdentityBindingGAgent and wait for the binding read model, but the projection path is not clearly activated/host-composed for this actor. The result can be a committed binding fact that remains unobservable by IExternalIdentityBindingQueryPort.ResolveAsync.

## Current status

Latest `dev` includes explicit projection activation for the cluster-singleton `AevatarOAuthClientGAgent` bootstrap path. This issue remains open for the per-subject `ExternalIdentityBindingGAgent` OAuth binding path, which still needs an explicit projection activation/composition contract before callback writes rely on the binding read model.

## Why it matters

Committed actor facts must enter the unified projection/observation pipeline. OAuth binding should not depend on query fallback or incidental host composition.

## Acceptance criteria

- ExternalIdentityBindingGAgent has an explicit projection kind/activation path or an equivalent committed-state projection activation contract.
- AddChannelIdentityProjection / MapIdentityOAuthEndpoints composition is verified in the relevant host.
- OAuth callback observes binding through the projection read model only after the projection path is activated.
- Add regression coverage for callback -> committed event -> observable binding read model.

## Source

Follow-up from StudioMember async bind PR review; out of scope for the binding-run PR.

PR #548, opened 2026-04-30
Add async StudioMember binding protocol

### LINKED ISSUE #516: Refactor StudioMember binding admission into actor-owned async protocol
## Context

Follow-up from #462 and PR #506.

PR #506 intentionally keeps the current implementation scope small after review: it addresses the NyxID relay admission/routing items, but does **not** implement the larger StudioMember binding lifecycle refactor. The earlier implementation attempt touched Studio actor state/events, Studio application contracts, projection continuations, endpoints, frontend, and tests, and was reverted because the architecture impact is large enough to require a separate design and review.

## Problem

`StudioMemberService.BindAsync` still depends on the StudioMember read model before issuing the bind command. If the authoritative StudioMember actor has committed creation but the projection/read model lags, bind can fail with `STUDIO_MEMBER_NOT_FOUND` even though the member exists.

This is the remaining #462 item after the relay scope work:

- write/admission path depends on projection read-model freshness;
- binding orchestration boundaries need to follow CLAUDE.md actor/command/event rules;
- projection must remain a committed-fact materialization pipeline, not a second business orchestration framework.

## Design Direction

Before implementation, propose and review a plan that covers:

- an actor-owned StudioMember bind admission/command protocol;
- how bind request, progress, completion, and failure are represented as durable actor facts/events;
- whether a short-lived binding-run actor is needed for cross-actor scope binding steps;
- how cross-actor waits are continuation/eventized instead of blocking an actor turn;
- how read models expose binding status without becoming the authoritative source;
- API/frontend contract changes, if any, and migration/removal of obsolete synchronous paths.

## Acceptance Criteria

- `BindAsync` no longer rejects valid members because the StudioMember read model is stale.
- Business orchestration is not executed inside projection materialization turns.
- Message delivery goes through the standard dispatch/command contracts, not direct stream/provider/actor calls.
- Current-state read models remain projection copies of committed actor facts.
- Design document is reviewed before broad code changes begin.
- Focused tests cover stale read-model admission, async completion/failure, and retry/duplicate handling.

## Out of Scope for PR #506

This issue is intentionally **not** implemented in PR #506 to keep that PR reviewable and avoid mixing relay admission cleanup with a larger Studio binding lifecycle redesign.

louis4li:
## Summary

Implements the StudioMember async binding protocol from issue #516.

- Adds StudioMember actor-owned binding run admission and execution state.
- Moves bind API semantics to an honest accepted receipt plus run status polling.
- Adds frontend polling/refresh handling for accepted binding runs.
- Removes process-local platform binding execution from `StartAsync`; the run actor now drives platform binding through self continuation and can recover pending runs from persisted state.
- Adds a binding-run current-state read model/query port so the status endpoint reads the run actor's authority-shaped projection instead of member readmodel summaries.
- Keeps platform binding completion asynchronous: `ExecuteAsync` now returns after starting the platform work, and completion/failure flows back through typed continuation events.
- Promotes GAgent endpoint kind from an internal string to a typed StudioMember proto enum before mapping to GAgent Service endpoint kind.

## Scope Notes

This PR intentionally keeps the fix scoped to StudioMember async binding correctness. It does not actorize or redesign the broader legacy ScopeBinding command service.

Adjacent architecture issues are tracked separately in #543-#547 and are not prerequisites for this async binding protocol unless a later change explicitly wires them into this flow.

Follow-up #616 tracks replacing `ScopeBindingCommandApplicationService` readmodel visibility polling with an explicit readiness/observation contract. That cleanup is derivative architecture debt; PR #548 is functional without expanding into it.

Failure codes remain open strings for now because current control flow branches on typed run status, while failure code is only diagnostic/display data. If retry/API behavior starts branching by reason, we should promote the reason to a typed enum plus detail fields.

## Smoke Test

Ran `Aevatar.Mainnet.Host.Api` locally on `http://127.0.0.1:5101` with auth disabled and in-memory runtime/projection providers.

- `GET /api/health` returned `200` ready.
- `POST /api/scopes/{scopeId}/members` returned `201`.
- `PUT /api/scopes/{scopeId}/members/{memberId}/binding` returned accepted with binding run `bind-d0f71c089c944c88be8276e65050f300`.
- `GET /api/scopes/{scopeId}/members/{memberId}/binding-runs/{bindingRunId}` reached `status: "succeeded"`.
- `GET /api/scopes/{scopeId}/members/{memberId}/binding` returned `lastBinding.revisionId: "rev-smoke-gagent-1"`.
- `GET /api/scopes/{scopeId}/members/{memberId}/endpoints/chat/contract` returned `200`.

## Validation

- `dotnet build src/Aevatar.Mainnet.Host.Api/Aevatar.Mainnet.Host.Api.csproj --nologo` — passed with existing warnings.
- `dotnet test test/Aevatar.Studio.Tests/Aevatar.Studio.Tests.csproj --no-restore --nologo --filter "ScopeBindingStudioMemberPlatformBindingCommandServiceTests|ActorDispatchStudioMemberCommandServiceTests|StudioMemberBindingRunGAgentStateTests|StudioMemberServiceBindingTests"` — 37 passed.
- `dotnet test test/Aevatar.Studio.Tests/Aevatar.Studio.Tests.csproj --no-restore --nologo` — 473 passed.
- `bash tools/ci/test_stability_guards.sh` — passed.
- `bash tools/ci/architecture_guards.sh` — passed.

## Related

- Closes #516
- Follow-up: #616


apps/aevatar-console-web/src/shared/studio/api.ts
```diff
@@ -1205,8 +1210,162 @@ function decodeStudioMemberBindingContract(
...
+}
+
+function decodeStudioMemberBindingRunStatusResponse(
+  value: unknown
+): StudioMemberBindingRunStatusResponse {
+  const record = expectRecord(value, "StudioMemberBindingRunStatusResponse");
+  return {
+    status: normalizeStudioMemberBindingRunStatus(
+      readOptionalScalar(record, ["status", "Status"])
+    ),
```
eanzhao, line 1279: **[codex]** severity=blocker, category=bug

`readOptionalScalar` is used here but is not imported from `@/shared/api/http/decoders`, and the current decoder module does not define it. This will fail the web build; either add/export that decoder helper or use an existing helper/inline lookup for the optional `status` field.

agents/Aevatar.GAgents.StudioMember/StudioMemberBindingRunGAgent.cs
```diff
@@ -0,0 +1,642 @@
...
+    [EventHandler(EndpointName = "executePlatformBinding", AllowSelfHandling = true)]
+    public async Task HandlePlatformBindingExecuteRequested(StudioMemberPlatformBindingExecuteRequested evt)
+    {
+        if (!CanAcceptPlatformBindingCommand(evt.BindingRunId, evt.PlatformBindingCommandId))
+            return;
+
+        if (State.PlatformExecutionInFlight && !evt.RecoveryExecution)
+            return;
+
+        await PersistDomainEventAsync(new StudioMemberPlatformBindingExecutionStarted
```
eanzhao, line 143: **[kimi]** severity=major, category=di

`Services.GetService<IStudioMemberPlatformBindingCommandPort>()` is a service-locator anti-pattern. CLAUDE.md forbids `GetService<T>` outside framework integration code. The existing `StudioMemberGAgent` in the same directory does not use this pattern; the new run actor should receive the port via constructor injection like other DI-friendly grains.

apps/aevatar-console-web/src/pages/studio/index.tsx
```diff
@@ -3843,56 +3871,144 @@ const StudioPage: React.FC = () => {
...
+
+      for (let attempt = 0; attempt < MEMBER_BINDING_RUN_POLL_ATTEMPTS; attempt += 1) {
+        if (attempt > 0) {
+          await waitForStudioMemberBindingRunTick();
+        }
+
+        try {
+          latestRun = await studioApi.getMemberBindingRun(
+            receipt.scopeId,
+            receipt.memberId,
```
eanzhao, line 4175: **[codex]** severity=major, category=bug

This catch treats every `getMemberBindingRun` failure as projection lag. A 401/403/500 or malformed response will be silently swallowed for all retry attempts and then surfaced as a non-terminal accepted bind, hiding real failures from the user. Only swallow the expected short-lived 404/read-model-lag case, and rethrow other `StudioApiError` statuses.

src/Aevatar.Studio.Projection/CommandServices/ActorDispatchStudioMemberCommandService.cs
```diff
@@ -292,6 +297,78 @@ private static StudioMemberImplementationRef BuildImplementationRefMessage(
...
+                }
+                break;
+            default:
+                throw new InvalidOperationException(
+                    $"Unknown implementationKind '{implementationKindName}'.");
+        }
+
+        request.RequestHash = ComputeRequestHash(request);
+        return request;
+    }
```
eanzhao, line 355: **[Consensus: 2 models]** severity=major, category=bug

Both reviewers flag that `ParseGAgentEndpointKind` maps null/empty (i.e. the absent endpoint kind) to `StudioMemberGAgentEndpointKind.Command`, which silently assigns command semantics to an unspecified value. The proto already defines `STUDIO_MEMBER_GAGENT_ENDPOINT_KIND_UNSPECIFIED` — null/empty should map to it (or be rejected), and only the explicit "command"/"chat" strings should resolve to their respective kinds.

<details>

- **glm-5.1**: `ParseGAgentEndpointKind` defaults `null` or empty string to `StudioMemberGAgentEndpointKind.Command`. A missing endpoint kind should map to `Unspecified` — silently assigning `Command` semantics to absent data can produce incorrect routing behavior that's hard to diagnose. Suggest:

```csharp
"command" => StudioMemberGAgentEndpointKind.Command,
"chat" => StudioMemberGAgentEndpointKind.Chat,
_ => StudioMemberGAgentEndpointKind.Unspecified,
```

or throw for truly unrecognized values while mapping null/empty to `Unspecified`.
- **codex**: Defaulting a missing/empty GAgent endpoint kind to `Command` changes the API semantics: an omitted kind now creates a command endpoint instead of preserving `Unspecified`/validation failure. Since the proto has `STUDIO_MEMBER_GAGENT_ENDPOINT_KIND_UNSPECIFIED`, map null/empty to `Unspecified` or reject it explicitly before dispatching.

</details>

src/Aevatar.Studio.Projection/CommandServices/ScopeBindingStudioMemberPlatformBindingCommandService.cs
```diff
@@ -0,0 +1,309 @@
...
+            CancellationToken.None);
+        return Task.CompletedTask;
+    }
+
+    private async Task RunBindingAsync(
+        string replyActorId,
+        string commandId,
+        StudioMemberPlatformBindingStartRequested request,
+        CancellationToken ct)
+    {
```
eanzhao, line 91: **[Consensus: 2 models]** severity=major, category=concurrency

`ExecuteAsync(CancellationToken ct, ...)` accepts a token from the caller but threads `CancellationToken.None` into both `Task.Run` and `RunBindingAsync`, so the caller's abort signal (e.g. host shutdown / actor deactivation) cannot stop the in-flight binding work. Either pass `ct` (or a linked token) through, or drop the parameter from the interface so callers don't expect cancellation support that doesn't exist.

<details>

- **glm-5.1**: `ExecuteAsync` accepts a `CancellationToken ct` parameter but completely ignores it, passing `CancellationToken.None` to both `Task.Run` and `RunBindingAsync`. If the caller's token is cancelled (e.g. host shutdown), the platform binding work continues uncancelled. Either use `ct` (or a linked timeout token) inside `RunBindingAsync`, or remove the parameter from the interface to avoid misleading callers into thinking cancellation is honored.
- **kimi**: `ExecuteAsync` accepts a `CancellationToken ct` but passes `CancellationToken.None` to both `Task.Run` and `RunBindingAsync`. This prevents cooperative cancellation and silently drops the caller's abort signal. Pass `ct` through: `_ = Task.Run(() => RunBindingAsync(..., ct), ct);`.

</details>

src/Aevatar.Studio.Hosting/Endpoints/StudioMemberEndpoints.cs
```diff
@@ -148,7 +152,10 @@ internal static async Task<IResult> HandleBindAsync(
 
         try
```
eanzhao, line 153: **[deepseek-v4-pro]** severity=minor, category=bug

`StudioMemberNotFoundException` is dead code here. The new `BindAsync` no longer queries `_memberQueryPort` — it normalizes input and dispatches a binding run, then returns 202. A missing member is now surfaced asynchronously via the binding run status, so this catch will never fire. Remove or replace with a catch for exceptions `StartBindingRunAsync` can actually throw.

src/Aevatar.Studio.Hosting/Endpoints/StudioMemberEndpoints.cs
```diff
@@ -192,6 +198,35 @@ internal static async Task<IResult> HandleGetBindingAsync(
...
+        {
+            return Results.Ok(await memberService.GetBindingRunAsync(scopeId, memberId, bindingRunId, ct));
+        }
+        catch (StudioMemberNotFoundException ex)
+        {
+            return NotFound(ex);
+        }
+        catch (StudioMemberBindingRunNotFoundException ex)
+        {
+            return BindingRunNotFound(ex);
```
eanzhao, line 214: **[deepseek-v4-pro]** severity=minor, category=bug

`StudioMemberNotFoundException` is dead code here. `GetBindingRunAsync` only queries `_bindingRunQueryPort`, not `_memberQueryPort`. The only domain exception it can throw is `StudioMemberBindingRunNotFoundException` (already caught below). Remove this dead catch clause.

src/Aevatar.Studio.Projection/CommandServices/ScopeBindingStudioMemberPlatformBindingCommandService.cs
```diff
@@ -0,0 +1,309 @@
...
+    }
+
+    public Task ExecuteAsync(
+        string replyActorId,
+        string platformBindingCommandId,
+        StudioMemberPlatformBindingStartRequested request,
+        CancellationToken ct = default)
+    {
+        ArgumentException.ThrowIfNullOrWhiteSpace(replyActorId);
+        ArgumentException.ThrowIfNullOrWhiteSpace(platformBindingCommandId);
```
eanzhao, line 55: **[kimi]** severity=minor, category=design

`StartAsync` duplicates the fallback command-id format (`platform-{bindingRunId}-1`) already defined in `StudioMemberBindingRunGAgent.BuildPlatformBindingCommandId`. If either side changes, the run actor and the platform port will disagree on the canonical id. Extract a shared helper (e.g., `StudioMemberPlatformBindingCommandId.Format`) so the convention lives in one place.

src/Aevatar.Studio.Projection/CommandServices/ScopeBindingStudioMemberPlatformBindingCommandService.cs
```diff
@@ -0,0 +1,309 @@
...
+    private async Task RunBindingAsync(
+        string replyActorId,
+        string commandId,
+        StudioMemberPlatformBindingStartRequested request,
+        CancellationToken ct)
+    {
+        ScopeBindingUpsertResult result;
+        try
+        {
+            result = await _scopeBindingCommandPort
```
eanzhao, line 95: **[glm-5.1]** severity=minor, category=concurrency

The fire-and-forget `_ = Task.Run(...)` discards the task reference. While `RunBindingAsync` has comprehensive try/catch blocks, if an unhandled path ever throws (e.g. `ArgumentNullException` from `request.Clone()` on a malformed recovery input), the exception becomes an unobserved task exception on the finalizer. Consider wrapping in a top-level try/catch or capturing the task for observation:

```csharp
var _ = RunBindingFireAndForget(replyActorId, commandId, executionRequest);

private async Task RunBindingFireAndForget(...) { try { await RunBindingAsync(...); } catch (Exception ex) { _logger.LogError(ex, "..."); } }
```

apps/aevatar-console-web/src/shared/studio/api.ts
potter-sun: Fixed in e9127f13. `readOptionalScalar` is now defined locally in `apps/aevatar-console-web/src/shared/studio/api.ts`, so it no longer depends on a missing decoder export. Verified with `npm --prefix apps/aevatar-console-web run tsc`.

apps/aevatar-console-web/src/pages/studio/index.tsx
potter-sun: Fixed in e9127f13. `waitForMemberBindingRun` now only treats `StudioApiError` 404 as short-lived read-model lag; all other errors are rethrown so auth/server/decoder failures are surfaced instead of being reported as a non-terminal accepted bind. Covered by frontend tests and verified with `npm --prefix apps/aevatar-console-web run tsc`.

agents/Aevatar.GAgents.StudioMember/StudioMemberBindingRunGAgent.cs
louis4li: Fixed in 80a3f3d9. `StudioMemberBindingRunGAgent` now receives `IStudioMemberPlatformBindingCommandPort` via constructor injection and no longer calls `Services.GetService<T>()` inside the actor. Verified with the targeted StudioMember test set.

src/Aevatar.Studio.Projection/CommandServices/ActorDispatchStudioMemberCommandService.cs
louis4li: Fixed in 80a3f3d9. Missing/empty GAgent endpoint `kind` now fails fast instead of silently defaulting to `Command`; only explicit `command` and `chat` values are accepted. Added `StartBindingRunAsync_ShouldRejectMissingGAgentEndpointKind` coverage.

src/Aevatar.Studio.Projection/CommandServices/ScopeBindingStudioMemberPlatformBindingCommandService.cs
louis4li: Fixed in 80a3f3d9. The detached `ExecuteAsync` contract no longer accepts a misleading `CancellationToken`; the implementation starts fire-and-forget binding work explicitly with `CancellationToken.None`, so callers are not promised request-cancellation behavior that cannot be honored after dispatch.

src/Aevatar.Studio.Hosting/Endpoints/StudioMemberEndpoints.cs
louis4li: Fixed in 80a3f3d9. Removed the dead `StudioMemberNotFoundException` catch from the bind endpoint and removed the corresponding obsolete endpoint test; missing members are now represented asynchronously through the binding run result.

src/Aevatar.Studio.Hosting/Endpoints/StudioMemberEndpoints.cs
louis4li: Fixed in 80a3f3d9. Removed the dead `StudioMemberNotFoundException` catch from `GetBindingRunAsync`; the endpoint now keeps the actual `StudioMemberBindingRunNotFoundException` path for typed 404s.

src/Aevatar.Studio.Projection/CommandServices/ScopeBindingStudioMemberPlatformBindingCommandService.cs
louis4li: Fixed in 80a3f3d9. Extracted the platform binding command-id convention to `StudioMemberConventions.BuildPlatformBindingCommandId(...)` and reused it from both the run actor and platform binding command service.

src/Aevatar.Studio.Projection/CommandServices/ScopeBindingStudioMemberPlatformBindingCommandService.cs
louis4li: Fixed in 80a3f3d9. Wrapped the detached binding execution in `RunBindingFireAndForgetAsync` with a top-level `try/catch` and structured error logging, so unexpected failures are observed instead of becoming unobserved task exceptions.

eanzhao (approved):
Nothing to post.
LGTM

agents/Aevatar.GAgents.StudioMember/StudioMemberBindingRunGAgent.cs
```diff
@@ -0,0 +1,644 @@
...
+        }
+
+        await PersistDomainEventAsync(evt);
+        await SendAdmissionRequestAsync();
+    }
+
+    [EventHandler(EndpointName = "admitBindingRun")]
+    public async Task HandleAdmitted(StudioMemberBindingAdmittedEvent evt)
+    {
+        if (!CanAcceptRunEvent(evt.BindingRunId))
```
jason-aelf: Admission handling needs to be idempotent here. Right now a repeated admission for the same `bindingRunId` is accepted in any non-terminal state. If the run is already in `PlatformBindingPending`, `ApplyAdmitted` regresses the state back to `Admitted`, then this handler sends another platform start and generates a new `PlatformBindingCommandId`.

That can cause the result of the first platform command to be ignored because the command id no longer matches the current state.

Please only accept admission for the `AdmissionPending -> Admitted` transition. Repeated admission with the same admitted snapshot should be a no-op after that. Also add a test that duplicate admission does not change status, command id, or attempt count.

agents/Aevatar.GAgents.StudioMember/StudioMemberBindingRunGAgent.cs
```diff
@@ -0,0 +1,644 @@
...
+            case StudioMemberBindingRunStatus.AdmissionPending:
+                await SendAdmissionRequestAsync(ct);
+                break;
+            case StudioMemberBindingRunStatus.Admitted:
+                await SendPlatformBindingStartRequestedAsync(State.UpdatedAtUtc, ct);
+                break;
+            case StudioMemberBindingRunStatus.PlatformBindingPending:
+                await SendPlatformBindingPendingAndExecuteAsync(
+                    State.UpdatedAtUtc ?? Timestamp.FromDateTimeOffset(DateTimeOffset.UtcNow),
+                    recoveryExecution: true,
```
jason-aelf: Activation recovery should not always re-execute platform binding. This path schedules `recoveryExecution: true`, and the execute handler uses that flag to bypass the `PlatformExecutionInFlight` guard.

If the actor is reactivated while a fresh platform binding execution is still in flight, for example after host recycle, this can trigger the external side effect twice.

Please check `PlatformExecutionInFlight && !IsPlatformExecutionStale()` during activation. Fresh in-flight execution should only restore the watchdog or wait for the result; recovery execution should only happen when there is no in-flight execution or the execution is stale. Add tests for both fresh in-flight and stale in-flight activation.

agents/Aevatar.GAgents.StudioMember/StudioMemberBindingRunGAgent.cs
louis4li: Fixed in 547372a5. Admission is now accepted only for the `AdmissionPending -> Admitted` transition; duplicate admission after platform binding has started is a no-op, so it does not regress status or rotate `PlatformBindingCommandId` / attempt count.

Added coverage: `DuplicateAdmission_AfterPlatformBindingPending_ShouldNotRegressOrRotateCommandId`.

agents/Aevatar.GAgents.StudioMember/StudioMemberBindingRunGAgent.cs
louis4li: Fixed in 547372a5. Activation recovery now matches the watchdog semantics: fresh `PlatformExecutionInFlight` only restores the watchdog, while stale or missing in-flight execution schedules a `RecoveryExecution=true` execute for the same platform command id.

Added coverage: `ActivateAsync_WhenInFlightIsFresh_ShouldOnlyRestoreWatchdog` and `ActivateAsync_WhenInFlightIsStale_ShouldScheduleRecoveryExecute`.

agents/Aevatar.GAgents.StudioMember/StudioMemberBindingRunGAgent.cs
louis4li: Follow-up pushed in dde60f79. While checking the duplicate-admission fix, I also tightened the adjacent platform-start transition so `StudioMemberPlatformBindingStartRequested` is only accepted from `Admitted`; repeated/late start messages after `PlatformBindingPending` no longer restart platform binding, rotate command id, or increment attempt count.

Added coverage: `DuplicatePlatformBindingStart_AfterPlatformBindingPending_ShouldNotRegressOrIncrementAttempt` and `HandlePlatformBindingStartRequested_AfterPlatformBindingPending_ShouldNotRestartPlatformBinding`.

agents/Aevatar.GAgents.StudioMember/StudioMemberBindingRunGAgent.cs
louis4li: Follow-up pushed in dde60f79. I also guarded the adjacent accepted-event path: `StudioMemberPlatformBindingAccepted` now has to match the current platform command id and is ignored once `PlatformExecutionInFlight` is true, so a repeated/late accepted event cannot clear in-flight state or schedule another execute after activation/watchdog recovery has already started execution.

Added coverage: `DuplicatePlatformBindingAccepted_AfterExecutionStarted_ShouldNotClearInFlight` and `HandlePlatformBindingAccepted_AfterExecutionStarted_ShouldNotRescheduleExecute`.

potter-sun (approved):

Issue #549, opened 2026-04-30
AevatarOAuthClientGAgent rolling-deploy race wedges /init on every deploy (single-pod prod, OCC v3→v4)


eanzhao:
## Symptom (production, aismart-app-mainnet, 2026-04-30)

Every cold deploy of `aevatar-console-backend` leaves the well-known `aevatar-oauth-client` actor wedged. The /init Lark card's authorize URL fails with NyxID `invalid_redirect_uri`, and the actor's persisted state never reflects the latest DCR result.

Pod log on a fresh deploy:

```
warn: AevatarOAuthClientGAgent[0] received unrecognised event type Google.Protobuf.WellKnownTypes.Any (×3)
... Actor aevatar-oauth-client (AevatarOAuthClientGAgent) created via Orleans runtime
... Aevatar OAuth client EnsureProvisioned dispatched
... drift warning: stored='' resolved='https://aevatar-console-backend-api.aevatar.ai/api/oauth/nyxid-callback'
... [DCR HTTP call begins, succeeds → client_id=861dfa51-...]
... Event sourcing commit hit optimistic concurrency conflict; agentId=projection.durable.scope:aevatar-oauth-client:aevatar-oauth-client eventType=ProjectionScopeWatermarkAdvancedEvent expectedVersion=3 actualVersion=4 droppedFromPending=1
... Runtime envelope handling failed after retry exhausted (or retry disabled) for actor projection.durable.scope:aevatar-oauth-client:..., envelope d6dcd18..., event type 'aevatar.CommittedStateEventPublished'.
... Registered aevatar OAuth client at NyxID: client_id=861dfa51-d2de-4ca7-9d5f-29914c09d042
... Event sourcing commit hit optimistic concurrency conflict; agentId=aevatar-oauth-client eventType=AevatarOAuthClientProvisionedEvent expectedVersion=3 actualVersion=4 droppedFromPending=1
fail: AevatarOAuthClientGAgent[0] at HandleEnsureProvisioned ... line 165
```

Status endpoint shows the OLD wildcard-redirect-URI client is still authoritative:

```
GET /api/oauth/aevatar-client/status
{
  "client_id": "c7f5cf02-c87b-45b9-86af-7e52baf7f987",
  "redirect_uri_registered": null,
  "redirect_uri_resolved": "https://aevatar-console-backend-api.aevatar.ai/api/oauth/nyxid-callback",
  "redirect_uri_drifted": true,
  ...
}
```

[...]

PR #552, opened 2026-04-30
Fix drifted NyxID OAuth client recovery

eanzhao:
## Problem
Production /init can stay broken on every deploy: `/api/oauth/aevatar-client/status` reports a client with `redirect_uri_registered = null` and `redirect_uri_drifted = true` while the Lark card's authorize URL keeps pointing at the stale client id, and NyxID rejects the flow with `invalid_redirect_uri`.

Root cause is a multi-layer issue tracked in issue #549:

1. **Bootstrap declared success too early**: returned as soon as the actor command was published to the stream. If the actor handler / projection failed afterwards, no signal made it back and the cluster kept serving the broken readmodel.
2. **Broker still issued authorize URLs against drifted snapshots**: even when the runtime knew the redirect URI was stale, the broker built and returned the authorize URL anyway, which NyxID rejected.
3. **Status endpoint masked drift as `ready`**: the boolean check on broker observation hid the drift state from ops triage.
4. **Actor died on optimistic concurrency** instead of absorbing it — when a peer pod won the rolling-deploy race the loser propagated the OCC and the bootstrap retry burned another orphan client at NyxID on every backoff attempt.
5. **Cosmetic noise**: every persisted event triggered a "unrecognised event type Any" warning at activation replay because the actor's pre-check did not account for the framework's Any wrapping.

## Solution
- Bootstrap keeps retrying until the OAuth client readmodel actually observes a non-drifted `redirect_uri` (`WaitForProvisionedReadModelAsync` polling).
- `NyxIdRemoteCapabilityBroker` now refuses to issue authorize URLs or exchange codes while the snapshot redirect URI is missing or drifted; callers surface this as the existing "initializing NyxID client, retry later" path.
- Status endpoint reports `redirect_uri_drifted` instead of `ready` when stored ≠ resolved.
- `AevatarOAuthClientGAgent.HandleEnsureProvisioned` catches `EventStoreOptimisticConcurrencyException` at the drift-branch DCR commit (and the follow-up HMAC seed), refreshes State from the store via a new `GAgentBase<TState>.RefreshStateFromStoreAsync` helper, and absorbs the race as a no-op when the peer's commit already matches the command's intended (authority, redirect_uri, hmac-seeded) shape. The orphan `client_id` is logged so ops can clean it up. If the peer's commit was something unrelated (e.g. a stray rebuild request that did not heal drift), the absorber returns `false` and OCC propagates so the bootstrap retry loop re-evaluates against fresh state.
- Drop the misleading "unrecognised event type Any" warning in `AevatarOAuthClientGAgent.TransitionState`. The framework's `StateTransitionMatcher.TryExtract` already unwraps Any payloads via `ProtobufContractCompatibility.TryUnpack`.
- Test harness `InMemoryEventStore` now throws the proper `EventStoreOptimisticConcurrencyException` so the OCC absorber can be exercised end-to-end.

## Tests
- New broker tests pin "do not emit authorize URL for missing/mismatched redirect URI".
- New OAuth client GAgent tests pin the OCC absorber:
  - peer-healed: handler absorbs OCC, State reflects peer's commit, loser's DCR is logged as orphan.
  - peer-not-healed: handler rethrows OCC so bootstrap retry kicks in.

## Out of scope (tracked in issue #549 layered fix)
- Distributed lock around the DCR call so only one pod runs DCR per cluster-membership transition (eliminates orphan creation entirely instead of just absorbing the loser).
- Replace the bootstrap's fire-and-forget stream-produce dispatch with a synchronous grain RPC (Orleans turn-based serialization works for direct grain calls; stream produce is what breaks it).
- Actor rebuild admin command + endpoint for ops recovery from any wedged actor.
- Multi-silo + shared event store integration test simulating the K8s rolling-deploy two-pod overlap.

## Verification
- `dotnet test test/Aevatar.GAgents.ChannelRuntime.Tests/Aevatar.GAgents.ChannelRuntime.Tests.csproj --nologo` (826/826)
- `dotnet test test/Aevatar.Foundation.Core.Tests/Aevatar.Foundation.Core.Tests.csproj --nologo` (230/230)
- `git diff --check`

### DIFF
--- a/agents/Aevatar.GAgents.Channel.Identity/Broker/NyxIdRemoteCapabilityBroker.cs
+++ b/agents/Aevatar.GAgents.Channel.Identity/Broker/NyxIdRemoteCapabilityBroker.cs
@@ -81,13 +81,16 @@ public sealed class NyxIdRemoteCapabilityBroker : INyxIdCapabilityBroker, INyxId
         ExternalSubjectRefExtensions.EnsureValid(externalSubject);
 
         var snapshot = await _clientProvider.GetAsync(ct).ConfigureAwait(false);
+        var redirectUri = ResolveRedirectUri();
+        EnsureRedirectUriCurrent(snapshot, redirectUri);
+
         var pkce = PkceHelper.GeneratePair();
         var correlationId = Guid.NewGuid().ToString("N");
         var stateToken = await _stateTokenCodec
             .EncodeAsync(correlationId, externalSubject, pkce.CodeVerifier, ct)
             .ConfigureAwait(false);
 
-        var url = BuildAuthorizeUrl(snapshot, stateToken, pkce.CodeChallenge);
+        var url = BuildAuthorizeUrl(snapshot, redirectUri, stateToken, pkce.CodeChallenge);
         var expiresAt = _timeProvider.GetUtcNow().Add(_options.StateTokenLifetime).ToUnixTimeSeconds();
         return new BindingChallenge
         {
@@ -243,6 +246,7 @@ public sealed class NyxIdRemoteCapabilityBroker : INyxIdCapabilityBroker, INyxId
 
         var snapshot = await _clientProvider.GetAsync(ct).ConfigureAwait(false);
         var redirectUri = ResolveRedirectUri();
+        EnsureRedirectUriCurrent(snapshot, redirectUri);
 
         var form = new List<KeyValuePair<string, string>>
         {
@@ -284,9 +288,12 @@ public sealed class NyxIdRemoteCapabilityBroker : INyxIdCapabilityBroker, INyxId
         return new BrokerAuthorizationCodeResult(payload.BindingId, payload.IdToken, payload.AccessToken);
     }
 
-    private string BuildAuthorizeUrl(AevatarOAuthClientSnapshot snapshot, string stateToken, string codeChallenge)
+    private string BuildAuthorizeUrl(
+        AevatarOAuthClientSnapshot snapshot,
+        string redirectUri,
+        string stateToken,
+        string codeChallenge)
     {
-        var redirectUri = ResolveRedirectUri();
         var queryParts = new List<string>
         {
             "response_type=code",
@@ -298,6 +305,18 @@ public sealed class NyxIdRemoteCapabilityBroker : INyxIdCapabilityBroker, INyxId
             "code_challenge_method=S256",
         };
         return $"{snapshot.NyxIdAuthority.TrimEnd('/')}{AuthorizeEndpoint}?{string.Join("&", queryParts)}";
+    }
+
+    private static void EnsureRedirectUriCurrent(AevatarOAuthClientSnapshot snapshot, string resolvedRedirectUri)
+    {
+        if (!string.IsNullOrEmpty(snapshot.RedirectUri)
+            && string.Equals(snapshot.RedirectUri, resolvedRedirectUri, StringComparison.Ordinal))
+        {
+            return;
+        }
+
+        throw new AevatarOAuthClientNotProvisionedException(
+            "Aevatar OAuth client redirect_uri is not current. Bootstrap must re-run DCR before issuing NyxID authorize URLs.");
     }
 
     private static bool IsInvalidGrant(string body)
--- a/agents/Aevatar.GAgents.Channel.Identity/Provisioning/AevatarOAuthClientGAgent.cs
+++ b/agents/Aevatar.GAgents.Channel.Identity/Provisioning/AevatarOAuthClientGAgent.cs
@@ -159,38 +159,34 @@ public sealed class AevatarOAuthClientGAgent : GAgentBase<AevatarOAuthClientStat
             .RegisterPublicClientAsync(cmd.NyxidAuthority, clientName, cmd.RedirectUri, CancellationToken.None)
             .ConfigureAwait(false);
 
-        try
-        {
-            await PersistDomainEventAsync(new AevatarOAuthClientProvisionedEvent
+        // Cluster-shared Garnet event store + brief two-pod overlap during
+        // a K8s rolling deploy lets two grain activations of this well-
+        // known actor each replay v=N, each call DCR (each getting its own
+        // client_id from NyxID), and each try to commit Provisioned at
+        // expectedVersion=N. One wins, one sees OCC. See issue #549.
+        //
+        // Pass an absorber callback to PersistDomainEventAsync rather than
+        // catching OCC ourselves: the framework refreshes State from the
+        // store before invoking the callback, and there is no protected
+        // "replay state" helper a future handler could misuse outside an
+        // active commit path (PR #552 review codex/glm-5.1).
+        await PersistDomainEventAsync(
+            new AevatarOAuthClientProvisionedEvent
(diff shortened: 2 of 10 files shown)


agents/Aevatar.GAgents.Channel.Identity/Provisioning/AevatarOAuthClientGAgent.cs
```diff
@@ -178,11 +200,79 @@ await PersistDomainEventAsync(new AevatarOAuthClientProvisionedEvent
...
+                await PersistDomainEventAsync(BuildHmacKeyRotatedEvent());
+                Logger.LogInformation("Seeded HMAC key for aevatar OAuth client");
+            }
+            catch (EventStoreOptimisticConcurrencyException occ)
+            {
+                // Same race shape: a peer wrote at v+2 (e.g. their own
+                // HMAC seed) between our Provisioned commit and this
+                // seed. Refresh — if HMAC is now non-empty, the peer
+                // already seeded; absorb. The cluster has one valid key.
+                if (await TryAbsorbPeerProvisioningAsync(cmd, registration.ClientId, occ).ConfigureAwait(false))
```
eanzhao: **[codex]** severity=major, category=bug

In this HMAC-seed OCC path, `registration.ClientId` is not an orphan: this handler already committed the `Provisioned` event successfully, so that client id may be the active cluster client. If another activation wins only the HMAC seed, `TryAbsorbPeerProvisioningAsync` logs `Ops should delete the orphan client` with the active client id. Split the absorber/logging paths: DCR commit conflicts can log the loser DCR id as orphan, but HMAC seed conflicts should refresh and absorb without orphan cleanup messaging.

agents/Aevatar.GAgents.Channel.Identity/Broker/NyxIdRemoteCapabilityBroker.cs
```diff
@@ -243,6 +246,7 @@ public async Task<BrokerAuthorizationCodeResult> ExchangeAuthorizationCodeAsync(
 
         var snapshot = await _clientProvider.GetAsync(ct).ConfigureAwait(false);
         var redirectUri = ResolveRedirectUri();
+        EnsureRedirectUriCurrent(snapshot, redirectUri);
```
eanzhao, line 249: **[codex]** severity=minor, category=bug

This new `AevatarOAuthClientNotProvisionedException` can now be thrown from `ExchangeAuthorizationCodeAsync` after state-token decode succeeds, but `HandleNyxIdOAuthCallbackAsync` catches it only via the generic `catch (Exception)` and returns/logs `token_exchange_failed` 502 instead of the existing initializing/retry path. Add a specific callback catch for `AevatarOAuthClientNotProvisionedException` so redirect drift is surfaced consistently and not logged as a NyxID token-exchange failure.

src/Aevatar.Foundation.Core/GAgentBase.TState.cs
```diff
@@ -87,6 +87,24 @@ protected virtual TState TransitionState(TState current, IMessage evt)
         return current;
     }
 
+    /// <summary>
```
eanzhao, line 91: **[kimi]** severity=minor, category=arch

Adding `RefreshStateFromStoreAsync` as a protected base-class method introduces a general-purpose replay escape hatch for all actors. While the XML doc limits intent to OCC absorption, any derived actor can invoke it in normal command handlers. CLAUDE.md: "抽象一旦能被滥用，就等于设计未完成". Consider scoping more narrowly—e.g. as an internal helper or via a constrained interface—so actors cannot casually replay state inside handlers.

src/Aevatar.Foundation.Core/GAgentBase.TState.cs
```diff
@@ -87,6 +87,24 @@ protected virtual TState TransitionState(TState current, IMessage evt)
...
+    /// Re-reads persisted events from the event store and replaces
+    /// in-memory <see cref="State"/> with the rebuilt result. Intended for
+    /// use from event-handler bodies that have caught
+    /// <c>EventStoreOptimisticConcurrencyException</c> and need to absorb
+    /// peer-committed facts before deciding whether to retry, no-op, or
+    /// rethrow. Does not invoke <see cref="OnStateChangedAsync"/> — the
+    /// projection-publish pipeline is driven by the peer's own
+    /// <c>PersistDomainEventsAsync</c> commit, not by this re-load.
+    /// </summary>
+    protected async Task RefreshStateFromStoreAsync(CancellationToken ct = default)
```
eanzhao: **[kimi]** severity=minor, category=design

`RefreshStateFromStoreAsync` should call `eventSourcing.DiscardPendingEvents()` after (or before) `ReplayAsync`. When `PersistDomainEventAsync` catches `EventStoreOptimisticConcurrencyException`, `EventSourcingBehavior.ConfirmEventsAsync` only removes the committed prefix from `_pending`; any events raised mid-flight (during the async `AppendAsync`) survive as a suffix. If the caller absorbs the OCC and returns, those stale pending events will be committed on the next `ConfirmEventsAsync`, silently corrupting the event stream.

test/Aevatar.GAgents.ChannelRuntime.Tests/Identity/NyxIdRemoteCapabilityBrokerTests.cs
```diff
@@ -0,0 +1,126 @@
+using Aevatar.GAgents.Channel.Abstractions;
```
eanzhao, line 1: **[glm-5.1]** severity=minor, category=test

Tests cover `StartExternalBindingAsync` redirect-URI guards, but `ExchangeAuthorizationCodeAsync` received the same `EnsureRedirectUriCurrent` guard (broker line ~291) without a corresponding test. A mismatched redirect URI during token exchange is arguably the higher-impact failure mode — the authorization code has already been issued and will be exchanged against the wrong client. Consider adding at least one test for the code-exchange path.

test/Aevatar.GAgents.ChannelRuntime.Tests/Identity/NyxIdRemoteCapabilityBrokerTests.cs
```diff
@@ -0,0 +1,126 @@
...
+{
+    private static readonly byte[] HmacKey =
+        Convert.FromHexString("0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef");
+
+    private readonly string? _savedOverride;
+
+    public NyxIdRemoteCapabilityBrokerTests()
+    {
+        _savedOverride = Environment.GetEnvironmentVariable(NyxIdRedirectUriResolver.OverrideEnvVar);
+        Environment.SetEnvironmentVariable(NyxIdRedirectUriResolver.OverrideEnvVar, null);
```
eanzhao, line 25: **[codex]** severity=minor, category=test

This new test class mutates the process-wide `AEVATAR_OAUTH_REDIRECT_BASE_URL` env var while `NyxIdRedirectUriResolverTests` also mutates it, and the test project does not disable xUnit class-level parallelization. These classes can race and restore each other's saved values, causing flaky redirect assertions. Put both classes in the same non-parallel collection or avoid process env mutation in these broker tests.

PR #555, opened 2026-05-01
Activate binding projection scope + materialize legacy bindings on commit

eanzhao:
## Problem
After PR #552 merged, production still loops users back to /init forever. Pod logs from `aevatar-console-backend` 2026-05-01:

```
ExternalIdentityBindingGAgent[0]: CommitBinding discarded: already bound for
  lark:...:ou_937adb03f3538c5e041bb3034c4e348e
  (existing=bnd_41ebae..., incoming=bnd_7fbc76ed...)
warn: Aevatar.Channel.Identity.OAuthCallback[0]: Projection readiness timed out
  for actor=external-identity-binding:..., expected binding=bnd_7fbc76ed...
```

The binding actor's `State` already holds `bnd_41ebae` from a previous deploy, but the binding readmodel is empty. The OAuth callback's `WaitForBindingStateAsync` polls for the new `bnd_7fbc76ed` (or any matching binding) and times out, returning `binding_pending_propagation`. The next inbound message's gate also queries the empty readmodel and re-sends the `/init` card. Loop.

## Root cause
Every other GAgent (`AevatarOAuthClient`, `ChannelBotRegistration`, `Device`, `UserAgentCatalog`) has an explicit `EnsureProjectionForActorAsync` wiring; `ExternalIdentityBindingGAgent` never did. Without it the projector never subscribes to the binding actor's committed event stream, so even a perfectly-committed binding never lands in the readmodel.

This is the same shape as PR #539 fixed for `AevatarOAuthClientGAgent` — the binding scope was simply missed when that PR went out.

## Solution
1. **`ExternalIdentityBindingProjectionPort`** wrapping `MaterializationProjectionPortBase`. Registered as singleton in `IdentityServiceCollectionExtensions`.
2. **`HandleNyxIdOAuthCallbackAsync` calls `EnsureProjectionForActorAsync(bindingActorId)`** before the `queryPort.ResolveAsync` check and the `actor.HandleEventAsync` dispatch — so the projector is subscribed before any read or write touches that actor.
3. **`ExternalIdentityBindingProjectionRebuildRequestedEvent`** (Apply identity). `HandleCommitBinding`'s "already bound" discard branch now persists this event instead of returning silently. The projector sees the state-root republication and materializes the existing binding — so legacy actors heal on the next `/init` even though no new `ExternalIdentityBoundEvent` fires.
4. **`TransitionState` cleanup**: drop the misleading `unrecognised event type Any` warning (`StateTransitionMatcher.TryExtract` already unwraps Any), add `ExternalIdentityBindingProjectionRebuildRequestedEvent` as identity in the apply switch.

## Tests
- `HandleCommitBinding_IsIdempotentUnderConcurrentInit` extended to assert the version advances by one when the second commit is "already bound" — pins that the discard branch must emit a rebuild event so the projector re-publishes the existing binding's state root.
- Existing 8 binding-actor behaviour tests still pass without modification.

## Verification
- `dotnet test test/Aevatar.GAgents.ChannelRuntime.Tests/Aevatar.GAgents.ChannelRuntime.Tests.csproj --nologo` (828/828)
- `dotnet test test/Aevatar.Foundation.Core.Tests/Aevatar.Foundation.Core.Tests.csproj --nologo` (230/230)

## Refs
Issue #549 (binding-projection-scope-missing follow-up). PR #552 fixed the OAuth-client side of this same pattern.

### DIFF
--- a/agents/Aevatar.GAgents.Channel.Identity/ExternalIdentityBindingGAgent.cs
+++ b/agents/Aevatar.GAgents.Channel.Identity/ExternalIdentityBindingGAgent.cs
@@ -19,27 +19,22 @@ namespace Aevatar.GAgents.Channel.Identity;
 public sealed partial class ExternalIdentityBindingGAgent : GAgentBase<ExternalIdentityBindingState>
 {
     /// <inheritdoc />
-    protected override ExternalIdentityBindingState TransitionState(ExternalIdentityBindingState current, IMessage evt)
-    {
-        // Log unrecognised event types so a proto schema drift (or stale
-        // event store entry from a removed event type) surfaces in operator
-        // dashboards rather than silently being dropped via OrCurrent().
-        // Production deployments should treat these as a regression signal.
-        if (evt is not null
-            && evt is not ExternalIdentityBoundEvent
-            && evt is not ExternalIdentityBindingRevokedEvent)
-        {
-            Logger.LogWarning(
-                "ExternalIdentityBindingGAgent received unrecognised event type {EventType}; state unchanged",
-                evt.GetType().FullName);
-        }
-
-        return StateTransitionMatcher
+    /// <remarks>
+    /// <see cref="StateTransitionMatcher"/> handles <c>Any</c>-wrapped payloads
+    /// transparently via <c>ProtobufContractCompatibility.TryUnpack</c>, so the
+    /// event-store's wrapped form ("type.googleapis.com/...") is matched the
+    /// same as a directly-typed instance. No "unrecognised event type"
+    /// pre-check fires here — the earlier guard incorrectly classified every
+    /// Any-wrapped event as unknown and produced noisy warnings on every
+    /// activation replay.
+    /// </remarks>
+    protected override ExternalIdentityBindingState TransitionState(ExternalIdentityBindingState current, IMessage evt) =>
+        StateTransitionMatcher
             .Match(current, evt)
             .On<ExternalIdentityBoundEvent>(ApplyBound)
             .On<ExternalIdentityBindingRevokedEvent>(ApplyRevoked)
+            .On<ExternalIdentityBindingProjectionRebuildRequestedEvent>(static (state, _) => state)
             .OrCurrent();
-    }
 
     // ─── Commands ───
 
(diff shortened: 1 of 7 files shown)


agents/Aevatar.GAgents.Channel.Identity/ExternalIdentityBindingGAgent.cs
```diff
@@ -82,8 +77,23 @@ public async Task HandleCommitBinding(CommitBindingCommand cmd)
...
+            // projector materializes the existing binding into the readmodel.
+            // Without this, a legacy binding actor whose projection scope
+            // was never activated (issue #549 follow-up: the binding scope
+            // missed an EnsureProjectionForActorAsync wiring while every
+            // other GAgent had one) leaves the readmodel empty, the OAuth
+            // callback's readiness wait times out, and the next inbound
+            // message's binding gate keeps re-sending the user back to /init.
+            // Apply is identity, so the binding facts are not mutated by
+            // this event.
+            await PersistDomainEventAsync(new ExternalIdentityBindingProjectionRebuildRequestedEvent
```
eanzhao, line 90: 这个 rebuild event 可以把 existing binding materialize 出来，但 callback 端仍然调用 `WaitForBindingStateAsync(subject, exchange.BindingId, ...)` 等 incoming binding。legacy already-bound 时 actor 保留的是 `State.BindingId`（旧值），不会切到 `cmd.BindingId`（incoming），所以 readiness 会等不到并返回 `binding_pending_propagation`。更重要的是，NyxID `/oauth/token` 已经为 incoming code 创建了新的 broker binding；因为 callback 没识别到 actor 的 discard/rebuild 分支，这个 incoming binding 也不会走 `TryRevokeOrphanBindingAsync`。

建议在 callback 的 timeout 后重新 `ResolveAsync(subject)`：如果读到了一个 active binding 且不同于 `exchange.BindingId`，说明这里走了 already-bound discard/rebuild，应 revoke incoming binding 并返回 `already_bound`；或者让 readiness 支持“任意 active binding”并在不同 id 时显式 revoke incoming binding。需要加 callback-level 测试覆盖这个 legacy already-bound 场景。

eanzhao (commented):
我看到一个 legacy already-bound 分支还没有收口：PR 的 no-op rebuild 会把旧 binding materialize 出来，但 callback 仍然等新 exchange 出来的 binding_id，因此会超时并留下 NyxID 侧新建的 orphan binding。详见 inline comment。

agents/Aevatar.GAgents.Channel.Identity/ExternalIdentityBindingGAgent.cs
eanzhao: Rechecked latest head `9da663e`. This addresses the issue I raised: after readiness timeout the callback re-resolves the binding readmodel, treats `existing != incoming` as the legacy already-bound/rebuild path, revokes the incoming NyxID binding, and returns `already_bound`. The new callback tests cover legacy already-bound, still-empty pending propagation, and happy path. Local verification passed: `dotnet test test/Aevatar.GAgents.ChannelRuntime.Tests/Aevatar.GAgents.ChannelRuntime.Tests.csproj --nologo` (831/831), `dotnet test test/Aevatar.Foundation.Core.Tests/Aevatar.Foundation.Core.Tests.csproj --nologo` (230/230), `bash tools/ci/test_stability_guards.sh`, `bash tools/ci/query_projection_priming_guard.sh`, and `git diff --check`.

Issue #556, opened 2026-05-01
Cross-channel LLM selection + setup guide (no LLM in the loop)


eanzhao:
## Problem

Today the only Lark slash command for LLM preferences is `/model` (`agents/Aevatar.GAgents.NyxidChat/Slash/ModelChannelSlashCommandHandler.cs`), and it has three gaps observed in production 2026-05-01:

1. **No discovery.** `/model list` shows only the user's currently-set model name and the bot owner's default. It does NOT enumerate which LLM-capable NyxID services this user actually has access to. A user who just `/init`-bound but has no LLM service configured at NyxID has no idea what to type next.
2. **No route switch.** `/model use <name>` only writes `UserConfig.DefaultModel` (`ModelChannelSlashCommandHandler.HandleUseAsync` line 152-159 deliberately preserves `current.PreferredLlmRoute`). Switching between e.g. `chrono-llm`, an OpenAI service the user added, or an Anthropic key requires the Studio web console — not reachable from a chat client.
3. **No empty-state UX.** First-time user with 0 LLM services configured has no path inside the chat to add one. The bot can't ask the LLM (the LLM isn't configured yet).

Parallel symptom: `/models` (typo) falls through to `NyxRelayAgentBuilderFlow.cs:545`, which prints its own "Supported commands" list that does NOT include `/init`/`/whoami`/`/unbind`/`/model` because the agent-builder flow is a separate slash router. Discoverability of channel-identity slash commands is broken across the board.

## Goals + constraints

| Constraint | Meaning |
|---|---|
| **Zero LLM in the discovery + setup path** | First-run user has 0 LLM configured. Selection flow must be 100% deterministic code paths. |
| **Channel-agnostic application layer** | Lark / Telegram / Discord + Aevatar Studio web + CLI all share one application contract. Renderer is chat-only — Studio/CLI consume the DTO directly. |
| **NyxID services may *be* LLM providers** | `chrono-llm` is a NyxID-internal hosted service; user-added OpenAI/Anthropic keys are also NyxID services. Aevatar must NOT classify LLM-capability itself — it asks NyxID. |
| **Selection persists to per-binding `UserConfig`** | Reuses the existing `PreferredLlmRoute` + `DefaultModel` override chain in `ConversationReplyGenerator.BuildEffectiveMetadataAsync` (`agents/Aevatar.GAgents.NyxidChat/ConversationReplyGenerator.cs:139-201`). |
[...]

PR #557, opened 2026-05-01
Complete cross-channel LLM selection flow

eanzhao:
## Summary

Refs #556.

This PR now implements the full cross-channel LLM selection flow from the issue:

- adds channel-neutral LLM option / selection contracts (`UserLlmOptionsView`, `UserLlmOption`, setup hints, presets, preset activation records) and backs them with a real NyxID `/api/v1/llm/services` catalog client
- routes option discovery and preset provisioning through application services, with `llm:status` / `llm:proxy` capability issuance via `INyxIdCapabilityBroker`, and persists selection into UserConfig (`PreferredLlmRoute`, `DefaultModel`)
- removes slug/blacklist LLM classification from the affected paths and migrates Studio, NyxID tool status, Streaming Proxy participant discovery, and demo LLM reporting away from `/api/v1/llm/status`
- implements `/model` plus `/models` and `/llm` aliases, unknown-command help descriptors, list/use/preset/reset flows, and channel card action handling for service/preset selection
- adds Telegram native inline button rendering for card actions so selection cards are actionable outside the chat renderer
- adds Studio endpoints for LLM options and preferences, keeps legacy `/models` compatibility backed by the new services response, and wires the Studio frontend to save runtime + LLM preferences together
- adds CLI `aevatar model list/use/preset/reset` commands using the same DTO surface

## Validation

- `dotnet test test/Aevatar.GAgents.ChannelRuntime.Tests/Aevatar.GAgents.ChannelRuntime.Tests.csproj --nologo --no-restore`
- `dotnet test test/Aevatar.GAgents.Platform.Telegram.Tests/Aevatar.GAgents.Platform.Telegram.Tests.csproj --nologo --no-restore`
- `dotnet test test/Aevatar.Tools.Cli.Tests/Aevatar.Tools.Cli.Tests.csproj --nologo --no-restore`
- `dotnet test test/Aevatar.AI.Tests/Aevatar.AI.Tests.csproj --nologo --no-restore`
- `bash tools/ci/playground_asset_drift_guard.sh`
- `dotnet build aevatar.slnx --nologo --no-restore`


agents/platforms/Aevatar.GAgents.Platform.Telegram/TelegramMessageComposer.cs
```diff
@@ -44,6 +39,33 @@ public TelegramOutboundMessage Compose(MessageContent intent, ComposeContext con
...
+        var buttonActions = GetButtonActions(intent);
+        if (buttonActions.Length > 0 && capabilities.SupportsActionButtons)
+        {
+            var inlineKeyboard = buttonActions
+                .Select(action => new[]
+                {
+                    new
+                    {
+                        text = action.Label.Trim(),
+                        callback_data = BuildCallbackData(action),
```
eanzhao: **[Consensus: 5 models]** severity=`blocker`, category=`bug`

Telegram's `callback_data` field has a hard 64-byte limit. `BuildCallbackData` JSON-serializes the full action payload (`action_id`, `submitted_value`, and all `Arguments`), which will easily exceed 64 bytes for the LLM selection actions introduced by this PR. Telegram will reject the message, so the inline buttons won't render. **Fix:** encode a compact payload under 64 bytes (e.g. an opaque short token) or store the action arguments server-side behind a short nonce.

<details><summary>Per-model verbatim</summary>

- **codex**: Telegram `callback_data` is limited to 64 bytes, but `BuildCallbackData` serializes a verbose JSON object containing `action_id`, `submitted_value`, and all arguments. The LLM selection actions generated by this PR will commonly exceed that limit, causing Telegram to reject the message instead of rendering the buttons. Encode a compact payload under 64 bytes or store the action arguments server-side behind a short nonce.

- **gemini**: Telegram `callback_data` has a strict 64-byte limit. `JsonSerializer.Serialize(payload)` containing `action_id`, `submitted_value`, and additional arguments (like `llm_action` and `service_id`) can easily exceed this limit, causing message delivery to fail. Suggest a more compact encoding for selection action payloads.

- **kimi**: `BuildCallbackData` JSON-serializes the full action payload (`action_id`, `submitted_value`, plus all `Arguments`). The resulting string easily exceeds Telegram Bot API's hard **64-byte** limit for `callback_data` on inline keyboard buttons. For example a single service-selection button produces ~140 bytes. This will cause Telegram API errors when users tap buttons. Either compress/map to short tokens server-side, store state in a session cache and send a short key, or shard the payload.

- **mimo-v2.5-pro**: `BuildCallbackData` serialises the full `ActionElement` (action_id + submitted_value + all Arguments) into Telegram's `callback_data`. Telegram enforces a **64-byte** limit on `callback_data`; payloads exceeding this are silently dropped by the Bot API.

For a typical LLM selection button the serialised JSON is ~80-100 bytes (`{"action_id":"llm_select_service","submitted_value":"svc-openai","value":{"llm_action":"select_service","service_id":"svc-openai"}}`), which will silently fail when the user taps the button.

**Fix:** Either hash/compress the payload (e.g. map each action to a short lookup key stored server-side, or use a compact binary encoding), or truncate/validate before sending. At minimum, throw or fall back to text-only if `callback_data.Length > 64`.

- **v4-pro**: `BuildCallbackData` serializes the full argument dictionary into JSON for Telegram's `callback_data` field, which has a strict **64-byte** limit. LLM selection cards with multi-key arguments (e.g. `llm_action` + `service_id` + potentially long values) will exceed this limit and be truncated by Telegram. Consider compressing the payload or using a shorter encoding scheme (e.g. concatenated `action|service_id` delimited string instead of full JSON).

</details>

agents/Aevatar.GAgents.NyxidChat/ServiceCollectionExtensions.cs
```diff
@@ -52,6 +54,10 @@ public static IServiceCollection AddNyxIdChat(this IServiceCollection services,
         // Registered here (not in Channel.Identity) because the handler depends
         // on Studio.Application UserConfig ports; Channel.Identity intentionally
         // does not pull Studio dependencies.
+        services.TryAddSingleton<INyxIdLlmServiceCatalogClient, NyxIdLlmServiceCatalogClient>();
```
eanzhao, line 57: **[Consensus: 2 models]** severity=`blocker`, category=`di`

`DefaultUserLlmOptionsService` and `DefaultUserLlmSelectionService` are registered as **Singleton** but their constructors accept optional `IUserConfigQueryPort?` / `IUserConfigDispatchPort?` (typically Scoped). This is a **captive dependency** — Singleton holds a reference to a Scoped service, breaking the Scoped lifetime contract. **Fix:** register both services as Scoped, or change the dependencies to be resolved per-call via `IServiceScopeFactory`.

<details><summary>Per-model verbatim</summary>

- **glm-5.1**: `DefaultUserLlmOptionsService` and `DefaultUserLlmSelectionService` are registered as **Singleton** but their constructors accept optional `IUserConfigQueryPort?` and `IUserConfigCommandService?`. If those interfaces are registered as Scoped (which is typical for per-request read-model ports), this creates a **captive dependency** — the Singleton holds a reference to a Scoped service that will be disposed while the Singleton is still alive.

Verify the lifetimes of `IUserConfigQueryPort`, `IUserConfigCommandService`, and `INyxIdCapabilityBroker` in the host composition root. If any are Scoped, either: (1) change these services to Scoped, (2) inject `IServiceScopeFactory` instead and create a scope per call, or (3) register the ports as Singleton if they are truly stateless.

- **v4-pro**: `DefaultUserLlmOptionsService` and `DefaultUserLlmSelectionService` are registered as **Singleton** but inject `IUserConfigQueryPort?` and `IUserConfigCommandService?`. If these ports are registered as **Scoped** (plausible in the Studio projection pipeline where `IUserConfigQueryPort.GetAsync(CancellationToken)` reads the ambient HTTP scope), the Singleton will hold a captive reference to a Scoped service. Either (a) make these services Scoped/Transient, (b) confirm the ports are also Singleton, or (c) inject `IServiceScopeFactory` and create a scope per call.

</details>

src/Aevatar.AI.ToolProviders.NyxId/NyxIdApiClient.cs
```diff
@@ -331,8 +331,22 @@ public Task<string> RotateNodeTokenAsync(string token, string id, CancellationTo
 
     // ─── LLM ───
 
-    public Task<string> GetLlmStatusAsync(string token, CancellationToken ct) =>
-        GetAsync(token, "/api/v1/llm/status", ct);
+    public Task<string> GetLlmServicesAsync(string token, CancellationToken ct) =>
+        GetAsync(token, "/api/v1/llm/services", ct);
```
eanzhao: **[codex]** severity=`blocker`, category=`bug`

`/api/v1/llm/services` is not implemented in the local NyxID source required by AGENTS.md; `/Users/chronoai/Code/NyxID/backend/src/routes.rs` only registers `/api/v1/llm/status` under the LLM router. As written, `/model`, Studio LLM options, and Streaming Proxy discovery will all get 404s unless a paired NyxID change lands first. Either update the NyxID dependency/contract in the same stack or keep a `/status` compatibility fallback until `/services` is actually available.

tools/Aevatar.Tools.Cli/Frontend/src/App.tsx
```diff
@@ -207,9 +207,11 @@ type RuntimeMode = 'remote' | 'local';
 type UserLlmRoute = string;
```
eanzhao, line 207: **[Consensus: 4 models]** severity=`major`, category=`bug`

`Promise.all([api.userConfig.save(runtime), api.userConfig.saveLlmPreference(llm)])` fires two writes concurrently, but the backend `UserConfigController` does a read-modify-write for each one (fetch state via `_queryPort.GetAsync`, dispatch a full `UserConfigUpdatedEvent`). One write can overwrite the other with stale data, AND if one fails while the other succeeds the user is left in an inconsistent state. **Fix:** add a single combined save endpoint, or sequence the calls (await save, then awaitsaveLlmPreference) with rollback on second-write failure.

<details><summary>Per-model verbatim</summary>

- **codex**: Saving runtime and LLM preferences concurrently can lose one side of the update. Both `PUT /user-config` and `PUT /user-config/llm/preference` read the current config, merge their own fields, and save a full `UserConfig`; whichever request writes last can overwrite the other request with stale values. Use a single backend operation for the combined save, or sequence the calls so the final write carries both the runtime fields and the returned LLM route/model values.

- **gemini**: Race condition: `api.userConfig.save` and `api.userConfig.saveLlmPreference` are called in parallel via `Promise.all`. Since the backend `UserConfigController` methods perform a read-modify-write operation (fetching state via `_queryPort.GetAsync` before dispatching a full `UserConfigUpdatedEvent`), one request will likely overwrite the other's changes with stale data. These should be executed sequentially or the backend should support partial updates for LLM preferences.

- **glm-5.1**: `Promise.all([api.userConfig.save(runtime), api.userConfig.saveLlmPreference(llm)])` runs two independent saves in parallel. If one succeeds and the other fails (e.g., network timeout), the user's config is left in a partially-saved state with no rollback. Consider either: (1) combining into a single endpoint call, or (2) saving sequentially with rollback on the second failure.

- **v4-pro**: `Promise.all` fires `api.userConfig.save()` and `api.userConfig.saveLlmPreference()` in parallel. If one succeeds and the other fails, the user's runtime config and LLM preference drift into an inconsistent state (e.g. routing updated but model not saved, or vice versa). Should either be a single atomic server endpoint, or executed sequentially with rollback on failure.

</details>

src/Aevatar.Studio.Hosting/Controllers/UserConfigController.cs
```diff
@@ -106,238 +100,509 @@ public sealed record SaveUserConfigRequest(
...
+        catch (InvalidOperationException exception)
+        {
+            return BadRequest(new { message = exception.Message });
+        }
+        catch (Exception exception) when (exception is not OperationCanceledException)
+        {
+            _logger.LogWarning(exception, "Failed to fetch user LLM options");
+            return StatusCode(502, new { message = "LLM options are temporarily unavailable." });
+        }
+    }
```
eanzhao, line 142: **[Consensus: 4 models]** severity=`major`, category=`arch`

`UserLlmOptionsView`, `UserLlmOption`, `UserLlmSetupHint`, `UserLlmPreset`, `UserLlmPresetActivation`, `UseExistingService`, `ProvisionThenUse` are defined twice — once in `agents/Aevatar.GAgents.NyxidChat/LlmSelection/UserLlmSelectionContracts.cs` (no JSON attrs) and once in `src/Aevatar.Studio.Hosting/Controllers/UserConfigController.cs` (with `[JsonPropertyName]`). Two parallel record hierarchies with identical shapes is a maintenance trap — any field added in one will silently drift in the other. **Fix:** move the contract types into a shared abstractions project (e.g. `Aevatar.Studio.Application.Abstractions`) referenced by both the Host controller and the agent layer, or have the agent layer reference `Aevatar.Studio.Hosting`'s public DTOs.

<details><summary>Per-model verbatim</summary>

- **glm-5.1**: `UserLlmOptionsView`, `UserLlmOption`, `UserLlmSetupHint`, `UserLlmPreset`, `UserLlmPresetActivation`, `UseExistingService`, `ProvisionThenUse` are all defined here as **public records** with JSON serialization attributes, but identically-named records exist in `UserLlmSelectionContracts.cs` for the channel domain. Two parallel type hierarchies with the same names in different namespaces is a maintenance trap. Unify on a single set of contracts (or at minimum, map explicitly between them with a single conversion point).

- **kimi**: `UserLlmOptionsView`, `UserLlmOption`, `UserLlmSetupHint`, `UserLlmPreset`, `UserLlmPresetActivation` are duplicated here and in `agents/Aevatar.GAgents.NyxidChat/LlmSelection/UserLlmSelectionContracts.cs`. Shared API contracts should live in a common abstractions project (e.g. `Aevatar.Studio.Application.Abstractions`) that both the agent layer and the Host layer reference, instead of maintaining two near-idententical shapes.

- **mimo-v2.5-pro**: The same domain types (`UserLlmOption`, `UserLlmSetupHint`, `UserLlmPreset`, `UserLlmPresetActivation` + subtypes) are defined in both `Aevatar.GAgents.NyxidChat.LlmSelection` (here, lines 1-123) and `Aevatar.Studio.Hosting.Controllers.UserConfigController` (inner records, lines ~680-710 of `UserConfigController.cs`). The two definitions are structurally identical but live in different namespaces and assemblies.

This means:
- The controller's `ParseLlmService` creates `Studio.Hosting…UserLlmOption` while `LoadLlmOptionsAsync` returns the NyxidChat `UserLlmOptionsView` containing `NyxidChat…UserLlmOption` — an implicit cast that will break if the types ever diverge.
- Any future change to the contract must be made in two places.

**Fix:** Delete the controller-local definitions and import the canonical types from `Aevatar.GAgents.NyxidChat.LlmSelection` (or a shared abstractions package). The controller's `ParseLlmService` should produce the NyxidChat `UserLlmOption` directly.

- **v4-pro**: `UserLlmOptionsView`, `UserLlmOption`, `UserLlmSetupHint`, `UserLlmPreset`, `UserLlmPresetActivation`, `UseExistingService`, `ProvisionThenUse` are defined twice: once in `NyxidChat/LlmSelection/UserLlmSelectionContracts.cs` (no JSON attrs) and again in this controller file (with `[JsonPropertyName]` attrs). Having two independent record hierarchies with identical shapes but different serialization semantics is a maintenance risk. Pick a single canonical definition and add JSON attrs to it.

</details>

agents/Aevatar.GAgents.NyxidChat/ChannelConversationTurnRunner.cs
```diff
@@ -97,6 +98,9 @@ public async Task<ConversationTurnResult> RunInboundAsync(
         if (bindingGateResult is not null)
             return bindingGateResult;
 
+        if (await TryHandleLlmSelectionCardActionAsync(activity, inbound, registration, runtimeContext, senderBindingId, ct).ConfigureAwait(false) is { } llmSelectionResult)
```
eanzhao, line 128: **[Consensus: 3 models]** severity=`major`, category=`di`

`ChannelConversationTurnRunner` resolves `IExternalIdentityBindingQueryPort`, `IUserLlmSelectionService`, and `IUserLlmOptionsService` at runtime via `_services.GetService<T>()`. This is the **service-locator anti-pattern** explicitly forbidden by `CLAUDE.md`'s DI guidance — it hides dependencies and prevents the DI container from validating them at startup. **Fix:** inject these services through the constructor (or a factory abstraction) and let DI fail fast if a registration is missing.

<details><summary>Per-model verbatim</summary>

- **glm-5.1**: Service-locator anti-pattern: `_services.GetService<IExternalIdentityBindingQueryPort>()` (and again on lines 275–277 for `IUserLlmSelectionService`, `IUserLlmOptionsService`, `IUserLlmOptionsRenderer<MessageContent>`). The turn runner already receives `IServiceProvider` but these dependencies should be injected directly into the constructor so the container validates the full graph at startup rather than silently returning null at runtime when a module is missing. If conditional availability is intentional, register null-object implementations or use a factory — don't reach into the container at call sites.

- **mimo-v2.5-pro**: The method resolves three services from `IServiceProvider` at runtime (`GetService<IExternalIdentityBindingQueryPort>()`, `GetService<IUserLlmOptionsService>()`, etc.) instead of receiving them via constructor injection. This is a service-locator anti-pattern that makes dependencies invisible, harder to test, and risks silent null-reference failures if registrations are missing.

`ChannelConversationTurnRunner` already takes `IServiceProvider` in its constructor; consider injecting the specific services directly (or at least documenting why service-locator is required here, e.g. optional/circular dependency).

- **v4-pro**: `_services.GetService<IExternalIdentityBindingQueryPort>()` and subsequent repeated `_services.GetService<T>()` calls in `TryHandleLlmSelectionCardActionAsync` use the service-locator pattern. While this is an existing convention in this class, the method now makes 4 separate `GetService` calls for the same handler invocation. Consider caching the resolved services in locals at the method top to reduce repeated locator lookups.

</details>

src/Aevatar.Studio.Hosting/Controllers/UserConfigController.cs
```diff
@@ -106,238 +100,509 @@ public sealed record SaveUserConfigRequest(
...
-        string bearerToken,
-        CancellationToken cancellationToken)
+    public sealed record SaveUserLlmPreferenceRequest(
+        [property: JsonPropertyName("serviceId")] string? ServiceId = null,
+        [property: JsonPropertyName("routeValue")] string? RouteValue = null,
+        [property: JsonPropertyName("model")] string? Model = null,
+        [property: JsonPropertyName("presetId")] string? PresetId = null,
+        [property: JsonPropertyName("reset")] bool? Reset = null);
+
+    private async Task<UserLlmOptionsView> LoadLlmOptionsAsync(CancellationToken cancellationToken)
```
eanzhao: **[Consensus: 2 models]** severity=`major`, category=`arch`

`UserConfigController` re-implements the same NyxID JSON-parsing stack (`ParseLlmServicesResult`, `ParseLlmService`, `ParseSetupHint`, `ParsePreset`, `ParsePresetActivation`, `ReadRequiredString`, `ReadOptionalString`, `ReadOptionalBool`, `ReadStringArray`) that `NyxIdLlmServiceCatalogClient` already implements. Duplicate parsing logic means any upstream NyxID response format change must be updated in two places. **Fix:** expose the parsing helpers on `INyxIdLlmServiceCatalogClient` (or a shared `Aevatar.AI.NyxId.Parsing` utility) and have the controller call into them.

<details><summary>Per-model verbatim</summary>

- **glm-5.1**: The controller duplicates the entire NyxID JSON parsing stack (`ParseLlmServicesResult`, `ParseLlmService`, `ParseSetupHint`, `ParsePreset`, `ParsePresetActivation`, `ReadRequiredString`, `ReadOptionalString`, `ReadOptionalBool`, `ReadStringArray`) that `NyxIdLlmServiceCatalogClient` already implements. This violates DRY and creates two independent parsing surfaces that can drift apart when the NyxID API evolves.

Refactor: inject `INyxIdLlmServiceCatalogClient` + `IUserLlmOptionsService` into the controller and delegate to them, converting the public-facing DTOs at the boundary.

- **v4-pro**: `ParseServicesResult`, `ParseLlmService`, `ParseSetupHint`, `ParsePreset`, `ParseActivation` — these private JSON-parsing helpers are nearly identical to the same-named methods in `UserConfigController` (line ~3038–3168). Duplicate parsing logic means any upstream NyxID response format change must be updated in two places. Extract shared parsing into `INyxIdLlmServiceCatalogClient` (or a shared utility) and have both consumers call it.

</details>

agents/Aevatar.GAgents.NyxidChat/LlmSelection/TextUserLlmOptionsRenderer.cs
```diff
@@ -0,0 +1,199 @@
...
+        }
+
+        lines.Add("");
+        lines.Add("用法:");
+        lines.Add("- `/model use <编号|service-name|model-name>` 切换 service 或只覆盖 model");
+        lines.Add("- `/model preset <preset-id>` 使用 setup preset");
+        lines.Add("- `/model reset` 清空你的 service/model 偏好,回退到 bot 默认");
+
+        var reply = new MessageContent { Text = string.Join('\n', lines) };
+        reply.Cards.Add(new CardBlock
```
eanzhao, line 53: **[glm-5.1]** severity=`major`, category=`bug`

`ReferenceEquals(option, view.Current)` is wrong for `record` types. In `UserConfigController.LoadLlmOptionsAsync` the current option is rebuilt via `current with { DefaultModel = ... }`, which creates a **new** record instance — so `ReferenceEquals` will return `false` even when the option represents the same service. The ✓ marker will never appear for Studio-initiated flows.

Fix: replace with a value-based comparison, e.g.:
```csharp
var marker = view.Current is not null &&
    string.Equals(option.ServiceId, view.Current.ServiceId, StringComparison.OrdinalIgnoreCase)
    ? " ✓" : string.Empty;
```

src/Aevatar.Studio.Hosting/Controllers/UserConfigController.cs
eanzhao: **[kimi]** severity=`major`, category=`arch`

`LoadLlmOptionsAsync`, `ActivatePresetAsync`, `ParseLlmServicesResult`, `IsSelectable`, and ~400 lines of JSON parsing / business orchestration live directly in a Host-layer Controller. This violates CLAUDE.md **"API 仅做宿主与组合，不承载核心业务编排"**. Extract this logic into a dedicated Application service (e.g. `IUserLlmOptionsAppService`) and keep the controller as a thin translation layer.

src/Aevatar.Studio.Hosting/Controllers/UserConfigController.cs
```diff
@@ -106,238 +100,509 @@ public sealed record SaveUserConfigRequest(
...
-                    var proxyUrl = provider.ProxyUrl!.Trim().TrimEnd('/');
-                    var baseUrl = proxyUrl.StartsWith("http", StringComparison.OrdinalIgnoreCase)
-                        ? proxyUrl
-                        : $"{authorityBase}{proxyUrl}";
+        if (string.IsNullOrWhiteSpace(provisionEndpointId))
+            throw new InvalidOperationException("LLM preset provisioning endpoint is missing.");
+
+        var body = await SendNyxIdAsync(
+            HttpMethod.Post,
+            $"/api/v1/llm/services/{Uri.EscapeDataString(provisionEndpointId.Trim())}",
```
eanzhao: **[kimi]** severity=`major`, category=`security`

`ActivateProvisioningPresetAsync` builds a URL path with `Uri.EscapeDataString(provisionEndpointId)` but performs **no** `..` or `://` validation. `NyxIdApiClient.ProvisionLlmServiceAsync` (line 342) has exactly this guard. A malicious or compromised preset response could traverse paths (e.g. `..` resolves to `/api/v1/llm/` on most HTTP servers). Either reuse `NyxIdApiClient` or add the same `Trim('/') + Contains("..") + Contains("://")` validation before constructing the path.

src/Aevatar.Studio.Hosting/Controllers/UserConfigController.cs
```diff
@@ -106,238 +100,509 @@ public sealed record SaveUserConfigRequest(
...
-            ],
-            ["google-ai"] = [
-                "gemini-2.5-pro-preview-06-05",
-                "gemini-2.5-flash-preview-05-20",
-                "gemini-2.0-flash",
-            ],
-            ["cohere"] = [
-                "command-r-plus", "command-r", "command-a-03-2025",
-            ],
+            UseExistingService existing => ActivateExistingPreset(options.Services, existing),
```
eanzhao: **[gemini]** severity=`major`, category=`bug`

`ToLegacyStatusResponse` uses `ToDictionary` with `option.ServiceSlug` as the key. If a user has multiple services with the same slug (e.g., multiple OpenAI keys), this will throw `ArgumentException`. Suggest grouping by slug and merging model lists to ensure uniqueness in the legacy response dictionary.

agents/Aevatar.GAgents.NyxidChat/ChannelConversationTurnRunner.cs
```diff
@@ -489,6 +493,191 @@ private static bool IsPrivateChat(InboundMessage inbound)
         return null;
```
eanzhao, line 517: **[mimo-v2.5-pro]** severity=`major`, category=`bug`

`ExecuteLlmSelectionCardActionAsync` catches `InvalidOperationException | ArgumentException` but not `NotSupportedException`. However, `ApplyPresetAsync` can throw `NotSupportedException` when it hits an unsupported activation type in the switch default (`DefaultUserLlmSelectionService.cs:662`). An unhandled `NotSupportedException` will bubble up as a 500 instead of a friendly user-facing message.

**Fix:** Add `NotSupportedException` to the exception filter, matching the pattern used in `HandlePresetAsync` in `ModelChannelSlashCommandHandler.cs:246`:
```csharp
catch (Exception ex) when (ex is InvalidOperationException or ArgumentException or NotSupportedException)
```

agents/Aevatar.GAgents.NyxidChat/LlmSelection/DefaultUserLlmOptionsService.cs
```diff
@@ -0,0 +1,106 @@
...
+            return string.Empty;
+
+        var handle = await _broker
+            .IssueShortLivedAsync(query.Subject, new CapabilityScope { Value = StatusScope }, ct)
+            .ConfigureAwait(false);
+        return handle.AccessToken;
+    }
+
+    private async Task<UserLlmOption?> ResolveCurrentAsync(
+        UserLlmOptionsQuery query,
```
eanzhao, line 57: **[glm-5.1]** severity=`minor`, category=`bug`

`catch { return null; }` in `ResolveCurrentAsync` swallows **all** exceptions (including `OutOfMemoryException`-derived via `ThreadAbortException` on .NET Framework, or `OperationCanceledException` which should propagate). Use `catch (Exception) when (ex is not OperationCanceledException)` at minimum, and consider logging the failure rather than silently returning null.

agents/Aevatar.GAgents.NyxidChat/LlmSelection/DefaultUserLlmSelectionService.cs
```diff
@@ -0,0 +1,228 @@
...
+        CancellationToken ct)
+    {
+        ArgumentException.ThrowIfNullOrWhiteSpace(model);
+        return SaveModelOnlyAsync(context, model.Trim(), ct);
+    }
+
+    public async Task ApplyPresetAsync(
+        UserLlmSelectionContext context,
+        string presetId,
+        CancellationToken ct)
```
eanzhao, line 72: **[glm-5.1]** severity=`minor`, category=`bug`

Same issue as above: `catch { return new StudioUserConfig(string.Empty); }` in `ReadCurrentAsync` swallows all exceptions silently. At minimum, narrow to non-critical exceptions and log.

agents/Aevatar.GAgents.NyxidChat/LlmSelection/DefaultUserLlmSelectionService.cs
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    public async Task SetByServiceAsync(
+        UserLlmSelectionContext context,
+        string serviceId,
+        string? modelOverride,
+        CancellationToken ct)
+    {
+        ArgumentException.ThrowIfNullOrWhiteSpace(serviceId);
+
```
eanzhao, line 44: **[glm-5.1]** severity=`minor`, category=`arch`

`ToOption`, `NormalizeRequired`, and `NormalizeOptional` are copy-pasted verbatim from `DefaultUserLlmOptionsService`. Extract these to a shared static helper (e.g., `NyxIdLlmServiceMapping`) to avoid divergence.

agents/Aevatar.GAgents.NyxidChat/LlmSelection/NyxIdLlmServiceCatalogClient.cs
```diff
@@ -0,0 +1,252 @@
...
+            ? property.ValueKind switch
+            {
+                JsonValueKind.True => true,
+                JsonValueKind.False => false,
+                _ => null,
+            }
+            : null;
+
+    private static int? TryReadInt(JsonElement element, string propertyName) =>
+        element.TryGetProperty(propertyName, out var property) && property.ValueKind == JsonValueKind.Number
```
eanzhao: **[glm-5.1]** severity=`minor`, category=`bug`

`ParsePreset` uses `ReadRequiredString(element, "id")` for the fallback `Title`, which will throw `InvalidOperationException` if `title` is missing but `id` is present. The intent is to use `id` as the fallback title, but the call `ReadOptionalString(element, "title") ?? ReadRequiredString(element, "id")` re-validates `id` unnecessarily — the `Id` property already validated it. Simplify to `ReadOptionalString(element, "title") ?? id` to avoid a redundant validation that could throw on the same element.

agents/Aevatar.GAgents.NyxidChat/LlmSelection/DefaultUserLlmOptionsService.cs
```diff
@@ -0,0 +1,106 @@
...
+
+    public async Task<UserLlmOptionsView> GetOptionsAsync(UserLlmOptionsQuery query, CancellationToken ct)
+    {
+        ArgumentNullException.ThrowIfNull(query);
+
+        var accessToken = await IssueAccessTokenAsync(query, ct).ConfigureAwait(false);
+        var catalog = await _catalogClient.GetServicesAsync(query, accessToken, ct).ConfigureAwait(false);
+        var available = catalog.Services.Select(ToOption).ToArray();
+        var current = await ResolveCurrentAsync(query, available, ct).ConfigureAwait(false);
+        var setupHint = available.Length == 0
```
eanzhao: **[kimi]** severity=`minor`, category=`perf`

When `catalog.SetupHint` is null and `available.Length == 0`, this falls back to `_catalogClient.GetSetupHintAsync`, which internally calls `GetServicesAsync` again—issuing a redundant identical HTTP request to the same endpoint. Since `GetServicesAsync` already returned the full response, this second round-trip is wasteful. Remove the fallback or make `GetSetupHintAsync` a no-op/client-side pass-through.

agents/platforms/Aevatar.GAgents.Platform.Telegram/TelegramMessageComposer.cs
```diff
@@ -120,6 +140,29 @@ private static string BuildRenderedText(MessageContent intent, int maxLength)
         return textInfo.SubstringByTextElements(0, maxLength);
     }
 
+    private static ActionElement[] GetButtonActions(MessageContent intent) =>
+        intent.Actions
+            .Where(static action => action.Kind == ActionElementKind.Button && !string.IsNullOrWhiteSpace(action.Label))
```
eanzhao: **[gemini]** severity=`minor`, category=`design`

`GetButtonActions` only includes `ActionElementKind.Button`. Telegram supports `url` buttons, so this should also include `ActionElementKind.Link` to ensure that setup and documentation links (like the 'Go to NyxID' button in LLM options) are rendered as native buttons.

tools/Aevatar.Tools.Cli/Hosting/ModelCommandHandler.cs
```diff
@@ -0,0 +1,292 @@
...
+        string ServiceSlug,
+        string DisplayName,
+        string? DefaultModel,
+        string Status,
+        string Source);
+
+    private sealed record CliLlmSetupHint(string? SetupUrl, IReadOnlyList<CliLlmPreset> Presets);
+
+    private sealed record CliLlmPreset(string Id, string Title);
+}
```
eanzhao, line 319: **[v4-pro]** severity=`minor`, category=`design`

`UseAsync` resolves an option by number/name but does not check `Allowed` or `Status == "ready"` before calling the save API. The user gets no CLI-side feedback — the server rejects it with an error the user only sees as a generic failure. Add pre-flight validation and a friendly message ("LLM service 'X' is not ready/allowed") before the API call.

agents/Aevatar.GAgents.NyxidChat/LlmSelection/DefaultUserLlmSelectionService.cs
```diff
@@ -0,0 +1,228 @@
...
+            throw new InvalidOperationException($"LLM preset '{presetId}' is not available.");
+
+        switch (preset.Activation)
+        {
+            case UseExistingService existing:
+                await SaveAsync(
+                    context,
+                    existing.RouteValue,
+                    existing.DefaultModel ?? string.Empty,
+                    ct).ConfigureAwait(false);
```
eanzhao: **[gemini]** severity=`nit`, category=`bug`

In `ApplyPresetAsync`, if `provisionedOption.DefaultModel` is null, the code saves `string.Empty` to UserConfig. While safe, it might be better to preserve the existing `DefaultModel` from UserConfig if the preset doesn't specify one.

eanzhao:
## Multi-model fix-check verdict (`/opencode-pr-fix-check`) — HEAD `dbafbbe0`

Six reviewers (deepseek-v4-pro, glm-5.1, mimo-v2.5-pro, kimi, gemini, codex) independently re-evaluated each of the 21 prior inline comments against the current post-fix diff.

**Result: 21 resolved | 0 partial | 0 unresolved | 0 cannot_determine**

17 / 21 are unanimous (6/6). Four are 5/6 — three of those dissents are non-actionable (procedural / verified-wrong); one is a real design observation worth flagging.

### Real design caveat (codex, on both Telegram callback_data rows: #3173706286, #3174021823)
The 64-byte cap is enforced with compact `a`/`s`/`v` keys and progressive stripping — Telegram rejection is fully addressed. **Caveat:** when the compact payload is still too large, the fallback drops `v` (submitted value) and then `s` (service slug), so a long LLM service id can produce a button without the actual selected value, which would degrade UX silently. Codex suggests an alternative: short callback nonce that maps to the full payload server-side. Current behavior is graceful but lossy; whether to address this depends on observed LLM service id lengths.

### Non-actionable dissents (for completeness)
- **#3174021950** (`NyxIdApiClient.cs:335`, deepseek=cannot_determine) — procedural CD; deepseek couldn't verify NyxID server deployment state. The 404→`/status` fallback is verified in place across `NyxIdApiClient`, `NyxIdLlmCatalogHttpClient`, and `StreamingProxyNyxParticipantCoordinator.FetchReadyProvidersAsync` (which was the gap codex flagged on the prior round — now closed).
- **#3174022597** (`ChannelConversationTurnRunner.cs:517`, glm-5.1=unresolved) — glm misread the catch filter. Verified at line 646: `when (ex is InvalidOperationException or ArgumentException or NotSupportedException)`. The other 5 reviewers (including codex) confirmed resolved.

---

<sub>Generated by `/opencode-pr-fix-check` against HEAD `dbafbbe0`. All 6 reviewers parsed cleanly. Two minified bundle files (`demos/.../wwwroot/app.js`, `tools/.../wwwroot/app.js`) were excluded from the diff sent to reviewers — no inline comments targeted them.</sub>

PR #558, opened 2026-05-02
Fix NyxID LLM route scope

eanzhao:
## Summary

- fix NyxID LLM route/model selection to request the OAuth `proxy` scope instead of `llm:status` / `llm:proxy`
- add `proxy` to the Aevatar broker OAuth client DCR and authorization scope so new `/init` bindings can mint tokens accepted by NyxID LLM routes
- detect legacy OAuth client registrations missing the required scope and re-run DCR from the bootstrap actor
- add `/route` as an alias and support `/route use <service> <model>` such as `/route use chrono-llm gpt-5.5`

## Root Cause

NyxID broker binding token exchange validates requested scopes against the scopes captured in the original OAuth binding. The previous implementation requested `llm:status` for `/model` catalog reads, but existing bindings only allowed the scopes requested by `/init`, which did not include `llm:status`. NyxID LLM routes accept `proxy`, `proxy:*`, or `llm:proxy`, so the Aevatar side should request `proxy` and make sure `/init` captures that scope.

Existing user bindings created before this change still need to run `/init` again so their binding includes `proxy`.

## Validation

- `dotnet test test/Aevatar.GAgents.ChannelRuntime.Tests/Aevatar.GAgents.ChannelRuntime.Tests.csproj --nologo --no-restore --filter "FullyQualifiedName~ModelSlashCommandHandlerTests|FullyQualifiedName~NyxIdRemoteCapabilityBrokerTests|FullyQualifiedName~NyxIdDynamicClientRegistrationClientTests|FullyQualifiedName~AevatarOAuthClientGAgentTests"`
- `bash tools/ci/test_stability_guards.sh`
- `bash tools/ci/architecture_guards.sh`
- `git diff --check`

PR #374, opened 2026-04-24
Progressive Lark reply via NyxID relay edit-message (#352)

eanzhao:
Closes aevatarAI/aevatar#352.

Consumes the new `POST /api/v1/channel-relay/reply/update` endpoint shipped in ChronoAIProject/NyxID#480 (merged as ChronoAIProject/NyxID#483) so Lark bot replies render progressively instead of leaving the user in a 5–30s silent wait.

**Redesigned after PR #368 merge**: PR #368 moved the reply token from `OutboundDeliveryContext` (serializable) into `ConversationGAgent._nyxRelayReplyTokens` (actor-owned in-memory). This invalidated the earlier design where the inbox runtime called the outbound port directly, because the inbox runtime is a stream subscriber and has no access to the actor-scoped token. The current design flips the responsibility: the inbox runtime only signals deltas, and the actor is the sole caller of the outbound port.

## Flow

```
Inbox runtime (async LLM, no outbound port access):
  LLM delta → TurnStreamingReplySink.OnDeltaAsync (throttle 750ms)
    → actor.HandleEventAsync(LlmReplyStreamChunkEvent)
  LLM stream ends → TurnStreamingReplySink.FinalizeAsync (bypass throttle)
    → actor.HandleEventAsync(LlmReplyStreamChunkEvent) final flush
    → actor.HandleEventAsync(LlmReplyReadyEvent)

ConversationGAgent (single-threaded turn, owns reply token + streaming state):
  HandleLlmReplyStreamChunkAsync:
    - resolve runtimeContext with reply token (same path as HandleLlmReplyReadyAsync)
    - read per-correlation streaming state (PlatformMessageId, Disabled, EditCount)
    - first chunk → runner.RunStreamChunkAsync(chunk, null) → placeholder send
    - subsequent → runner.RunStreamChunkAsync(chunk, PMID) → edit
    - any failure → mark state.Disabled, drop further chunks
  HandleLlmReplyReadyAsync:
    - if streaming state healthy and LLM completed:
        force final update if text changed since last flush,
        persist ConversationTurnCompletedEvent directly (no re-send)
    - otherwise: existing RunLlmReplyAsync fallback path
    - cleanup streaming state + reply token in both paths
```

## Architectural rules respected (CLAUDE.md)

| Rule | How |
|------|-----|
| Reply token must stay in actor | Inbox runtime can't call port; only dispatches `EventEnvelope`s. Actor's `_nyxRelayReplyTokens` stays the only authoritative store. |
[...]

PR #561, opened 2026-05-03
Self-heal binding rejections + streaming-failed reply-token loop

eanzhao:
## Production failure modes (observed 2026-05-03)

Two independent symptoms left a Lark user visibly stuck even though the cluster was otherwise healthy. Both healed inside aevatar — no NyxID changes.

### #1 — `/llm` says "binding 已失效, 请 /init" while `/init` refuses with "已绑定"

After [PR #558](https://github.com/aevatarAI/aevatar/pull/558) re-DCR'd the cluster's OAuth client to add the `proxy` scope, the user's existing NyxID binding (issued for the previous `client_id`) was rejected at broker token-exchange. `ModelChannelSlashCommandHandler` caught the rejection and pointed the user at `/init` — which then refused because the local readmodel still holds the (now-dead) `binding_id`. The user loops forever.

Pod log evidence:
```
oauth_broker_service.rs: "Scope 'proxy' is not allowed for this binding"
ModelChannelSlashCommandHandler -> "当前 NyxID 绑定已失效, 请先发送 /init 重新绑定"
InitChannelSlashCommandHandler  -> "已绑定 NyxID 账号, 需要切换账号请先发送 /unbind 再发送 /init"
```

### #2 — Bot replies `...` forever after LLM failure

The streaming sink fires the first chunk via `channel-relay/reply`, consuming the single-use reply token and placing a `...` placeholder. If the LLM then fails (e.g. upstream 429 `usage_limit_reached`), pre-fix the runtime fell through to `RunLlmReplyAsync` which issued a fresh `/reply` against the dead token:

```
NyxID API request failed: POST .../api/v1/channel-relay/reply -> 401
body={"error":"unauthorized","message":"Unauthorized: Reply token already used"}
```

User stares at `...` with no error message ever appearing.

### #1 — `ModelChannelSlashCommandHandler` self-heals stale local binding

When the broker throws `BindingRevokedException` / `BindingNotFoundException` / `BindingScopeMismatchException`, the handler now dispatches `RevokeBindingCommand` to the local `ExternalIdentityBindingGAgent` (reason `auto_self_heal_*`) so the readmodel flips to revoked, and tells the user the binding was cleared and `/init` will work next.

Mirrors the dispatch shape `UnbindChannelSlashCommandHandler` uses for explicit `/unbind`, but skips the NyxID-side revoke (NyxID is the one that just told us the binding is gone). When `IActorRuntime` is not registered (some demo / CLI hosts), degrades to "tell the user, hope they `/unbind`" rather than crashing the slash.

### #2 — `ConversationGAgent.TryCompleteStreamedReplyAsync` handles the `Failed` branch

When streaming has already committed the placeholder (`state.PlatformMessageId` set, not `Disabled`), the `Failed` branch now edits the placeholder in place via `RunStreamChunkAsync` (which uses `channel-relay/reply/update` — no reply token needed) with the classified failure text, then persists `ConversationTurnCompletedEvent` so the runtime envelope retry loop does not refire and consume the dead token again.

If the edit also fails (rare — Lark can refuse stale-message edits), persists the last flushed partial as terminal: same defence-in-depth pattern PR #374 already established for the `Completed` path.

## Tests

- **4 new** `ModelSlashCommandHandlerTests` pinning the binding self-heal for each rejection shape (`scope_mismatch`, `remote_revoked`, `remote_not_found`) plus the degraded-path when `IActorRuntime` is not registered. Asserts a `RevokeBindingCommand` envelope is dispatched to the local actor with the correct subject + reason code.
- **2 new** `ConversationGAgentDedupTests`:
  - `WhenStreamingStartedThenLlmFailed_EditsPlaceholderInsteadOfReusingToken` — pins the placeholder edit fires + `RunLlmReplyAsync` is NOT invoked + persisted outbound reflects the failure text.
  - `WhenStreamingStartedAndFailedEditAlsoFails_PersistsLastFlushedAsTerminalWithoutReusingToken` — pins the defence-in-depth fallback when the edit attempt is also rejected.

## Verification

- `dotnet test test/Aevatar.GAgents.ChannelRuntime.Tests/Aevatar.GAgents.ChannelRuntime.Tests.csproj` (851/851)
- `dotnet test test/Aevatar.GAgents.Channel.Protocol.Tests/Aevatar.GAgents.Channel.Protocol.Tests.csproj --filter ~ConversationGAgentDedupTests` (36/36)
- `dotnet test test/Aevatar.Foundation.Core.Tests/Aevatar.Foundation.Core.Tests.csproj` (230/230)

## Refs

- Issue #549 (binding readmodel + projection scope; same root family)
- PR #558 (added `proxy` scope to DCR — exposed the binding-revoked loop here)
- PR #374 (original streaming-final-edit defence-in-depth pattern this PR extends to the Failed branch)

### DIFF
--- a/agents/Aevatar.GAgents.NyxidChat/Slash/ModelChannelSlashCommandHandler.cs
+++ b/agents/Aevatar.GAgents.NyxidChat/Slash/ModelChannelSlashCommandHandler.cs
@@ -1,8 +1,11 @@
+using Aevatar.Foundation.Abstractions;
 using Aevatar.GAgents.Channel.Abstractions;
 using Aevatar.GAgents.Channel.Abstractions.Slash;
+using Aevatar.GAgents.Channel.Identity;
 using Aevatar.GAgents.Channel.Identity.Abstractions;
 using Aevatar.GAgents.NyxidChat.LlmSelection;
 using Aevatar.Studio.Application.Studio.Abstractions;
+using Google.Protobuf.WellKnownTypes;
 using Microsoft.Extensions.Logging;
 
 namespace Aevatar.GAgents.NyxidChat.Slash;
@@ -18,18 +21,21 @@ public sealed class ModelChannelSlashCommandHandler : IChannelSlashCommandHandle
     private readonly IUserLlmOptionsService? _optionsService;
     private readonly IUserLlmSelectionService? _selectionService;
     private readonly IUserLlmOptionsRenderer<MessageContent>? _renderer;
+    private readonly IActorRuntime? _actorRuntime;
     private readonly ILogger<ModelChannelSlashCommandHandler> _logger;
 
     public ModelChannelSlashCommandHandler(
         ILogger<ModelChannelSlashCommandHandler> logger,
         IUserLlmOptionsService? optionsService = null,
         IUserLlmSelectionService? selectionService = null,
-        IUserLlmOptionsRenderer<MessageContent>? renderer = null)
+        IUserLlmOptionsRenderer<MessageContent>? renderer = null,
+        IActorRuntime? actorRuntime = null)
     {
         _logger = logger ?? throw new ArgumentNullException(nameof(logger));
         _optionsService = optionsService;
         _selectionService = selectionService;
         _renderer = renderer;
+        _actorRuntime = actorRuntime;
     }
 
     public string Name => "model";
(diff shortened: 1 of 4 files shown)


agents/Aevatar.GAgents.NyxidChat/Slash/ModelChannelSlashCommandHandler.cs
```diff
@@ -93,6 +111,76 @@ await _optionsService.GetOptionsAsync(BuildQuery(context, bindingId), ct).Config
...
+        }
+        catch (Exception ex) when (!ct.IsCancellationRequested)
+        {
+            _logger.LogError(ex,
+                "/model failed to self-heal local binding actor={ActorId} after NyxID-side rejection: reason={Reason}",
+                actorId,
+                reason);
+        }
+
+        return new MessageContent { Text = userMessage };
```
eanzhao: 这里不应该无条件返回带“本地已自动清理”的 `userMessage`。如果 `_actorRuntime` 缺失，或者 `CreateAsync` / `HandleEventAsync` 在上面的 catch 里失败，本地 binding actor/readmodel 实际上没有被 revoke，用户按提示去 `/init` 仍然会看到“已绑定”，回到这个 PR 要修的死循环。建议只在 dispatch 成功后返回“已自动清理”；失败路径返回明确的 fallback（例如提示重试 `/model` 或先 `/unbind`），并考虑像 `/unbind` 一样 retry 一次本地 revoke。测试也应覆盖 runtime 缺失/dispatch 失败时不声称已清理。

PR #566, opened 2026-05-04
Gracefully handle missing scope in UserMemoryStore (fixes #564)

eanzhao:
## Summary

- `ActorBackedUserMemoryStore` throws `InvalidOperationException` when no HTTP auth context is available (Orleans actor / channel conversation)
- Adds `TryResolveScopeId` / `TryResolveWriteActorId` that return null instead of throwing
- `ReadProjectedStateAsync` returns null when scope is unavailable → `GetAsync` returns `Empty` → `BuildPromptSectionAsync` returns empty string silently
- Write operations still throw since they only run from Studio API with auth context
- Eliminates the recurring warning on every LLM turn in channel conversations

## Verification

- `dotnet build` — 0 errors
- `dotnet test test/Aevatar.AI.Tests` — 500 passed, 0 failed

Closes #564

### DIFF
--- a/src/Aevatar.Studio.Infrastructure/ActorBacked/ActorBackedUserMemoryStore.cs
+++ b/src/Aevatar.Studio.Infrastructure/ActorBacked/ActorBackedUserMemoryStore.cs
@@ -198,7 +198,10 @@ internal sealed class ActorBackedUserMemoryStore : IUserMemoryStore
 
     private async Task<UserMemoryState?> ReadProjectedStateAsync(CancellationToken ct)
     {
-        var actorId = ResolveWriteActorId();
+        var actorId = TryResolveWriteActorId();
+        if (actorId is null)
+            return null;
+
         var document = await _documentReader.GetAsync(actorId, ct);
         if (document?.StateRoot == null ||
             !document.StateRoot.Is(UserMemoryState.Descriptor))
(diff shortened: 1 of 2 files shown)


eanzhao:
## Verdict

The fix **correctly addresses issue #564** — `GetAsync` and `BuildPromptSectionAsync` no longer throw on nil scope, eliminating the recurring warning log. The approach of adding `TryResolve*` variants and guarding the read path is sound and minimal.

However, there are a few things that should be addressed before merging:

---

### 1. `RemoveEntryAsync` silently returns false — inconsistent with PR description

The PR body states:

> Write operations (Save/Add/Remove) still throw since they only run from Studio API with auth context

But in practice `RemoveEntryAsync` does NOT throw when scope is nil — it calls `ReadProjectedStateAsync` (now Try*) and hits the `state is null` short-circuit, returning `false`:

```csharp
// ActorBackedUserMemoryStore.cs:126-131
public async Task<bool> RemoveEntryAsync(string id, CancellationToken ct = default)
{
    var state = await ReadProjectedStateAsync(ct);
    if (state is null || !state.Entries.Any(...))
        return false;  // <— returns false without throwing
    ...
}
```

`SaveAsync` and `AddEntryAsync` still throw (via `EnsureWriteActorAsync`), so the inconsistency is real. Either:
- Make `RemoveEntryAsync` consistently throw by guarding the write path separately from the read check, or  
- Update the PR description to reflect what actually happens.

In practice this is harmless since `RemoveEntryAsync` is only called from Studio API paths (which have scope), but it's worth fixing for clarity.

### 2. Missing test for nil-scope read path

The key behavioral change — `GetAsync` returns `Empty` (not throws) when scope is nil — has no test coverage. The existing `UserMemoryStore_NoScope_Throws` test only covers `AddEntryAsync` (write path). Consider adding something like:

```
UserMemoryStore_GetAsync_NoScope_ReturnsEmpty
UserMemoryStore_BuildPromptSectionAsync_NoScope_ReturnsEmpty
```

### 3. (Minor) `BuildPromptSectionAsync` try-catch is now dead for scope errors

After this fix, scope errors never reach the catch block in `BuildPromptSectionAsync`. The catch is still useful as defense-in-depth for projection store failures, but it's worth noting that the `"Failed to load user memory for prompt injection"` log message will now only fire for non-scope errors, which might be misleading when debugging.

---

Overall the fix is correct and does its job. The issues above are quality/clarity concerns rather than functional bugs.

PR #569, opened 2026-05-05
fix(skill-runner): treat zero-tool-call runs as failure for fetch-and-summarize skills (#439)

eanzhao:
## Summary

Closes the remaining gap in issue #439's fake-success failure mode that PR #471 review (SkillRunnerGAgent.cs:407) flagged: when the LLM bypasses `nyxid_proxy` entirely and produces output from prior context, both `failureCount` and `successCount` stay at zero, and the runner-layer safety net waved the run through as clean success.

PR #471 covered the **all-fail** mode (every `nyxid_proxy` call returned 4xx/5xx). This PR covers the **never-called** mode (the LLM hallucinated a report with no tool errors to count) for fetch-and-summarize skills like `daily_report`.

## Approach

A per-skill `RequiresNyxidProxySuccess` flag carried on `SkillRunnerState` / `InitializeSkillRunnerCommand` / `SkillRunnerInitializedEvent` (proto field 21 / 16 / 16). When set, `EnsureToolStatusAllowsCompletion` throws on `successCount == 0` regardless of `failureCount`, routing the run through `HandleTriggerAsync`'s existing catch path so `/agent-status` reports `error_count > 0` and a meaningful `last_error`.

- `daily_report` template opts in (the original #439 symptom was on this skill).
- Skills that don't depend on tool data (future pure-LLM transformations) leave the flag false and pass through unchanged.
- All-fail and never-called branches don't conflict: when both would fire, the all-fail message wins because it names the failed-call count, which is more actionable for the operator.

## Why a flag, not always-on

`Policy_NoToolCallsAtAll_FlagOff_Allows` exists for a reason: the SkillRunner is a generic scheduled-skill runtime, and a pure-LLM skill that legitimately fans out zero tool calls would be falsely failed by an unconditional "success ≥ 1" rule. The flag makes the contract explicit at the template-spec level — the place where "this is a fetch-and-summarize skill" semantics already live (next to `RequiredServiceSlugs`).

## Affected code

- `agents/Aevatar.GAgents.Scheduled/protos/skill_runner.proto` — new `requires_nyxid_proxy_success` field on `SkillRunnerState` (#21), `InitializeSkillRunnerCommand` (#16), `SkillRunnerInitializedEvent` (#16)
- `agents/Aevatar.GAgents.Scheduled/SkillRunnerGAgent.cs` — `HandleInitializeAsync` and `ApplyInitialized` plumb the flag, `ExecuteSkillAsync` passes `State.RequiresNyxidProxySuccess` to the safety net, `EnsureToolStatusAllowsCompletion` enforces it
- `agents/Aevatar.GAgents.Authoring.Lark/AgentBuilderTemplates.cs` — `DailyReportTemplateSpec` adds `RequiresNyxidProxySuccess: true`
- `agents/Aevatar.GAgents.Authoring.Lark/AgentBuilderTool.cs` — propagates the spec field onto `InitializeSkillRunnerCommand`
- `test/Aevatar.GAgents.ChannelRuntime.Tests/SkillRunnerToolFailureSafetyNetTests.cs` — split `Policy_NoToolCallsAtAll_*` into flag-on (throws) / flag-off (allows), pin all-fail-message-wins ordering when both branches would fire, plus three flag-on counterparts of the existing policy tests
- `test/Aevatar.GAgents.ChannelRuntime.Tests/AgentBuilderToolTests.cs` — pin that `daily_report` spec carries `RequiresNyxidProxySuccess: true`

## Test plan

- [x] `dotnet build aevatar.slnx --nologo` — 0 errors, 154 warnings (all pre-existing)
- [x] `dotnet test test/Aevatar.GAgents.ChannelRuntime.Tests/Aevatar.GAgents.ChannelRuntime.Tests.csproj` — **863 passed, 0 failed** (full suite)
- [x] `dotnet test … --filter "FullyQualifiedName~SkillRunnerToolFailureSafetyNetTests|FullyQualifiedName~AgentBuilderToolTests"` — 75 passed
- [x] All backend-relevant `architecture_guards.sh` checks pass (proto lint, projection state version, projection route mapping, channel guards, workflow guards, etc.). The playground asset drift guard requires the unrelated pnpm/vite frontend build (no frontend files in this diff — same constraint noted in PR #471's test plan).

## Issue acceptance criteria revisited (#439)

The original four boxes were already ticked by PR #458 (prompt) + PR #471 (tool/runner). This PR plugs the gap that PR #471 review explicitly noted as out of scope at the time:

- Original symptom (52 commits → "No meaningful public GitHub activity") could occur via either (a) all `nyxid_proxy` calls failing or (b) the LLM never calling `nyxid_proxy` at all. PR #471 caught (a). This PR catches (b) for daily_report.

### DIFF
--- a/agents/Aevatar.GAgents.Scheduled/SkillRunnerGAgent.cs
+++ b/agents/Aevatar.GAgents.Scheduled/SkillRunnerGAgent.cs
@@ -414,6 +414,18 @@ public sealed class SkillRunnerGAgent : AIGAgentBase<SkillRunnerState>
     /// </summary>
     private SkillRunnerStreamingReplySink? TryCreateStreamingSink()
     {
+        // Issue #439 (PR #569 review, codex P1 on SkillRunnerGAgent.cs:351): when the run
+        // is gated by EnsureToolStatusAllowsCompletion (RequiresNyxidProxySuccess set),
+        // streaming each delta would POST/PUT the partial text to Lark live — i.e. a
+        // hallucinated daily report would already be visible in the user's DM by the
+        // time the guard fires, and each retry would repost it. Disable live streaming
+        // for those skills so the message only POSTs through the chunked-dispatch path
+        // AFTER the guard has confirmed at least one nyxid_proxy success. Trade-off: the
+        // user no longer sees the report grow live, but output integrity wins over the
+        // streaming-edit UX for fetch-and-summarize skills.
+        if (State.RequiresNyxidProxySuccess)
+            return null;
+
         var client = _nyxIdApiClient ?? Services.GetService<NyxIdApiClient>();
         if (client is null)
         {
@@ -831,7 +843,15 @@ public sealed class SkillRunnerGAgent : AIGAgentBase<SkillRunnerState>
         next.ScopeId = evt.ScopeId ?? string.Empty;
         next.ProviderName = NormalizeProviderName(evt.ProviderName);
         next.Model = evt.Model ?? string.Empty;
-        next.RequiresNyxidProxySuccess = evt.RequiresNyxidProxySuccess;
+        // Legacy actors created before proto field 16 existed replay an init event whose
+        // RequiresNyxidProxySuccess deserializes as false, which would let them keep the
+        // pre-#439 zero-tool-call fake-success path — making post-fix behavior depend on
+        // creation time rather than template semantics. Derive the effective flag from
+        // the template name so known fetch-and-summarize skills get the safety net on
+        // replay regardless of when the actor was created. New templates that need this
+        // protection should be added to RequiresProxySuccessByTemplate.
+        next.RequiresNyxidProxySuccess = evt.RequiresNyxidProxySuccess
+            || RequiresProxySuccessByTemplate(evt.TemplateName);
 
         // Missing sampling fields intentionally use upstream model defaults;
         // missing runner limits fall back to SkillRunner defaults.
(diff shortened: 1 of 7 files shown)


agents/Aevatar.GAgents.Scheduled/SkillRunnerGAgent.cs
```diff
@@ -794,6 +831,7 @@ private static SkillRunnerState ApplyInitialized(SkillRunnerState current, Skill
         next.ScopeId = evt.ScopeId ?? string.Empty;
         next.ProviderName = NormalizeProviderName(evt.ProviderName);
         next.Model = evt.Model ?? string.Empty;
+        next.RequiresNyxidProxySuccess = evt.RequiresNyxidProxySuccess;
```
eanzhao: **[codex]** severity=major, category=bug

Older daily_report actors replay `SkillRunnerInitializedEvent` records that do not contain field 16, so `evt.RequiresNyxidProxySuccess` deserializes to `false` and those existing runners still allow the 0/0 fake-success path.

Derive an effective default for known fetch-and-summarize templates during state application/execution, e.g. set this true when `TemplateName == "daily_report"` unless an explicit migration/backfill guarantees all existing runners get the new field.

Concretely, since `SkillRunnerState.template_name` already exists at field 2:

```csharp
next.RequiresNyxidProxySuccess = evt.RequiresNyxidProxySuccess
    || evt.TemplateName == "daily_report";
```

would close the gap for already-deployed actors on replay without requiring schema migration.

PR #570, opened 2026-05-05
Add operator rebuild endpoint to unwedge OAuth client at NyxID (#549)

eanzhao:
## Summary

Issue #549 — production wedge that leaves \`AevatarOAuthClientGAgent\`
pinned to a broken \`client_id\` after K8s rolling deploys. The OCC
absorber landed in commit \`9326204d\` keeps the cluster from deadlocking
on every restart, but it can't replace the wedged snapshot itself. This
PR adds the operator-grade unblock that points \`aevatar\` at a freshly-
created NyxID client without DB access.

- \`POST /api/oauth/aevatar-client/rebuild\` accepts a JSON body
  \`{ client_id, redirect_uri?, oauth_scope?, client_id_issued_at_unix? }\`
  with an \`X-Aevatar-Admin-Token\` header (constant-time compared
  against \`ChannelIdentity:Admin:RebuildToken\`). Empty config →
  fail-secure 503. Wrong/missing token → 401. Bad body → 400.
  Happy path: dispatches \`ProvisionAevatarOAuthClientCommand\`,
  waits ≤30s for the readmodel to reflect the pin, returns 200; if
  the readmodel hasn't caught up, returns 202 \`rebuild_pending_propagation\`.
- \`ProvisionAevatarOAuthClientCommand\` proto gains a \`redirect_uri\`
  field (#5). The handler now writes it to \`AevatarOAuthClientProvisionedEvent\`
  and treats \`(client_id, authority, redirect_uri, oauth_scope)\` as
  the same-snapshot key. Pre-fix the check was \`(client_id, authority)\`
  only, so a partial heal silently no-op'd and let the next bootstrap
  pass detect drift and re-DCR away the operator's pinned id.
- \`AevatarOAuthAdminOptions\` binds \`ChannelIdentity:Admin\`. Production
  sets \`ChannelIdentity__Admin__RebuildToken\` via env; tests/dev
  clusters leave it unset and the endpoint stays disabled.

Other layers from issue #549 (distributed-lock bootstrap, sync grain
RPC dispatch, \`ProjectionScopeGAgentBase\` OCC absorber, multi-silo
CI test) are intentionally out of scope and tracked as follow-ups —
the OCC absorber already keeps the cluster live; this PR closes the
loop on the ops handoff that turns the new \`17cecaad-…\` NyxID client
into the cluster's authoritative snapshot.

## Test plan
- [x] \`dotnet build agents/Aevatar.GAgents.Channel.Identity/Aevatar.GAgents.Channel.Identity.csproj\` — clean.
- [x] \`dotnet test test/Aevatar.GAgents.ChannelRuntime.Tests --filter "Identity"\` — 27/27 identity tests pass (3 new HandleProvision unit tests + 6 new endpoint tests + existing suites).
- [x] \`dotnet test test/Aevatar.GAgents.ChannelRuntime.Tests\` — 869/869 pass.
- [x] \`bash tools/ci/test_stability_guards.sh\` — passes (no polling waits added beyond the operator-grade observation loop).
- [ ] Production rollout: set \`ChannelIdentity__Admin__RebuildToken\`, then \`curl -H "X-Aevatar-Admin-Token: …" -d '{"client_id":"17cecaad-214b-4521-9dba-d435462e4095"}' POST /api/oauth/aevatar-client/rebuild\`. Verify \`/api/oauth/aevatar-client/status\` reflects the new id with no drift.

### DIFF
--- a/agents/Aevatar.GAgents.Channel.Identity/Endpoints/IdentityOAuthEndpoints.cs
+++ b/agents/Aevatar.GAgents.Channel.Identity/Endpoints/IdentityOAuthEndpoints.cs
@@ -1,3 +1,5 @@
+using System.Security.Cryptography;
+using System.Text;
 using Aevatar.Foundation.Abstractions;
 using Aevatar.Foundation.Core;
 using Aevatar.GAgents.Channel.Abstractions;
@@ -9,6 +11,7 @@ using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Routing;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
 
 namespace Aevatar.GAgents.Channel.Identity.Endpoints;
 
@@ -19,6 +22,8 @@ namespace Aevatar.GAgents.Channel.Identity.Endpoints;
 public static class IdentityOAuthEndpoints
 {
     private static readonly TimeSpan ProjectionWaitTimeout = TimeSpan.FromSeconds(3);
+    private static readonly TimeSpan RebuildObservationTimeout = TimeSpan.FromSeconds(30);
+    private static readonly TimeSpan RebuildObservationPollDelay = TimeSpan.FromMilliseconds(250);
     private const int MaxWebhookBodyBytes = 64 * 1024;
 
     public static IEndpointRouteBuilder MapIdentityOAuthEndpoints(this IEndpointRouteBuilder app)
@@ -32,6 +37,14 @@ public static class IdentityOAuthEndpoints
             .WithTags("ChannelIdentity")
             .AllowAnonymous();
         app.MapGet("/api/oauth/aevatar-client/status", HandleAevatarOAuthClientStatusAsync)
+            .WithTags("ChannelIdentity")
+            .AllowAnonymous();
+        // Operator-only: rebuild the cluster-singleton OAuth client snapshot
+        // to point at an admin-supplied client_id (issue #549 production
+        // unblock). Auth is by static admin token header — see
+        // AevatarOAuthAdminOptions. AllowAnonymous because the auth check is
+        // done inline; no ASP.NET auth handler is wired for this module.
+        app.MapPost("/api/oauth/aevatar-client/rebuild", HandleAevatarOAuthClientRebuildAsync)
             .WithTags("ChannelIdentity")
             .AllowAnonymous();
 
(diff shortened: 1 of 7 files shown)


agents/Aevatar.GAgents.Channel.Identity/Endpoints/IdentityOAuthEndpoints.cs
```diff
@@ -328,6 +341,241 @@ internal static async Task<IResult> HandleAevatarOAuthClientStatusAsync(
...
+            {
+                // Projection has not yet materialized the very first state
+                // root for this actor — possible on a brand-new cluster
+                // where rebuild is the first provisioning event.
+            }
+
+            await Task.Delay(pollDelay, ct).ConfigureAwait(false);
+        }
+        return null;
+    }
```
eanzhao, line 598: **[Consensus: 4 models]** severity=major, category=security

Four reviewers flag the new `ConstantTimeEquals` helper. The substantive issues:
1. **Length leak**: both inputs are converted to `byte[]` via `Encoding.UTF8.GetBytes` before `FixedTimeEquals` is invoked. `FixedTimeEquals` itself returns `false` for unequal-length spans, so the helper's running time depends on the configured-token length — which leaks `RebuildToken.Length` to an attacker who can time the responses.
2. **Heap allocation per call**: `Encoding.UTF8.GetBytes` allocates two `byte[]` per call. This is minor in practice but unnecessary — convert to `Span<byte>` with stackalloc (or pre-encode the configured token once at options-bind time) to keep the comparison purely arithmetic.
3. **Early `null` return** breaks the constant-time contract for that input shape; the call site today guards via `TryGetValue` (`presented` is non-null), but the helper signature advertises constant-time semantics that future callers may rely on.

**Note on @kimi's specific claim**: `CryptographicOperations.FixedTimeEquals` does NOT throw `CryptographicException` on length mismatch — it returns `false` (per the .NET docs). So the "unhandled exception → HTTP 500 → length leak" failure mode @kimi describes is incorrect. The real length leak is the simpler one above (encoding step before the FixedTime call), which still warrants hardening.

**Suggested fix**: pre-encode the configured token once at `AevatarOAuthAdminOptions` bind time (or in a `Lazy<byte[]>`); for the presented token, use `Encoding.UTF8.GetByteCount` + `Convert.TryFromBase64String`-style stackalloc; then call `FixedTimeEquals` on the spans. Alternatively, since this is an admin break-glass path with a high-entropy token, the team may decide the residual length leak is acceptable and just convert to spans + drop the early-`null` return.

<details>
<summary>Per-model verbatim (v4-pro, mimo-v2.5-pro, kimi, gemini)</summary>

- **v4-pro**: `ConstantTimeEquals` returns `false` immediately when `right is null`, breaking the constant-time guarantee for `null` inputs. The sole call site guards this (`TryGetValue` short-circuits before the call, so `right` is never null), but the method signature advertises constant-time semantics to future callers. Fix: compare against a zero-filled buffer of the same length as `left` (`new byte[leftBytes.Length]`) instead of early-returning, so the timing profile does not distinguish `null` from a wrong-value string.
- **mimo-v2.5-pro**: `ConstantTimeEquals` calls `Encoding.UTF8.GetBytes(left)` and `Encoding.UTF8.GetBytes(right)` on every invocation, allocating heap `byte[]` whose size is derived from the secret length. GC pressure from these allocations is observable and undermines the constant-time guarantee of `FixedTimeEquals`. Fix: use `Span<byte>` with a stackalloc or a pre-allocated buffer, or compare character-by-character with `CryptographicOperations.FixedTimeEquals` over `ReadOnlySpan<char>` → `ReadOnlySpan<byte>` via `MemoryMarshal.AsBytes`.
- **kimi**: `CryptographicOperations.FixedTimeEquals` throws `CryptographicException` when the two spans have different lengths. An attacker sending a token with a different length than the configured token will trigger an unhandled exception (likely HTTP 500), leaking the configured token's length and creating an oracle. **Fix:** check lengths before calling, or catch `CryptographicException` and return `false`.

```csharp
if (right is null || left.Length != right.Length) return false;
return CryptographicOperations.FixedTimeEquals(
    Encoding.UTF8.GetBytes(left),
    Encoding.UTF8.GetBytes(right));
```
- **gemini**: The `ConstantTimeEquals` implementation correctly uses `CryptographicOperations.FixedTimeEquals` for the content comparison, but converting both strings to bytes first means it still leaks the length of the configured token (as `FixedTimeEquals` returns false immediately if lengths differ). For an internal admin token this is likely acceptable, but for higher security contexts, hashing the inputs before comparison would hide the length.
</details>

agents/Aevatar.GAgents.Channel.Identity/Endpoints/IdentityOAuthEndpoints.cs
```diff
@@ -328,6 +341,241 @@ internal static async Task<IResult> HandleAevatarOAuthClientStatusAsync(
...
+            observationPollDelay: RebuildObservationPollDelay,
+            ct);
+
+    /// <summary>
+    /// Implementation seam exposed for tests so the readmodel-propagation
+    /// timeout can be tightened without waiting the full operator-grade
+    /// 30-second budget on every assertion. Production routes call the
+    /// thin overload above with the canonical defaults.
+    /// </summary>
+    internal static async Task<IResult> HandleAevatarOAuthClientRebuildCoreAsync(
```
eanzhao, line 395: **[Consensus: 5 models]** severity=major, category=arch

Five reviewers flag the same architectural concern: the endpoint directly injects `IActorRuntime`, manually constructs `EventEnvelope`, calls `actor.HandleEventAsync`, and then polls the read-model. This violates CLAUDE.md:

- §Actor 设计原则: "endpoint 不直接依赖 `IActorRuntime`"
- §Command / Envelope / Dispatch: "命令骨架内聚 — 标准生命周期 `Normalize -> Resolve Target -> Build Context -> Build Envelope -> Dispatch -> Receipt -> Observe`" + "禁止绕过 publisher 直操底层传输对象"
- §顶级架构约束: "API 仅做宿主与组合，不承载业务编排"

**Important caveat**: @glm-5.1 correctly notes this same pattern already exists in `AevatarOAuthClientBootstrapService` (line 240/255). So the violation is *consistent with the existing codebase*, not newly introduced — but five reviewers independently flagging it suggests the rebuild path is a good wedge to refactor both call sites behind an `IAevatarOAuthClientRebuildService` (or extend `AevatarOAuthClientBootstrapService` with a `RebuildAsync` method) so the endpoint becomes thin: `await rebuildService.RebuildAsync(req, ct)` → DTO mapping → `Results.Ok/Accepted/...`.

**One-pass option for this PR**: extract the dispatch+observe block into a private static helper in this file (cheap), or accept the violation as consistent with existing pattern and file a follow-up ticket to refactor both bootstrap and rebuild together. The author should pick one and say so in the PR description so reviewers know which path was chosen.

<details>
<summary>Per-model verbatim (codex, kimi, mimo-v2.5-pro, gemini, glm-5.1)</summary>

- **codex**: This endpoint now depends directly on `IActorRuntime` and later builds/dispatches the actor envelope itself, which violates CLAUDE.md: Actor design principle (`endpoint 不直接依赖 IActorRuntime`) and Command/Dispatch (`Runtime` lifecycle vs `Dispatch Port` delivery). Move the rebuild flow behind an application/command port such as `IAevatarOAuthClientRebuildCommandService`; the endpoint should only validate HTTP/auth and call that service, while the service uses `IActorRuntime` for activation and `IActorDispatchPort.DispatchAsync` for delivery.
- **kimi**: The endpoint directly injects `IActorRuntime`, manually constructs `EventEnvelope`, and calls `actor.HandleEventAsync`. This violates CLAUDE.md §Actor设计原则 ("endpoint 不直接依赖 `IActorRuntime`") and §Command / Envelope / Dispatch ("禁止绕过 publisher 直操底层传输对象"). The rebuild orchestration should be encapsulated in an application-layer service/port (e.g., `IAevatarOAuthClientAdminService`) that the endpoint invokes; the endpoint should only validate input, map DTOs, and return HTTP results.
- **mimo-v2.5-pro**: The endpoint manually constructs `EventEnvelope`, routes it via `DirectRoute`, calls `actor.HandleEventAsync`, then polls the readmodel for up to 30s. This embeds the full command lifecycle (envelope build → dispatch → observe) in the HTTP handler, violating CLAUDE.md "命令骨架必须内聚". Consider extracting the dispatch+observe logic into a service (e.g. `AevatarOAuthClientRebuildService`) that the endpoint delegates to — matching the pattern of `AevatarOAuthClientBootstrapService`.
- **gemini**: This endpoint performs business orchestration (authority resolution, projection activation, actor command dispatch, and read-model polling) directly in the host layer. While common for admin/operator utilities, it violates the strict layering rule in `CLAUDE.md`: "API 仅做宿主与组合，不承载业务编排" and "endpoint 不直接依赖 IActorRuntime ... 等基础设施抽象". Consider moving this logic to an application-layer service (e.g., `IAevatarOAuthClientAdminService`) in the future.
- **glm-5.1**: This dispatch follows the same `actor.HandleEventAsync` pattern as `AevatarOAuthClientBootstrapService` (line 255), so it's consistent with the codebase. However, that service uses `EnsureAevatarOAuthClientProvisionedCommand` (which goes through the actor's DCR-mediated path), while this endpoint sends `ProvisionAevatarOAuthClientCommand` directly — bypassing the `EnsureProvisioned` → DCR flow entirely. That's the intended design (ops already has the client_id), but consider adding a comment noting that this deliberately skips the DCR mediation layer so a future reader doesn't assume the pattern is identical to bootstrap.
</details>

agents/Aevatar.GAgents.Channel.Identity/Endpoints/IdentityOAuthEndpoints.cs
```diff
@@ -328,6 +341,241 @@ internal static async Task<IResult> HandleAevatarOAuthClientStatusAsync(
...
+        {
+            return Results.BadRequest(new
+            {
+                error = "client_id_required",
+                detail = "Body must include client_id (the NyxID-issued OAuth client_id this cluster should pin to).",
+            });
+        }
+
+        var authority = NyxIdAuthorityResolver.Resolve(logger);
+        var redirectUri = string.IsNullOrWhiteSpace(body.redirect_uri)
```
eanzhao: **[Consensus: 2 models]** severity=major, category=bug

Two reviewers flag the missing `redirect_uri` validation, with two distinct (both real) failure modes:

1. **@codex (drift bug, severity major)**: An operator may pass a `redirect_uri` that differs from `NyxIdRedirectUriResolver.Resolve()`. The endpoint persists it as the authoritative snapshot and the `WaitForRebuildObservedAsync` loop confirms a value that doesn't match what the next bootstrap pass will compute → bootstrap detects drift and re-DCRs the freshly-pinned client, **reintroducing the very wedge this PR exists to fix**. Mitigation: either reject any non-resolver `redirect_uri` (recommended for the operator-grade safety the endpoint advertises), or add a warning header + clear log line if the operator overrides it.
2. **@mimo-v2.5-pro (input validation, severity minor)**: No `Uri.TryCreate` check before the value is persisted. Garbage like `javascript:alert(1)` flows through to the readmodel and the broker. Defense in depth.

**Suggested fix**: after the `body.redirect_uri.Trim()` resolution, run `Uri.TryCreate(redirectUri, UriKind.Absolute, out var uri) && (uri.Scheme == "https" || uri.Scheme == "http")` and 400 on failure. Then compare against `NyxIdRedirectUriResolver.Resolve(logger)` — if different, either reject or log a structured warning so the drift is visible at incident time.

<details>
<summary>Per-model verbatim (codex, mimo-v2.5-pro)</summary>

- **codex**: Accepting an arbitrary `redirect_uri` stores it as the authoritative snapshot and the wait loop then confirms the same value, even if it differs from `NyxIdRedirectUriResolver.Resolve()`. That returns `rebuilt` for a client the broker will reject as drifted and the next bootstrap pass will re-DCR away. Resolve the expected URI first, reject a supplied mismatched `redirect_uri` with 400, or remove the override and always persist the resolver output.
- **mimo-v2.5-pro**: No validation on the operator-supplied `redirect_uri`. An operator could pass an arbitrary string (e.g. `javascript:alert(1)` or a malformed URI) that gets persisted into the actor state and readmodel. Add `Uri.TryCreate(redirectUri, UriKind.Absolute, out _)` validation after trimming, returning 400 on malformed input.
</details>

agents/Aevatar.GAgents.Channel.Identity/Endpoints/IdentityOAuthEndpoints.cs
```diff
@@ -328,6 +341,241 @@ internal static async Task<IResult> HandleAevatarOAuthClientStatusAsync(
...
+            ? NyxIdRedirectUriResolver.Resolve(logger)
+            : body.redirect_uri.Trim();
+        var oauthScope = string.IsNullOrWhiteSpace(body.oauth_scope)
+            ? AevatarOAuthClientScopes.AuthorizationScope
+            : body.oauth_scope.Trim();
+        var issuedAtUnix = body.client_id_issued_at_unix
```
eanzhao: **[codex]** severity=major, category=bug

`client_id_issued_at_unix` is persisted without range validation, but `AevatarOAuthClientProjectionProvider.GetAsync` later calls `DateTimeOffset.FromUnixTimeSeconds(document.ClientIdIssuedAtUnix)`. A bad admin body such as `long.MaxValue` can make the rebuild/status/broker read path throw instead of returning the documented 400 for a bad body. Validate the supplied timestamp with `DateTimeOffset.FromUnixTimeSeconds` (and preferably require a non-negative Unix second) before dispatching, returning 400 on failure.

agents/Aevatar.GAgents.Channel.Identity/Endpoints/IdentityOAuthEndpoints.cs
```diff
@@ -328,6 +341,241 @@ internal static async Task<IResult> HandleAevatarOAuthClientStatusAsync(
...
+        };
+        await actor.HandleEventAsync(provisionEnvelope, ct).ConfigureAwait(false);
+
+        logger.LogWarning(
+            "Operator rebuild dispatched for AevatarOAuthClientGAgent: client_id={ClientId}, authority={Authority}, redirect_uri={RedirectUri}.",
+            body.client_id,
+            authority,
+            redirectUri);
+
+        var observed = await WaitForRebuildObservedAsync(
```
eanzhao, line 530: **[Consensus: 2 models]** severity=minor, category=perf

Two reviewers flag the same operational concern: `WaitForRebuildObservedAsync` holds the HTTP connection open for up to 30s. Many proxies/LBs default to 30–60s idle timeouts (Cloudflare 100s by default; some corporate proxies 10–15s); a 30s server-side wait can race a 30s LB cutoff and the operator sees a 504 even though the rebuild succeeded.

Two viable mitigations: (a) tighten `RebuildObservationTimeout` to ≤15s and document the LB-budget contract in the endpoint summary, or (b) flip the API to fire-and-forget — return `202 Accepted` immediately with a `Location` pointing back to `GET /api/oauth/aevatar-client/status`, and let the operator poll. Option (b) matches the existing status endpoint's read-side contract and would remove the polling loop entirely. The PR already returns 202 on the timeout path, so the upgrade path is small.

<details>
<summary>Per-model verbatim (mimo-v2.5-pro, glm-5.1)</summary>

- **mimo-v2.5-pro**: `WaitForRebuildObservedAsync` holds the HTTP connection open for up to 30s polling the readmodel. Many reverse proxies and load balancers have a 30s default timeout, so the happy path may get cut off. Consider returning 202 immediately with a poll URL, or tightening the production timeout to ≤15s to stay within typical proxy timeouts.
- **glm-5.1**: The 30-second polling loop holds the HTTP connection open. Many production load balancers (AWS ALB, nginx) default to 60s idle timeouts, so this is safe, but some stricter proxies (e.g. Cloudflare 100s, some corporate proxies at 10–15s) may terminate the connection before the 202 fires. Consider documenting the expected LB timeout budget in the PR or ops runbook, or switching to an async pattern (dispatch → return 202 immediately with a poll URL, matching the existing status endpoint).
</details>

agents/Aevatar.GAgents.Channel.Identity/Endpoints/IdentityOAuthEndpoints.cs
```diff
@@ -328,6 +341,241 @@ internal static async Task<IResult> HandleAevatarOAuthClientStatusAsync(
...
+
+        return Results.Ok(new
+        {
+            status = "rebuilt",
+            client_id = observed.ClientId,
+            client_id_issued_at = observed.ClientIdIssuedAt,
+            nyxid_authority = observed.NyxIdAuthority,
+            redirect_uri_registered = observed.RedirectUri,
+            oauth_scope_registered = observed.OauthScope,
+            broker_capability_observed = observed.BrokerCapabilityObserved,
```
eanzhao, line 557: **[glm-5.1]** severity=minor, category=security

The 200 response includes `broker_capability_observed` and `broker_capability_observed_at` from the snapshot. For an admin-only endpoint this is low risk, but it exposes internal runtime state that isn't necessary for the operator's workflow (confirming the new client_id is pinned). Consider trimming the response to only the fields the operator needs: `client_id`, `client_id_issued_at`, `nyxid_authority`, `redirect_uri_registered`, `oauth_scope_registered`.

test/Aevatar.GAgents.ChannelRuntime.Tests/Identity/IdentityOAuthClientRebuildEndpointTests.cs
```diff
@@ -0,0 +1,351 @@
...
+        cmd.RedirectUri.Should().Be("https://aevatar.test/api/oauth/nyxid-callback");
+        cmd.OauthScope.Should().Be(AevatarOAuthClientScopes.AuthorizationScope);
+        cmd.NyxidAuthority.Should().NotBeNullOrWhiteSpace();
+
+        var doc = await ReadJsonAsync(result);
+        doc.RootElement.GetProperty("status").GetString().Should().Be("rebuilt");
+        doc.RootElement.GetProperty("client_id").GetString().Should().Be(OperatorClientId);
+    }
+
+    [Fact]
```
eanzhao, line 156: **[mimo-v2.5-pro]** severity=minor, category=test

The `Returns202_WhenReadmodelDoesNotReflectRebuildBeforeTimeout` test doesn't verify that the provision command was actually dispatched to the actor. If someone broke the dispatch logic, this test would still pass (202 + stale provider). Add an assertion on `runtime.Captured.Should().HaveCount(1)` to confirm the actor received the command even when the readmodel times out.

PR #590, opened 2026-05-07
Lark streaming: phase machine refactor (#405) + CardKit streaming (#589)

### LINKED ISSUE #405: Lark streaming reply: explicit phase state machine + centralized unavailable guard
## Background

PR #374 (`feat/2026-04-24_lark-streaming-reply`) split the streaming-failure handling on `ConversationGAgent` into two flags so we no longer reuse a dead NyxID `/reply` token after the first chunk lands:

- `Disabled` — initial send failed before the reply token was consumed; safe to fall back to a single-shot `/reply` via `RunLlmReplyAsync`.
- `SuppressInterim` — first chunk consumed the token; only `/reply/update` is valid; no fallback to `/reply` allowed.

The fix is correct, but the runtime state is now an ad-hoc combination of two booleans plus a derived `ReplyTokenConsumed` (boolean over `PlatformMessageId`). This will keep accreting flags as more failure modes are added (rate-limit, message recall, edit-unsupported, etc.) and the legality of state transitions only lives in the heads of reviewers.

In parallel, "should I skip this callback because the message is already unavailable?" logic is scattered across `ConversationGAgent.HandleLlmReplyStreamChunkAsync`, `TryStreamedLlmReplyFinalizeAsync`, and the static-fallback path. New handlers (e.g. tool-call hooks, reasoning hooks if/when added) will have to remember to mirror the same checks.

The OpenClaw Lark plugin (https://github.com/ColinLu50/openclaw-lark-stream) hits the same shape of problem and resolves it with two patterns we should adopt:

1. An explicit `CardPhase` state machine with a `PHASE_TRANSITIONS` map that rejects illegal transitions and records a `terminalReason` for observability.
2. A single `UnavailableGuard` that owns the "should this callback short-circuit" decision, so every entrypoint defers to one method instead of repeating the same checks.

### A. Phase state machine for the per-turn streaming runtime state

Replace the `Disabled` + `SuppressInterim` + derived `ReplyTokenConsumed` shape on `NyxRelayStreamingState` with an explicit phase enum, e.g.

```
Idle               // no chunk attempted yet
PlaceholderSent    // first send landed, token consumed
Streaming          // interim edits flowing
SuppressingInterim // post-send interim edit failed; final edit still allowed
DisabledPreSend    // initial send failed before token consumed; /reply fallback allowed
TerminalSucceeded
TerminalPartial    // last flushed text was persisted as the user-visible terminal state
```

Constraints:
- Must remain in-memory on the actor (per CLAUDE.md "运行态不持久化"); the dictionary already lives outside `State`.
- Define a `PhaseTransitions` table and reject illegal transitions with a log line at warn level (do NOT throw — actor turns must keep making progress).
- Capture a `TerminalReason` on entering any terminal phase for diagnostics.
- All branches that today read `state.Disabled || state.SuppressInterim` should be expressed through phase-level helpers (`AllowsReplyFallback`, `AllowsInterimEdit`, `AllowsFinalEdit`).

### B. Centralized unavailable-message guard

Introduce a single guard helper on `ConversationGAgent` (or a small dependency) that owns:
- "Is the upstream message recalled / deleted / 230099-class error?"
- "Has this turn already been terminated for unavailability?"
- "Should this callback `source` short-circuit?"

Every public handler that touches the streaming path (`HandleLlmReplyStreamChunkAsync`, `TryStreamedLlmReplyFinalizeAsync`, future reasoning/tool hooks) should defer to this guard at the top of the method instead of repeating ad-hoc checks. New handlers added in the future then only have to call `if (ShouldSkipForUnavailable(\"<source>\")) return;`.

## Out of scope

- Switching the outbound path from `/reply` + `/reply/update` (edit-message) to Lark CardKit 2.0 streaming cards. That is a separate Lark-only UX track and will be filed separately if/when reasoning/tool visualization is on the roadmap.
- The `TurnStreamingReplySink` pending-flush timer / reflush-on-conflict work — handled in a separate PR (sink-only change, Sink doesn't touch the actor's failure state).

## References

- PR #374 commit `0d3a91a0` ("Fix streaming failure degradation and first-visible latency")
- OpenClaw plugin: https://github.com/ColinLu50/openclaw-lark-stream — `src/card/streaming-card-controller.ts` (PHASE_TRANSITIONS), `src/card/unavailable-guard.ts`
- CLAUDE.md sections: 单线程事实源, 中间层状态约束（运行态在 actor 内，不持久化）

eanzhao:
## Summary

Two issues collapsed into one PR per local discussion: the phase-machine refactor lays the foundation, the CardKit work builds on top.

### Part 1 — `#405` phase machine refactor (text-edit path)

- Replace `NyxRelayStreamingState`'s `Disabled` + `SuppressInterim` booleans with an explicit `NyxRelayStreamingPhase` enum (`Idle` / `PlaceholderSent` / `Streaming` / `SuppressingInterim` / `DisabledPreSend` / `TerminalSucceeded` / `TerminalPartial`). Transition validator logs warn on illegal transitions and keeps the actor turn making progress (no throw).
- Capture `TerminalReason` on entry to terminal phases for diagnostics (`failed_self_heal:{code}`, `final_edit_failed:{code}`, `empty_final_text`, `no_reply_token`, `completed`, etc.).
- Single `ShouldSkipNyxRelayStreamingForUnavailable` guard so every streaming entry point defers to one helper. Phase-level helpers (`AllowsInterimEdit` / `AllowsFinalEdit` / `AllowsReplyFallback`) replace direct field reads.
- New partial file `ConversationGAgent.NyxRelayStreaming.cs` so the streaming-state types live in one home.

### Part 2 — `#589` CardKit streaming (new card path, behind a flag)

End-to-end CardKit streaming now lives behind `NyxIdRelayOptions.StreamingCardKitEnabled` (default false). Default behaviour is unchanged; flipping the flag routes streaming chunks through CardKit with automatic fallback to the legacy edit-message path on create-time failure.

**Outbound client:**
- `LarkCardKitClient` (`src/Aevatar.AI.ToolProviders.Lark/`) wrapping the four CardKit REST endpoints we need: card create, element-content streaming (sequence-ordered), settings patch (close streaming mode), full card replace. Routed through the existing `NyxIdApiClient.ProxyRequestAsync` — no NyxID changes; bot scopes `cardkit:card:read` + `cardkit:card:write` confirmed granted in the Feishu console.

**Phase machine + state:**
- `LarkCardStreamingPhase` machine in `ConversationGAgent.LarkCardStreaming.cs`: `Idle / Creating / Streaming / Completed / Aborted / Terminated / CreationFailed`. Distinct from the text-edit `NyxRelayStreamingPhase` — card streaming has its own lifecycle (allocate, bind, stream, close-mode) and bypasses channel-relay. State carries `CardId`, `CardMessageId`, monotonic `Sequence`, configurable `StreamingElementId` (default `streaming_main`), `LastFlushedText`, `TerminalReason`.

**Runner contract:**
- `IConversationCardTurnRunner` with three operations matching CardKit's lifecycle (`RunCardCreateAsync` / `RunCardStreamAsync` / `RunCardFinalizeAsync`) plus typed result records that surface `IsRateLimited` / `IsTableLimitExceeded` / `IsCardUnavailable` so the grain branches fallback decisions without parsing error-code strings. `NullConversationCardTurnRunner` lets the actor naturally fall back to the text-edit sink when CardKit DI isn't wired.
- Production impl `ChannelCardConversationTurnRunner` (`Aevatar.GAgents.NyxidChat`) composes `LarkCardKitClient` (cardkit/v1/* endpoints) with `LarkNyxClient.SendMessageAsync` (im/v1/messages with msg_type=interactive) to drive create → send → stream → finalize. Resolves the bot owner's NyxID access token from `activity.TransportExtras.NyxUserAccessToken` and the receive target from `nyx_lark_chat_id` (groups) / `nyx_lark_union_id` (DMs).

**Sink + actor wiring:**
- Proto `LlmReplyStreamChunkEvent` gains a `card_mode` bool (default false). `TurnStreamingReplySink` accepts a `cardMode` constructor flag and stamps every dispatched chunk; the inbox runtime sets it when `StreamingCardKitEnabled=true` and switches the throttle to `StreamingCardKitFlushIntervalMs` (default 200ms) with no interim cap.
- `ConversationGAgent.HandleLlmReplyStreamChunkAsync` branches into `HandleLarkCardStreamingChunkCoreAsync` for card-mode chunks. That method owns the entire CardKit lifecycle: `Idle → Creating → Streaming` on first chunk, `Streaming` self-loop on subsequent chunks (sequence pre-incremented per call). Returns `false` only when phase is `CreationFailed` so the caller falls back to the legacy edit-message path; otherwise returns `true` and the card path owns the chunk.
- `TryCompleteStreamedReplyAsync` defers to `TryCompleteCardStreamedReplyAsync` first; that runs `RunCardFinalizeAsync` (optional trailing element-content write + PATCH `/settings` closing `streaming_mode`), persists `ConversationTurnCompletedEvent` with `SentActivityId="lark-card-stream:{card_message_id}"`, and clears the card state.

**Failure routing:**
- `card.create` rate-limit (230020) / table-limit (230099/11310) / scope errors → `CreationFailed` → caller falls through to the edit-message sink for the rest of the turn (still has reply_token + receive_id from the same activity).
- Mid-stream rate-limit drops the frame, sequence preserved.
- Mid-stream table-limit / unavailable → `Terminated` with last flushed text persisted as visible.
- Mid-stream fallback to edit-message is **out of scope for v1** (default impl matches issue #589 Scope D's "stop at full-turn fallback").

## Behaviour

- Part 1 is a pure refactor with no behaviour change.
- Part 2 is fully wired end-to-end behind the `StreamingCardKitEnabled` flag. Default flag value is `false`; the legacy text-edit path remains the only behaviour until the flag is enabled per-deployment.

## Test plan

- [x] `dotnet test test/Aevatar.GAgents.Channel.Protocol.Tests` — **132/132 pass** (125 existing + 7 new card-mode tests covering create, stream, sequence increment, rate-limit, table-limit, finalize prefix, and pre-flight fallback to edit-message).
- [x] `dotnet test test/Aevatar.GAgents.ChannelRuntime.Tests` — **896/896 pass**.
- [x] `tools/ci/channel_card_literal_guard.sh`, `tools/ci/channel_inbox_gagent_guard.sh` — green.
- [x] `git grep` confirms no remaining `Disabled` / `SuppressInterim` / `ReplyTokenConsumed` field reads in production code.
- [ ] Real-tenant smoke against the aevatar Lark bot with `StreamingCardKitEnabled=true` (post-merge).

Closes #405. Closes #589 — flag-gated rollout; mid-stream fallback (Scope D second tier) is out of scope per the v1 plan and can be added if real traffic shows we hit that path.

### DIFF
--- a/agents/Aevatar.GAgents.Channel.Runtime/Conversation/ConversationGAgent.LarkCardStreaming.cs
+++ b/agents/Aevatar.GAgents.Channel.Runtime/Conversation/ConversationGAgent.LarkCardStreaming.cs
@@ -1,0 +1,174 @@
+using Microsoft.Extensions.Logging;
+
+namespace Aevatar.GAgents.Channel.Runtime;
+
+public sealed partial class ConversationGAgent
+{
+    private readonly Dictionary<string, LarkCardStreamingState> _larkCardStreamingStates = new(StringComparer.Ordinal);
+
+    /// <summary>
+    /// Per-turn phase of the Lark CardKit streaming pipeline. Distinct from
+    /// <see cref="NyxRelayStreamingPhase"/> (which models channel-relay edit-message
+    /// streaming): card streaming has its own lifecycle (allocate card entity, bind to
+    /// chat, stream element content, close streaming mode) and goes through the API-key
+    /// proxy directly rather than channel-relay's <c>/reply{,/update}</c> surface.
+    /// </summary>
+    /// <remarks>
+    /// Fallback semantics: when card creation fails (<see cref="CreationFailed"/>), the
+    /// dispatcher routes the turn to the legacy text-edit sink (<c>NyxRelayStreamingPhase</c>
+    /// machine). Once <see cref="Streaming"/> is reached, the card path owns the turn —
+    /// mid-stream rate-limit / table-limit failures terminate the turn at
+    /// <see cref="Terminated"/> with the last flushed text persisted as partial.
+    /// </remarks>
+    private enum LarkCardStreamingPhase
+    {
+        Idle,
+        Creating,
+        Streaming,
+        Completed,
+        Aborted,
+        Terminated,
+        CreationFailed,
+    }
+
+    private enum LarkCardStreamingGuardSource
+    {
+        AcceptInterimChunk,
+        Finalize,
+    }
+
+    /// <summary>
+    /// Actor-scoped, in-memory streaming state for one CardKit-driven turn. Keyed by
+    /// <c>correlation_id</c>, same lifecycle as <see cref="NyxRelayReplyTokenContext"/>.
+    /// </summary>
+    /// <param name="Phase">Lifecycle phase; gates interim updates and finalization.</param>
+    /// <param name="CardId">
+    /// CardKit card entity id returned by <c>cardkit/v1/cards</c>. Null until
+    /// <see cref="LarkCardStreamingPhase.Streaming"/>; required for every element-content
+    /// and settings update afterwards.
+    /// </param>
+    /// <param name="CardMessageId">
+    /// Lark IM message id returned by the <c>im/v1/messages</c> send that bound the card
+    /// to a chat. Used by the unavailable-guard to detect upstream message recall.
+    /// </param>
+    /// <param name="OriginalCardId">
+    /// Preserved card id for terminal full-card update if mid-stream we fall back to text
+    /// patch (table-limit class errors). Currently always equal to <see cref="CardId"/>;
+    /// reserved for the mid-stream-fallback follow-up (#589 Scope D).
+    /// </param>
+    /// <param name="LastFlushedText">
+    /// Last text successfully streamed into the card element. Persisted as the user-visible
+    /// terminal state when finalization fails after streaming started.
+    /// </param>
+    /// <param name="Sequence">
+    /// Monotonic counter passed to every CardKit write. Pre-incremented before each call;
+    /// Lark rejects stale writes deterministically.
+    /// </param>
+    /// <param name="StreamingElementId">
+    /// Element id within the card to stream into. Defaults to <c>streaming_main</c>;
+    /// must match the card template's element naming.
+    /// </param>
+    /// <param name="TerminalReason">Diagnostic reason captured on entry to terminal phases.</param>
+    private sealed record LarkCardStreamingState(
+        LarkCardStreamingPhase Phase,
+        string? CardId,
+        string? CardMessageId,
+        string? OriginalCardId,
+        string LastFlushedText,
+        long Sequence,
+        string StreamingElementId,
+        string? TerminalReason)
+    {
+        public const string DefaultStreamingElementId = "streaming_main";
+
+        public static LarkCardStreamingState Initial { get; } = new(
+            LarkCardStreamingPhase.Idle,
+            CardId: null,
+            CardMessageId: null,
+            OriginalCardId: null,
+            LastFlushedText: string.Empty,
+            Sequence: 0,
+            StreamingElementId: DefaultStreamingElementId,
+            TerminalReason: null);
+
+        /// <summary>Phase permits accepting a new chunk (initial or interim).</summary>
+        public bool AllowsInterimEdit =>
+            Phase is LarkCardStreamingPhase.Idle
+                  or LarkCardStreamingPhase.Streaming;
+
+        /// <summary>
+        /// Card creation already failed — dispatcher should route subsequent chunks to the
+        /// text-edit sink for the rest of this turn.
+        /// </summary>
+        public bool AllowsTextEditFallback =>
+            Phase is LarkCardStreamingPhase.Idle
+                  or LarkCardStreamingPhase.CreationFailed;
+
+        /// <summary>Phase permits attempting a finalize (close streaming + optional final update).</summary>
+        public bool AllowsFinalize =>
+            Phase is LarkCardStreamingPhase.Streaming;
+    }
+
+    private static bool IsTerminalLarkCardStreamingPhase(LarkCardStreamingPhase phase) =>
+        phase is LarkCardStreamingPhase.Completed
+              or LarkCardStreamingPhase.Aborted
+              or LarkCardStreamingPhase.Terminated
+              or LarkCardStreamingPhase.CreationFailed;
+
+    private static bool IsLegalLarkCardStreamingTransition(LarkCardStreamingPhase from, LarkCardStreamingPhase to) =>
+        (from, to) switch
+        {
+            (LarkCardStreamingPhase.Idle, LarkCardStreamingPhase.Creating) => true,
+
+            (LarkCardStreamingPhase.Creating, LarkCardStreamingPhase.Streaming) => true,
+            (LarkCardStreamingPhase.Creating, LarkCardStreamingPhase.CreationFailed) => true,
+            (LarkCardStreamingPhase.Creating, LarkCardStreamingPhase.Terminated) => true,
+
+            (LarkCardStreamingPhase.Streaming, LarkCardStreamingPhase.Streaming) => true,
+            (LarkCardStreamingPhase.Streaming, LarkCardStreamingPhase.Completed) => true,
+            (LarkCardStreamingPhase.Streaming, LarkCardStreamingPhase.Aborted) => true,
+            (LarkCardStreamingPhase.Streaming, LarkCardStreamingPhase.Terminated) => true,
+
+            _ => false,
+        };
+
+    private LarkCardStreamingState GetOrInitLarkCardStreamingState(string correlationId) =>
+        _larkCardStreamingStates.GetValueOrDefault(correlationId) ?? LarkCardStreamingState.Initial;
+
+    private static bool ShouldSkipLarkCardStreamingForUnavailable(
+        LarkCardStreamingState state,
+        LarkCardStreamingGuardSource source) =>
+        source switch
+        {
+            LarkCardStreamingGuardSource.AcceptInterimChunk => !state.AllowsInterimEdit,
+            LarkCardStreamingGuardSource.Finalize => !state.AllowsFinalize,
+            _ => false,
+        };
+
+    private LarkCardStreamingState TransitionLarkCardStreamingPhase(
+        string correlationId,
+        LarkCardStreamingState current,
+        LarkCardStreamingPhase next,
+        string? terminalReason = null,
+        Func<LarkCardStreamingState, LarkCardStreamingState>? fieldUpdate = null)
+    {
+        if (!IsLegalLarkCardStreamingTransition(current.Phase, next))
+        {
+            Logger.LogWarning(
+                "Illegal Lark card streaming phase transition {From}->{To} for correlation={CorrelationId}; keeping current state",
+                current.Phase, next, correlationId);
+            return current;
+        }
+
+        var carried = fieldUpdate?.Invoke(current) ?? current;
+        var updated = carried with
+        {
+            Phase = next,
+            TerminalReason = IsTerminalLarkCardStreamingPhase(next)
+                ? (terminalReason ?? carried.TerminalReason)
+                : carried.TerminalReason,
+        };
+        _larkCardStreamingStates[correlationId] = updated;
+        return updated;
+    }
+}
--- a/agents/Aevatar.GAgents.Channel.Runtime/Conversation/ConversationGAgent.LarkCardStreaming.cs
+++ b/agents/Aevatar.GAgents.Channel.Runtime/Conversation/ConversationGAgent.LarkCardStreaming.cs
@@ -185,7 +185,7 @@ public sealed partial class ConversationGAgent
     /// chunks through edit-message streaming."
     /// </summary>
     private async Task<bool> HandleLarkCardStreamingChunkCoreAsync(
-        LlmReplyStreamChunkEvent evt,
+        LlmReplyCardStreamChunkEvent evt,
         string correlationId)
     {
         var state = GetOrInitLarkCardStreamingState(correlationId);
--- a/agents/Aevatar.GAgents.Channel.Runtime/Conversation/ConversationGAgent.LarkCardStreaming.cs
+++ b/agents/Aevatar.GAgents.Channel.Runtime/Conversation/ConversationGAgent.LarkCardStreaming.cs
@@ -1,3 +1,5 @@
+using Aevatar.GAgents.Channel.Abstractions;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 
 namespace Aevatar.GAgents.Channel.Runtime;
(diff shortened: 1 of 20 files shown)


agents/Aevatar.GAgents.Channel.Runtime/Conversation/ConversationGAgent.LarkCardStreaming.cs
```diff
@@ -0,0 +1,429 @@
...
+    /// the card path from the legacy <c>nyx-relay-stream:</c> path.
+    /// </summary>
+    private async Task<bool> TryCompleteCardStreamedReplyAsync(
+        LlmReplyReadyEvent evt,
+        string correlationId,
+        string commandId,
+        ChatActivity? referenceActivity)
+    {
+        var state = GetOrInitLarkCardStreamingState(correlationId);
+        if (state.Phase is not LarkCardStreamingPhase.Streaming)
```
eanzhao: **[codex]** severity=major, category=bug

Mid-stream table-limit/unavailable transitions set the state to `Terminated`, but the ready path only handles `Streaming` and returns `false` for `Terminated`. That means `PersistCardStreamedCompletionAsync()` and token/state cleanup never run for the documented partial-card terminal path, and the caller can fall through to the legacy reply flow even though a card is already visible. Handle `Terminated`/`Aborted` here by persisting `LastFlushedText` with the card prefix and removing the relay token, or persist/cleanup when entering the terminal phase.

agents/Aevatar.GAgents.Channel.Runtime/Conversation/ConversationGAgent.cs
```diff
@@ -575,23 +540,38 @@ public async Task HandleLlmReplyStreamChunkAsync(LlmReplyStreamChunkEvent evt)
...
         }
 
+        // CardKit-mode chunks go through the card path. The card handler returns false ONLY
+        // when phase is CreationFailed (card create already failed pre-flight or on first
+        // chunk) — in that case the chunk falls through to the legacy text-edit path so the
+        // user still sees a reply. All other phases (Idle/Streaming/terminal) are handled
+        // end-to-end by the card handler.
+        if (evt.CardMode)
+        {
+            if (await HandleLarkCardStreamingChunkCoreAsync(evt, correlationId).ConfigureAwait(false))
```
eanzhao: **[kimi]** severity=major, category=concurrency

`ConfigureAwait(false)` should not be used in actor code. The existing `ConversationGAgent.cs` does not use it anywhere else; actor turns rely on a single-threaded execution context and these two new await points (also line 650) are the only occurrences in the file. If the continuation runs on a thread-pool thread, subsequent state mutations to `_nyxRelayStreamingStates` / `_nyxRelayReplyTokens` happen outside the actor scheduler. Remove `.ConfigureAwait(false)` from both new await expressions to match the existing actor pattern.

agents/Aevatar.GAgents.Channel.Runtime/Conversation/ConversationGAgent.cs
```diff
@@ -652,12 +644,14 @@ private async Task<bool> TryCompleteStreamedReplyAsync(
...
-        if (!_nyxRelayStreamingStates.TryGetValue(correlationId, out var state))
-            return false;
-        // Disabled means the initial send never landed, so the reply token is still usable
-        // and the caller may fall back to a single-shot /reply. A missing PlatformMessageId
-        // with SuppressInterim would be inconsistent, but treat it the same for safety.
-        if (state.Disabled || string.IsNullOrEmpty(state.PlatformMessageId))
+        // Card path takes precedence when active; falls through to text-edit when card never
+        // started (Idle), card creation failed (CreationFailed → text-edit fallback), or card
+        // finished as a terminal phase.
+        if (await TryCompleteCardStreamedReplyAsync(evt, correlationId, commandId, referenceActivity).ConfigureAwait(false))
```
eanzhao: **[kimi]** severity=major, category=concurrency

Same as line 556: remove `.ConfigureAwait(false)` to preserve the actor scheduling context. After this await returns `false`, the method continues to read/write `_nyxRelayStreamingStates`, which must stay on the actor's single-threaded turn.

agents/Aevatar.GAgents.Channel.Runtime/protos/conversation_events.proto
```diff
@@ -95,6 +95,11 @@ message LlmReplyStreamChunkEvent {
   // Current accumulated reply text (not a delta slice). Each chunk supersedes the previous one.
   string accumulated_text = 4;
   int64 chunk_at_unix_ms = 5;
+  // True when dispatched by TurnStreamingCardSink (CardKit path) instead of TurnStreamingReplySink.
+  // The actor branches into HandleLlmReplyCardStreamChunk semantics — drives LarkCardStreamingState,
+  // calls IConversationCardTurnRunner — instead of the legacy edit-message path. Default false
```
eanzhao: **[mimo-v2.5-pro]** severity=major, category=design

`card_mode` is a runtime-only routing flag (the comment says "must never be persisted"), but it's added to `LlmReplyStreamChunkEvent` — a proto message in `conversation_events.proto` that is part of the domain event surface. If a caller accidentally persists this event, the `card_mode` flag survives serialization and would re-trigger card routing on replay.

Consider either: (a) moving the card-mode chunk to a separate non-persisted proto (e.g. `LlmReplyCardStreamChunkEvent`), or (b) adding a guard in the event-store write path to strip `card_mode` before persistence. Option (a) is cleaner and eliminates the risk entirely.

agents/Aevatar.GAgents.NyxidChat/ChannelCardConversationTurnRunner.cs
```diff
@@ -0,0 +1,347 @@
...
+                    IdempotencyKey: $"{chunk.CorrelationId}-1"),
+                ct);
+        }
+        catch (Exception ex)
+        {
+            _logger.LogWarning(ex, "CardKit first stream threw for correlation={CorrelationId}, card_id={CardId}", chunk.CorrelationId, cardId);
+            return ConversationCardCreateResult.Failed("card_first_stream_threw", ex.Message);
+        }
+
+        if (LarkProxyResponseParser.TryParseError(firstStreamResponse, out var firstStreamError))
```
eanzhao, line 139: **[codex]** severity=major, category=bug

This treats the first element-content failure as a pre-send create failure, but the interactive card message has already been sent at lines 83-94. The actor will mark `CreationFailed` and fall back to the legacy text path, leaving an empty/streaming CardKit message visible alongside the fallback text reply. Move the initial text into the card before binding/sending, stream before send if the API allows it, or return a distinct post-send failure result with card/message ids so the actor can terminate/update the card instead of falling back to text.

agents/Aevatar.GAgents.NyxidChat/ServiceCollectionExtensions.cs
```diff
@@ -41,6 +41,7 @@ public static IServiceCollection AddNyxIdChat(this IServiceCollection services,
 
         // ─── Conversation turn-runner override + reply generator ───
         services.Replace(ServiceDescriptor.Singleton<IConversationTurnRunner, ChannelConversationTurnRunner>());
+        services.Replace(ServiceDescriptor.Singleton<IConversationCardTurnRunner, ChannelCardConversationTurnRunner>());
```
eanzhao: **[codex]** severity=major, category=di

`AddNyxIdChat()` now unconditionally registers `ChannelCardConversationTurnRunner`, but this module does not register its required `ILarkCardKitClient` / `ILarkNyxClient` dependencies; hosts that call only `AddNyxIdChat()` will fail provider validation or fail when card mode resolves the runner. Mainnet currently also calls `AddLarkTools()`, but the package registration itself is no longer self-contained. Register the Lark client dependencies here, or make the card runner registration conditional/factory-based so it returns `NullConversationCardTurnRunner` when Lark tools are absent.

agents/Aevatar.GAgents.Channel.Runtime/Conversation/ConversationGAgent.LarkCardStreaming.cs
```diff
@@ -0,0 +1,429 @@
...
+            if (streamResult.IsTableLimitExceeded || streamResult.IsCardUnavailable)
+            {
+                Logger.LogWarning(
+                    "Card stream terminal failure; ending turn. correlation={CorrelationId}, code={ErrorCode}",
+                    evt.CorrelationId, streamResult.ErrorCode);
+                TransitionLarkCardStreamingPhase(
+                    correlationId,
+                    state,
+                    LarkCardStreamingPhase.Terminated,
+                    terminalReason: $"stream_failed:{streamResult.ErrorCode}");
```
eanzhao, line 332: **[kimi]** severity=minor, category=design

When a mid-stream failure transitions to `Terminated` (e.g. table-limit / unavailable at line 299), `TryCompleteCardStreamedReplyAsync` later returns `false` at line 333 because phase is no longer `Streaming`. The text-edit fallback path also returns `false` (no text-edit state exists). Consequently no `ConversationTurnCompletedEvent` is persisted for the turn, so the event store has no terminal record despite the user seeing partial text in the card. Consider persisting a terminal `ConversationTurnCompletedEvent` (with the last-flushed text) when entering `Terminated`, or explicitly document this v1 gap.

agents/Aevatar.GAgents.Channel.Runtime/Conversation/ConversationGAgent.LarkCardStreaming.cs
```diff
@@ -0,0 +1,429 @@
...
+            TransitionLarkCardStreamingPhase(
+                correlationId,
+                state,
+                LarkCardStreamingPhase.Terminated,
+                terminalReason: $"finalize_threw:{ex.GetType().Name}");
+            await PersistCardStreamedCompletionAsync(evt, commandId, referenceActivity, state.CardMessageId ?? string.Empty, state.LastFlushedText);
+            return true;
+        }
+
+        var visibleText = finalDiffers && finalizeResult.Success ? finalText : state.LastFlushedText;
```
eanzhao: **[codex]** severity=minor, category=bug

`visibleText` assumes any finalize failure means the final text was not visible, but `RunCardFinalizeAsync()` can successfully write the final element content and then fail while closing `streaming_mode`. In that case the card shows `finalText` while the persisted `ConversationTurnCompletedEvent` records `LastFlushedText`. Split the final-write and close-streaming outcomes, or have `ConversationCardFinalizeResult` carry whether the final text landed so persistence matches the user-visible card.

agents/Aevatar.GAgents.NyxidChat/ChannelCardConversationTurnRunner.cs
```diff
@@ -0,0 +1,347 @@
...
+                    CardId: cardId,
+                    ElementId: streamingElementId,
+                    Content: chunk.AccumulatedText,
+                    Sequence: 1,
+                    IdempotencyKey: $"{chunk.CorrelationId}-1"),
+                ct);
+        }
+        catch (Exception ex)
+        {
+            _logger.LogWarning(ex, "CardKit first stream threw for correlation={CorrelationId}, card_id={CardId}", chunk.CorrelationId, cardId);
```
eanzhao, line 130: **[glm-5.1]** severity=minor, category=design

If `StreamElementContentAsync` (sequence 1) fails after `CreateCardAsync` + `SendMessageAsync` have already succeeded, an empty streaming card is visible in the chat with no cleanup. The method returns a failure result → grain enters `CreationFailed` → text-edit fallback sends a separate message. The user sees both an empty card and a text reply. Consider best-effort closing the orphaned card (e.g. `SetCardSettingsAsync(streaming_mode: false)`) before returning the failure, or at minimum logging the leaked `card_id` for ops visibility.

agents/Aevatar.GAgents.NyxidChat/ChannelCardConversationTurnRunner.cs
```diff
@@ -0,0 +1,347 @@
...
+        // union_id whenever it can resolve it, so this branch generally does not fire.
+        if (!string.IsNullOrWhiteSpace(chatId))
+            return ("chat_id", chatId);
+
+        return null;
+    }
+
+    /// <summary>
+    /// Minimal Lark CardKit 2.0 card with a single markdown element identified by
+    /// <paramref name="streamingElementId"/>. Streaming text is written via
```
eanzhao: **[mimo-v2.5-pro]** severity=minor, category=bug

Error code `230099` is classified inconsistently between the two failure classifiers:

- `ClassifyCreateFailure` (line ~275): sets **both** `isTableLimitExceeded` and `isCardUnavailable` for 230099.
- `ClassifyStreamFailure` (line ~284): sets **only** `isCardUnavailable` for 230099 (no `isTableLimitExceeded`).

In the actor, `IsTableLimitExceeded` triggers `Terminated` (terminal — turn ends), while `IsCardUnavailable` alone also triggers `Terminated`. The behavioral difference is subtle today, but the inconsistency is a maintenance hazard. If 230099 really means "table limit" in both contexts, add it to `ClassifyStreamFailure`; if it means "unavailable" in both, remove it from `ClassifyCreateFailure`'s `isTableLimitExceeded` path.

agents/Aevatar.GAgents.NyxidChat/ChannelCardConversationTurnRunner.cs
```diff
@@ -0,0 +1,347 @@
...
+    /// Returns null when the field is missing or malformed; the caller treats null as a
+    /// terminal create failure.
+    /// </summary>
+    private static string? ExtractCardId(string response)
+    {
+        try
+        {
+            using var document = JsonDocument.Parse(response);
+            if (document.RootElement.TryGetProperty("data", out var data) &&
+                data.TryGetProperty("card_id", out var cardIdProp) &&
```
eanzhao, line 325: **[glm-5.1]** severity=minor, category=bug

`ContainsLarkCode` uses raw substring matching (`error.Contains($"lark_code={code}")`) without a trailing delimiter. If Lark ever returns a code like `2300201`, searching for `lark_code=230020` would falsely match. Consider appending a non-digit boundary check (e.g. match `lark_code={code}` followed by a non-digit or end-of-string), or have `LarkProxyResponseParser.TryParseError` return a structured error code integer so the runner compares directly.

agents/Aevatar.GAgents.NyxidChat/ChannelCardConversationTurnRunner.cs
```diff
@@ -0,0 +1,347 @@
...
+        ConversationCardStreamResult.Failed(
+            errorCode: "card_stream_failed",
+            errorSummary: larkError,
+            isRateLimited: ContainsLarkCode(larkError, 230020),
+            isTableLimitExceeded: ContainsLarkCode(larkError, 230099) || ContainsLarkCode(larkError, 11310),
+            isCardUnavailable: ContainsLarkCode(larkError, 230100));
+
+    /// <summary>
+    /// Substring match against <see cref="LarkProxyResponseParser.TryParseError"/>'s output
+    /// shape (<c>"lark_code={n} ..."</c>). Cheap, allocation-free; the parser owns the
```
eanzhao: **[v4-pro]** severity=minor, category=bug

`ContainsLarkCode` uses `error.Contains($"lark_code={code}", ...)` substring matching. For numeric codes that are prefixes of other Lark error codes (e.g. checking for `23002` in a string containing `lark_code=230020`), this produces a false positive. Fix: use boundary-aware matching, e.g. `error.Contains($"lark_code={code} ") || error.EndsWith($"lark_code={code}")` or a regex like `lark_code={code}\b`.

PR #597, opened 2026-05-08
[codex] Refactor LLM reply continuation into agent runs

eanzhao:
## Summary
- Implements issue #596 Phase A by replacing the channel LLM reply hosted inbox singleton with run-scoped `AgentRunGAgent` continuation actors.
- Adds `IChannelLlmReplyRunDispatcher` and the NyxidChat dispatcher that creates/gets `AgentRunGAgent` by `correlationId` and dispatches typed `AgentRunStartRequested` commands.
- Moves the existing LLM reply gates, metadata enrichment, fallback timeout, streaming sink, reply generation, ready/drop dispatch, and relay token echo into the run actor while keeping reply tokens out of persisted run state/events.
- Updates `ConversationGAgent` and tests to use the run dispatcher path and retires the old inbox interface/hosted service path.

## Validation
- `dotnet build test/Aevatar.GAgents.ChannelRuntime.Tests/Aevatar.GAgents.ChannelRuntime.Tests.csproj --nologo`
- `dotnet test test/Aevatar.GAgents.Channel.Protocol.Tests/Aevatar.GAgents.Channel.Protocol.Tests.csproj --nologo`

Refs #596

### DIFF
--- a/agents/Aevatar.GAgents.NyxidChat/AgentRunDispatcher.cs
+++ b/agents/Aevatar.GAgents.NyxidChat/AgentRunDispatcher.cs
@@ -1,0 +1,65 @@
+using Aevatar.Foundation.Abstractions;
+using Aevatar.GAgents.Channel.Runtime;
+using Google.Protobuf.WellKnownTypes;
+using Microsoft.Extensions.Logging;
+
+namespace Aevatar.GAgents.NyxidChat;
+
+/// <summary>
+/// Thin Channel.Runtime port implementation that creates the run actor and
+/// dispatches the start command. It holds no run state.
+/// </summary>
+public sealed class AgentRunDispatcher : IChannelLlmReplyRunDispatcher
+{
+    private readonly IActorRuntime _actorRuntime;
+    private readonly IActorDispatchPort _actorDispatchPort;
+    private readonly TimeProvider _timeProvider;
+    private readonly ILogger<AgentRunDispatcher> _logger;
+
+    public AgentRunDispatcher(
+        IActorRuntime actorRuntime,
+        IActorDispatchPort actorDispatchPort,
+        ILogger<AgentRunDispatcher> logger,
+        TimeProvider? timeProvider = null)
+    {
+        _actorRuntime = actorRuntime ?? throw new ArgumentNullException(nameof(actorRuntime));
+        _actorDispatchPort = actorDispatchPort ?? throw new ArgumentNullException(nameof(actorDispatchPort));
+        _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        _timeProvider = timeProvider ?? TimeProvider.System;
+    }
+
+    public async Task DispatchAsync(NeedsLlmReplyEvent request, CancellationToken ct)
+    {
+        ArgumentNullException.ThrowIfNull(request);
+        if (string.IsNullOrWhiteSpace(request.CorrelationId))
+            throw new InvalidOperationException("Deferred LLM reply request requires correlation_id for AgentRunGAgent dispatch.");
+
+        var runId = request.CorrelationId.Trim();
+        var actorId = AgentRunGAgent.BuildActorId(runId);
+        var actor = await _actorRuntime.GetAsync(actorId)
+                    ?? await _actorRuntime.CreateAsync<AgentRunGAgent>(actorId, ct);
+
+        var command = new AgentRunStartRequested
+        {
+            Request = request.Clone(),
+        };
+        var envelope = new EventEnvelope
+        {
+            Id = Guid.NewGuid().ToString("N"),
+            Timestamp = Timestamp.FromDateTimeOffset(_timeProvider.GetUtcNow()),
+            Payload = Any.Pack(command),
+            Route = EnvelopeRouteSemantics.CreateDirect("channel-llm-reply-run-dispatcher", actor.Id),
+            Propagation = new EnvelopePropagation
+            {
+                CorrelationId = runId,
+            },
+        };
+
+        await _actorDispatchPort.DispatchAsync(actor.Id, envelope, ct);
+        _logger.LogInformation(
+            "Dispatched deferred LLM reply run: runId={RunId} actorId={ActorId} target={TargetActorId}",
+            runId,
+            actor.Id,
+            request.TargetActorId);
+    }
+}
--- a/agents/Aevatar.GAgents.NyxidChat/AgentRunDispatcher.cs
+++ b/agents/Aevatar.GAgents.NyxidChat/AgentRunDispatcher.cs
@@ -12,18 +12,18 @@ namespace Aevatar.GAgents.NyxidChat;
 public sealed class AgentRunDispatcher : IChannelLlmReplyRunDispatcher
 {
     private readonly IActorRuntime _actorRuntime;
-    private readonly IActorDispatchPort _actorDispatchPort;
+    private readonly IStreamProvider _streamProvider;
     private readonly TimeProvider _timeProvider;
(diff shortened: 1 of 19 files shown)


agents/Aevatar.GAgents.NyxidChat/AgentRunDispatcher.cs
```diff
@@ -0,0 +1,65 @@
...
+            Timestamp = Timestamp.FromDateTimeOffset(_timeProvider.GetUtcNow()),
+            Payload = Any.Pack(command),
+            Route = EnvelopeRouteSemantics.CreateDirect("channel-llm-reply-run-dispatcher", actor.Id),
+            Propagation = new EnvelopePropagation
+            {
+                CorrelationId = runId,
+            },
+        };
+
+        await _actorDispatchPort.DispatchAsync(actor.Id, envelope, ct);
```
eanzhao: **[codex]** severity=blocker, category=concurrency

This uses synchronous `IActorDispatchPort.DispatchAsync` to start the run actor. `ConversationGAgent` awaits this while handling the inbound turn, and `AgentRunGAgent` later dispatches `LlmReplyReadyEvent` back to that same conversation actor, so the real Local/Orleans dispatch implementations can form an A -> B -> A wait cycle and also block the webhook for the full LLM run. This violates CLAUDE.md: cross-actor waits must be continuationized. Start the run through a one-way enqueue/publish mechanism that returns after acceptance, then let the run actor deliver the ready/drop signal independently.

agents/Aevatar.GAgents.NyxidChat/AgentRunGAgent.cs
```diff
@@ -57,109 +66,106 @@ public ChannelLlmReplyInboxRuntime(
...
+        return ActorIdPrefix + correlationId.Trim();
     }
 
-    public async Task StopAsync(CancellationToken ct)
+    protected override AgentRunGAgentState TransitionState(AgentRunGAgentState current, IMessage evt) =>
+        StateTransitionMatcher
+            .Match(current, evt)
+            .On<AgentRunStartedEvent>(ApplyStarted)
+            .On<AgentRunReplyProducedEvent>(ApplyReplyProduced)
+            .On<AgentRunDroppedEvent>(ApplyDropped)
```
eanzhao, line 86: **[glm-5.1]** severity=major, category=arch

`AgentRunGAgent` reaches terminal states (`ReplyProduced`, `Dropped`, `Failed`) but has no self-deactivation or cleanup mechanism. Per AGENTS.md: *"默认优先 run/session/task-scoped actor"* and *"长期 actor 只保留给事实拥有者"*. Each LLM reply creates a persistent actor that accumulates indefinitely. Consider persisting the terminal event then scheduling a self-deactivation (e.g. `IActorRuntime.DeactivateAsync` or a delayed cleanup self-message) so the actor does not remain alive after its single-shot purpose is served.

agents/Aevatar.GAgents.NyxidChat/AgentRunGAgent.cs
```diff
@@ -57,109 +66,106 @@ public ChannelLlmReplyInboxRuntime(
...
+        var runId = NormalizeOptional(request.CorrelationId) ?? Id;
+        var startedAtUnixMs = _timeProvider.GetUtcNow().ToUnixTimeMilliseconds();
 
-    public async ValueTask DisposeAsync()
-    {
-        await StopAsync(CancellationToken.None);
-    }
-
-    internal const long MaxInboxRequestAgeMs = 5 * 60 * 1000;
+        if (State.Status is AgentRunStatus.ReplyProduced or AgentRunStatus.Dropped or AgentRunStatus.Failed)
```
eanzhao, line 104: **[Consensus: 2 models]** severity=major, category=bug

The terminal-state dedup guard at this line silently drops undelivered ready/drop notifications when conversation dispatch fails. The actor persists `AgentRunReplyProducedEvent` / `AgentRunFailedEvent` / `AgentRunDroppedEvent` *before* dispatching to the conversation actor; if that dispatch throws, a retry of the start command will hit this guard and return without re-emitting the signal — `ConversationGAgent.PendingLlmReplyRequests` stays stuck. The guard also has zero test coverage. Either move the guard to *after* a confirmed-delivered phase (dispatch retry as a persisted phase), or only mark the run terminal once the conversation acknowledges. Add a test covering both the legitimate-duplicate case and the failed-dispatch retry case.

<details>

- **kimi**: `HandleStartAsync` introduces a new terminal-state dedup guard (ignoring duplicate start commands when the run is already `ReplyProduced`, `Dropped`, or `Failed`). This idempotency behavior is not covered by any test in `AgentRunGAgentTests.cs`. Add a test that sends a second `AgentRunStartRequested` after the run reaches a terminal state and asserts it is ignored.
- **codex**: This terminal duplicate guard can drop undelivered results. The run persists `AgentRunReplyProducedEvent` / `AgentRunFailedEvent` before `DispatchReadyEventAsync`, and persists `AgentRunDroppedEvent` before the drop notification; if that conversation dispatch fails, a retry of the start command sees terminal state here and returns without re-emitting the ready/drop signal, leaving `ConversationGAgent.PendingLlmReplyRequests` stuck. Make delivery a retryable persisted phase, or only mark the run terminal after the conversation has accepted/acknowledged the notification.
</details>

agents/Aevatar.GAgents.NyxidChat/AgentRunGAgent.cs
```diff
@@ -57,109 +66,106 @@ public ChannelLlmReplyInboxRuntime(
...
-        if (_subscription is not null)
-            return;
-
-        _subscription = await _streamProvider
-            .GetStream(InboxStreamId)
-            .SubscribeAsync<NeedsLlmReplyEvent>(ProcessAsync, ct);
-
-        _logger.LogInformation("Started channel LLM reply inbox on {StreamId}", InboxStreamId);
+        ArgumentException.ThrowIfNullOrWhiteSpace(correlationId);
+        return ActorIdPrefix + correlationId.Trim();
```
eanzhao, line 78: **[glm-5.1]** severity=minor, category=arch

`HandleStartAsync` calls `PersistDomainEventAsync(AgentRunStartedEvent)` then synchronously calls `ProcessAsync` which runs the full LLM generation (up to 300s). If `PersistDomainEventAsync` for the started event succeeds but `ProcessAsync` throws an unguarded exception (e.g. from `EnsureTargetActorAsync` or a subsequent `PersistDomainEventAsync` call), the actor is left in `Started` state with no terminal event and no retry path. Wrap the `ProcessAsync` call in a try/catch that persists `AgentRunFailedEvent` on any unexpected exception to guarantee terminal state is always reached.

test/Aevatar.GAgents.ChannelRuntime.Tests/AgentRunGAgentTests.cs
```diff
@@ -712,6 +755,8 @@ private sealed class DispatchingActorRuntime(params (string Id, IActor Actor)[]
             static pair => pair.Actor,
```
eanzhao, line 1028: **[glm-5.1]** severity=minor, category=test

`NoOpEventSourcing.TransitionState` returns `current` unchanged, so `AgentRunGAgent.State` never advances past `AGENT_RUN_STATUS_UNSPECIFIED` in any test. This means the duplicate-terminal-start guard at `HandleStartAsync` line ~78 (`State.Status is AgentRunStatus.ReplyProduced or AgentRunStatus.Dropped or AgentRunStatus.Failed`) is never exercised. Either wire a real `IEventSourcingBehavior` that calls `TransitionState`, or add a dedicated test that pre-populates the state to a terminal status and asserts the guard fires.

PR #600, opened 2026-05-08
Fix #599: per-conversation LLM reply executor (drop silo-wide inbox)

eanzhao:
## Summary

Moves LLM reply work from a silo-wide hosted-service stream subscriber into a per-conversation executor seam owned by `ConversationGAgent`. Different conversations (different Lark users) now run their LLM replies in parallel instead of strict FIFO serialization across the whole silo.

Closes #599.

## Root cause

`ChannelLlmReplyInboxRuntime` registered a single Orleans persistent-stream `SubscribeAsync` per silo (1 pod = 1 silo = 1 callback). Orleans serializes `SubscribeAsync` callbacks FIFO, and `ProcessAsync` synchronously awaited 60-300s LLM calls — so every Lark user's reply queued behind every other user's reply.

Production logs (60 min window 2026-05-08 08:28-09:28, pod `aevatar-console-backend-7b69b89cd6-x9d9t`): 6 LLM reply requests, 4 timeouts, 67% failure rate. Adjacent users' `Processing` timestamps showed strict back-to-back FIFO across distinct user IDs.

## Approach (Solution C)

- New `IConversationLlmReplyExecutor` seam in `Aevatar.GAgents.Channel.Runtime`. Single method: `Task StartAsync(NeedsLlmReplyEvent, CancellationToken)`.
- New `ConversationLlmReplyExecutor` implementation in `Aevatar.GAgents.NyxidChat`:
  - Pre-LLM gates (stale age, missing relay credential, malformed payload), bot-owner config enrichment, the LLM call itself, streaming-sink wiring, and dispatch of the terminal signal back to the originating actor — same logic the old `ProcessAsync` had.
  - `StartAsync` is non-blocking: it spawns a `Task.Run` for the LLM work and returns immediately. The actor turn does not wait on the 60-300s call.
  - The background task only does external I/O and finishes by dispatching `LlmReplyReadyEvent` (or `DeferredLlmReplyDroppedEvent`) to `request.TargetActorId` via `IActorDispatchPort`. It never reads or writes actor state — only signals — preserving the actor's single-threaded execution invariant per CLAUDE.md.
- `ConversationGAgent.DispatchPendingLlmReplyAsync` now calls `executor.StartAsync` instead of `inbox.EnqueueAsync`.
- Deleted `ChannelLlmReplyInboxRuntime`, `ChannelLlmReplyInboxHostedService`, and `IChannelLlmReplyInbox`.
- DI registration drops the silo-wide hosted service; replaced with a singleton `ConversationLlmReplyExecutor`.

Durability: the request is persisted in `State.PendingLlmReplyRequests` before the actor calls `StartAsync`, exactly as before. Pod restart mid-flight rehydrates the actor and re-kicks the executor on activation via the existing `SchedulePendingLlmReplyDispatchesAsync` path.

CLAUDE.md fit: "Actor 即业务实体 — 一个 actor = 一个业务实体（数据与方法同住）；禁止按技术功能（读/写/投影）拆分同一业务实体为多个 actor". The LLM reply is part of the conversation entity's behavior; the original host-level worker was the violation.

## Files

- `agents/Aevatar.GAgents.Channel.Runtime/Conversation/IConversationLlmReplyExecutor.cs` (new)
- `agents/Aevatar.GAgents.NyxidChat/ConversationLlmReplyExecutor.cs` (renamed from `ChannelLlmReplyInboxRuntime.cs`)
- `agents/Aevatar.GAgents.Channel.Runtime/Conversation/IChannelLlmReplyInbox.cs` (deleted)
- `agents/Aevatar.GAgents.Channel.Runtime/Conversation/ConversationGAgent.cs` — `DispatchPendingLlmReplyAsync` rewired
- `agents/Aevatar.GAgents.NyxidChat/ServiceCollectionExtensions.cs` — drop hosted service, register executor singleton
- `test/Aevatar.GAgents.ChannelRuntime.Tests/ConversationLlmReplyExecutorTests.cs` (renamed from `ChannelLlmReplyInboxRuntimeTests.cs`, ported to executor-shape)
- `test/Aevatar.GAgents.Channel.Protocol.Tests/ConversationGAgentDedupTests.cs` — `RecordingInbox` → `RecordingExecutor`
- Misc comment / proto-doc / allowlist renames

## Test plan

- [x] `dotnet test test/Aevatar.GAgents.ChannelRuntime.Tests/Aevatar.GAgents.ChannelRuntime.Tests.csproj` — 895/895 pass
- [x] `dotnet test test/Aevatar.GAgents.Channel.Protocol.Tests/Aevatar.GAgents.Channel.Protocol.Tests.csproj` — 133/133 pass
- [x] `dotnet test test/Aevatar.GAgents.Platform.Lark.Tests/Aevatar.GAgents.Platform.Lark.Tests.csproj` — 17/17 pass
- [x] `bash tools/ci/architecture_guards.sh` — pass
- [ ] **Production verification** (post-deploy): script 10 simultaneous Lark DMs to different users; expect their `Processing` timestamps to interleave (no longer strict FIFO) and failure rate < 5% in the same 60-min window.

## Out of scope

- `SkillRunnerGAgent` non-reentrant grain (mentioned in #599 as a secondary bottleneck) — separate fix.
- Cold-start `EventStoreOptimisticConcurrencyException` against `user-agent-catalog-read-model` — independent issue, low frequency (1/hr).

### DIFF
--- a/agents/Aevatar.GAgents.NyxidChat/ConversationLlmReplyExecutor.cs
+++ b/agents/Aevatar.GAgents.NyxidChat/ConversationLlmReplyExecutor.cs
@@ -1,26 +1,39 @@
 using Aevatar.Foundation.Abstractions;
-using Aevatar.Foundation.Abstractions.Streaming;
 using Aevatar.AI.Abstractions.LLMProviders;
 using Aevatar.GAgents.Channel.Abstractions;
 using Aevatar.GAgents.Channel.Runtime;
 using Aevatar.GAgents.Channel.NyxIdRelay;
-using Aevatar.GAgents.NyxidChat;
 using Aevatar.Studio.Application.Studio.Abstractions;
 using Google.Protobuf.WellKnownTypes;
-using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
 
 namespace Aevatar.GAgents.NyxidChat;
 
-public sealed class ChannelLlmReplyInboxRuntime :
-    IHostedService,
-    IAsyncDisposable,
-    IChannelLlmReplyInbox
+/// <summary>
+/// Drives one LLM reply turn end-to-end on behalf of <see cref="ConversationGAgent"/>:
+/// pre-LLM gates (stale age, missing relay token, malformed payload), bot-owner config
+/// enrichment, the LLM call itself, streaming-sink wiring, and dispatch of the terminal
+/// signal back to the originating actor.
+/// </summary>
+/// <remarks>
+/// <para>
+/// Replaces the silo-wide <c>ChannelLlmReplyInboxRuntime</c> stream subscriber. The actor
+/// invokes <see cref="StartAsync"/> from inside its turn; the executor schedules the LLM
+/// work on a background task so the 60-300s call cannot pin the actor turn, then
+/// dispatches an <c>LlmReplyReadyEvent</c> (or <c>DeferredLlmReplyDroppedEvent</c>) back
+/// to <c>request.TargetActorId</c> via <see cref="IActorDispatchPort"/>.
+/// </para>
+/// <para>
+/// The background task only does external I/O and finishes by signalling the actor; it
+/// never reads or writes actor state directly. All actor-state mutations happen inside
+/// the actor's handler when the dispatched event arrives, preserving the actor's
+/// single-threaded execution invariant.
+/// </para>
+/// </remarks>
+public sealed class ConversationLlmReplyExecutor : IConversationLlmReplyExecutor
 {
-    internal const string InboxStreamId = "channel-runtime:llm-reply:inbox";
+    internal const string PublisherActorId = "channel-runtime.llm-reply-executor";
 
-    private readonly IStreamProvider _streamProvider;
-    private readonly IActorRuntime _actorRuntime;
     private readonly IActorDispatchPort _actorDispatchPort;
     private readonly IConversationReplyGenerator _replyGenerator;
     private readonly IInteractiveReplyCollector? _interactiveReplyCollector;
@@ -28,26 +41,19 @@ public sealed class ChannelLlmReplyInboxRuntime :
     private readonly INyxIdRelayScopeResolver? _scopeResolver;
     private readonly IUserConfigQueryPort? _userConfigQueryPort;
     private readonly TimeProvider _timeProvider;
-    private readonly ILogger<ChannelLlmReplyInboxRuntime> _logger;
-    private IAsyncDisposable? _subscription;
+    private readonly ILogger<ConversationLlmReplyExecutor> _logger;
 
-    public ChannelLlmReplyInboxRuntime(
-        IStreamProvider streamProvider,
-        IActorRuntime actorRuntime,
+    public ConversationLlmReplyExecutor(
+        IActorDispatchPort actorDispatchPort,
         IConversationReplyGenerator replyGenerator,
         IInteractiveReplyCollector? interactiveReplyCollector,
         Aevatar.GAgents.Channel.NyxIdRelay.NyxIdRelayOptions? relayOptions,
-        ILogger<ChannelLlmReplyInboxRuntime> logger,
+        ILogger<ConversationLlmReplyExecutor> logger,
         INyxIdRelayScopeResolver? scopeResolver = null,
         IUserConfigQueryPort? userConfigQueryPort = null,
-        TimeProvider? timeProvider = null,
-        IActorDispatchPort? actorDispatchPort = null)
+        TimeProvider? timeProvider = null)
     {
-        _streamProvider = streamProvider ?? throw new ArgumentNullException(nameof(streamProvider));
-        _actorRuntime = actorRuntime ?? throw new ArgumentNullException(nameof(actorRuntime));
-        _actorDispatchPort = actorDispatchPort
-            ?? actorRuntime as IActorDispatchPort
-            ?? throw new ArgumentNullException(nameof(actorDispatchPort));
+        _actorDispatchPort = actorDispatchPort ?? throw new ArgumentNullException(nameof(actorDispatchPort));
         _replyGenerator = replyGenerator ?? throw new ArgumentNullException(nameof(replyGenerator));
         _interactiveReplyCollector = interactiveReplyCollector;
         _relayOptions = relayOptions;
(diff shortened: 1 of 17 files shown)


agents/Aevatar.GAgents.NyxidChat/ConversationLlmReplyExecutor.cs
```diff
@@ -111,6 +90,42 @@ public async ValueTask DisposeAsync()
...
 
+    public Task StartAsync(NeedsLlmReplyEvent request, CancellationToken ct)
+    {
+        ArgumentNullException.ThrowIfNull(request);
+        // Snapshot the request: the caller may mutate the original (e.g. the actor
+        // re-enriches with a fresh reply token on retry) and we are about to hand off
+        // ownership to a background task.
+        var snapshot = request.Clone();
+
+        // Use Task.Run so the LLM work runs on the thread pool, not on the caller's
```
eanzhao, line 106: **[Consensus: 3 models]** severity=minor, category=design

The new `Task.Run` launch ignores the caller's `CancellationToken` (passes `CancellationToken.None`) and has no host-shutdown coordination — neither caller-token cancellation nor application-stopping events can stop an in-flight LLM call. Three reviewers flagged variants of this with different fix proposals; pick whichever fits (or none, if accepted as durability-by-rehydration).

<details>

- **deepseek-v4-pro**: `StartAsync` spawns fire-and-forget `Task.Run` work with `CancellationToken.None` and no tracking (no `ConcurrentBag<Task>` / `CountdownEvent` / `IHostedService.StopAsync` integration). In-flight LLM calls (up to 300s) that outlive the host's graceful-shutdown window will run into a tearing-down `IActorDispatchPort`, producing `ObjectDisposedException`-style noise in shutdown logs. Durability is covered by actor state rehydration, so this won't cause data loss, but consider a minimal shutdown gate: track in-flight tasks, expose `IAsyncDisposable` that signals a shared `CancellationTokenSource` and `await`s the pending set within a bounded drain window (e.g. 5s), then let any stragglers die naturally.
- **glm-5.1**: `StartAsync` accepts a `CancellationToken ct` parameter but passes `CancellationToken.None` to `Task.Run`, completely ignoring the caller's token. At minimum, add `ct.ThrowIfCancellationRequested()` before the `Task.Run` call so that a caller whose turn is already cancelled does not wastefully spawn an LLM round that may race the rehydration path. More robustly, pass `ct` to `Task.Run` as the second argument so the task is never scheduled if cancellation is already signaled:

  ```csharp
  ct.ThrowIfCancellationRequested();
  var snapshot = request.Clone();
  _ = Task.Run(async () => { ... }, ct);
  ```
- **mimo-v2.5-pro**: `_ = Task.Run(async () => { ... }, CancellationToken.None)` — the fire-and-forget pattern is correct for the non-blocking requirement, but there is no coordinated shutdown path. If the host shuts down while LLM calls are in-flight, `ProcessAsync` runs to completion (or its internal timeout) with no host-level cancellation. The old `IHostedService.StopAsync` → `DisposeAsync` at least tore down the stream subscription. Consider accepting a `CancellationToken` from a shared host-lifetime source (e.g. `IHostApplicationLifetime.ApplicationStopping`) to allow the internal `CancellationTokenSource` in `ProcessAsync` to be cancelled early on shutdown.

</details>

eanzhao, line 106: **[Consensus: 3 models]** severity=minor, category=design

The new `Task.Run` launch ignores the caller's `CancellationToken` (passes `CancellationToken.None`) and has no host-shutdown coordination — neither caller-token cancellation nor application-stopping events can stop an in-flight LLM call. Three reviewers flagged variants of this with different fix proposals; pick whichever fits (or none, if accepted as durability-by-rehydration).

<details>

- **deepseek-v4-pro**: `StartAsync` spawns fire-and-forget `Task.Run` work with `CancellationToken.None` and no tracking (no `ConcurrentBag<Task>` / `CountdownEvent` / `IHostedService.StopAsync` integration). In-flight LLM calls (up to 300s) that outlive the host's graceful-shutdown window will run into a tearing-down `IActorDispatchPort`, producing `ObjectDisposedException`-style noise in shutdown logs. Durability is covered by actor state rehydration, so this won't cause data loss, but consider a minimal shutdown gate: track in-flight tasks, expose `IAsyncDisposable` that signals a shared `CancellationTokenSource` and `await`s the pending set within a bounded drain window (e.g. 5s), then let any stragglers die naturally.
- **glm-5.1**: `StartAsync` accepts a `CancellationToken ct` parameter but passes `CancellationToken.None` to `Task.Run`, completely ignoring the caller's token. At minimum, add `ct.ThrowIfCancellationRequested()` before the `Task.Run` call so that a caller whose turn is already cancelled does not wastefully spawn an LLM round that may race the rehydration path. More robustly, pass `ct` to `Task.Run` as the second argument so the task is never scheduled if cancellation is already signaled:

  ```csharp
  ct.ThrowIfCancellationRequested();
  var snapshot = request.Clone();
  _ = Task.Run(async () => { ... }, ct);
  ```
- **mimo-v2.5-pro**: `_ = Task.Run(async () => { ... }, CancellationToken.None)` — the fire-and-forget pattern is correct for the non-blocking requirement, but there is no coordinated shutdown path. If the host shuts down while LLM calls are in-flight, `ProcessAsync` runs to completion (or its internal timeout) with no host-level cancellation. The old `IHostedService.StopAsync` → `DisposeAsync` at least tore down the stream subscription. Consider accepting a `CancellationToken` from a shared host-lifetime source (e.g. `IHostApplicationLifetime.ApplicationStopping`) to allow the internal `CancellationTokenSource` in `ProcessAsync` to be cancelled early on shutdown.

</details>

agents/Aevatar.GAgents.NyxidChat/ConversationLlmReplyExecutor.cs
```diff
@@ -111,6 +90,42 @@ public async ValueTask DisposeAsync()
...
+        // synchronization context (which, for an actor turn, must not be blocked or the
+        // very bottleneck this seam removes is reintroduced inside one actor). The
+        // background task only does external I/O and finishes by dispatching back to
+        // the actor — it never reads or writes actor state.
+        _ = Task.Run(async () =>
```
eanzhao, line 114: **[codex]** severity=major, category=concurrency

`Task.Run` returns immediately for every pending request, so one `ConversationGAgent` can now start multiple LLM replies for the same conversation before the first one finishes. That removes the only ordering boundary the old silo-FIFO provided and can produce out-of-order replies or interleaved streaming/ready signals for a single user (e.g. user types two messages quickly; both `DispatchPendingLlmReplyAsync` calls fire in turn order, but their `Task.Run` LLM calls finish in arbitrary order).

Keep the parallelism *across* different conversation actors (the headline win of this PR), but add an actor-owned in-flight gate in `ConversationGAgent` so only one correlation per conversation is started at a time — block subsequent `StartAsync` calls until the current correlation's `LlmReplyReadyEvent` / `DeferredLlmReplyDroppedEvent` retires it. The gate lives in actor state, which already tracks `PendingLlmReplyRequests`.

If you've decided per-conversation interleaving is acceptable (or already prevented elsewhere I'm not seeing), please add a comment in `DispatchPendingLlmReplyAsync` explaining why the FIFO loss is OK for one user's stream.

agents/Aevatar.GAgents.NyxidChat/ConversationLlmReplyExecutor.cs
```diff
@@ -292,7 +300,7 @@ private async Task DispatchReadyEventAsync(
             Id = Guid.NewGuid().ToString("N"),
             Timestamp = Timestamp.FromDateTimeOffset(_timeProvider.GetUtcNow()),
             Payload = Any.Pack(ready),
-            Route = EnvelopeRouteSemantics.CreateDirect(InboxStreamId, request.TargetActorId),
+            Route = EnvelopeRouteSemantics.CreateDirect(PublisherActorId, request.TargetActorId),
```
eanzhao, line 326: **[Consensus: 2 models]** severity=major, category=bug

If `DispatchAsync` (this line) or any code in `ProcessAsync` throws **after** the pre-LLM gates pass — e.g. transient `IActorDispatchPort` failure, target actor not yet initialized, network blip — the outer `catch` in `StartAsync` only logs and swallows the exception, so the `IConversationLlmReplyExecutor` contract's required terminal signal (`LlmReplyReadyEvent` or `DeferredLlmReplyDroppedEvent`) is never delivered. `State.PendingLlmReplyRequests` then leaks until the 5-minute stale-age retry fires.

Fix options:
1. Dispatch a terminal `DeferredLlmReplyDroppedEvent` (reason like `executor_crash`) from the outer `catch` in `StartAsync`, so the actor can retire the pending entry immediately. Cheapest fix.
2. Convert dispatch failures into an actor-owned durable retry path (more work, but matches the durability story for transient failures).

Either way, add a test with a throwing `IActorDispatchPort` to lock in the chosen behavior.

<details>

- **kimi** (at executor.cs:112, the outer `catch`): The outer `catch` block in `StartAsync` logs the crash but does not dispatch a terminal signal, violating the `IConversationLlmReplyExecutor` contract ("Implementations MUST eventually deliver one terminal signal... Without a terminal signal, the actor's pending entry would leak"). If `ProcessAsync` crashes after its internal gates but before dispatching (e.g. `_actorDispatchPort` throws on send), the actor's `State.PendingLlmReplyRequests` entry will linger until the 5-minute stale-age retry fires. Consider calling `NotifyActorOfDropAsync` in the catch block with reason `executor_crash` to honor the contract and allow the actor to retire the pending entry immediately.
- **codex** (at executor.cs:303, the dispatch site): If `DispatchAsync` throws here (for example Orleans reports the target actor is not initialized, or dispatch has a transient failure), the outer fire-and-forget catch only logs and swallows it, so the `IConversationLlmReplyExecutor` contract's required terminal signal is never delivered and `State.PendingLlmReplyRequests` can leak indefinitely. Convert dispatch failure into an actor-owned durable retry path (or otherwise guarantee a terminal signal) and add a test with a throwing `IActorDispatchPort`.

</details>

agents/Aevatar.GAgents.NyxidChat/ConversationLlmReplyExecutor.cs
```diff
@@ -111,6 +90,42 @@ public async ValueTask DisposeAsync()
     /// </summary>
     internal static readonly TimeSpan MetadataBuildBudget = TimeSpan.FromSeconds(15);
 
+    public Task StartAsync(NeedsLlmReplyEvent request, CancellationToken ct)
```
eanzhao, line 93: **[kimi]** severity=major, category=test

`StartAsync` — the primary public entry point of this seam — is not exercised by any unit test. The renamed `ConversationLlmReplyExecutorTests` still calls `ProcessAsync` directly, bypassing the `Task.Run` wrapping, exception-swallowing semantics, and non-blocking guarantee that are the whole point of this refactor.

Add at least these tests:

1. `StartAsync` returns a completed task immediately without awaiting the LLM call (e.g. fake LLM that blocks on a `TaskCompletionSource` — verify `StartAsync` returns before the TCS is signaled).
2. The background task eventually invokes `ProcessAsync` and dispatches the expected terminal signal.
3. An unexpected exception inside `ProcessAsync` is logged and does not surface as an unobserved task exception (i.e. the outer catch does its job — pairs with Issue 3 above).
4. The `request` is snapshotted (`Clone()`) before being handed to the background task, so post-`StartAsync` mutation by the caller can't affect the in-flight LLM call.

Without these, the non-blocking semantics that justify this PR are unverified — a future refactor could silently re-await the LLM call inside `StartAsync` and every existing test would still pass.

PR #625, opened 2026-05-12
Scaffold Responses API v1 prototype

eanzhao:
Closes #629

## Summary
- Adds `POST /v1/responses` ingress, opaque `response.id` ownership actor (`ResponseSessionGAgent`), forwarded tool call state actor (`ResponsesAgentToolStateGAgent`), and current-state projection plumbing for both
- Wires tool classifier (forward / substitute / additive) and caller scope resolution at the protocol boundary
- First slice for milestone 14 (Responses API v1 Minimal Prototype); also satisfies the acceptance items of the fine-grained planning issues #611, #612, #614, #617, #618, #621, #622

## Test plan
- [ ] `dotnet build aevatar.slnx --nologo`
- [ ] `dotnet test aevatar.slnx --nologo`
- [ ] CI architecture guards (`bash tools/ci/architecture_guards.sh`)
- [ ] Smoke: curl `POST /v1/responses` with crafted `X-NyxID-Identity-Token`, verify SSE returns
- [ ] Read-side: verify `ResponseSession` + `ResponsesAgentToolState` current-state readmodels populate after a request


src/Aevatar.Mainnet.Host.Api/Responses/ResponsesAevatarToolProvider.cs
```diff
@@ -0,0 +1,451 @@
...
+                cacheKey,
+                ct);
+            if (cached != null)
+            {
+                await RecordTraceAsync(scope, cacheKey, url, query: string.Empty, cacheHit: true, cached.ResultJson, ct);
+                return cached.ResultJson;
+            }
+
+            var token = AgentToolRequestContext.TryGet(LLMRequestMetadataKeys.NyxIdAccessToken) ?? string.Empty;
+            var result = await _webClient.FetchUrlAsync(token, url, ct);
```
eanzhao: **[codex]** severity=blocker, category=security

`WebFetch` passes the caller's NyxID bearer token to `_webClient.FetchUrlAsync(token, url, ct)` for an LLM/user-supplied URL. `WebApiClient.FetchUrlAsync` attaches that token as an `Authorization` header to the target URL, so a prompt can exfiltrate the user's token to an attacker-controlled host and also SSRF internal addresses. Do not send the bearer token to arbitrary fetch targets; route through a validated NyxID proxy or call fetch with no token, and add URL validation/private-network blocking before making the request.

src/platform/Aevatar.GAgentService.Core/GAgents/ResponseSessionGAgent.cs
```diff
@@ -0,0 +1,499 @@
...
+    }
+
+    private static DateTimeOffset ResolveExpiry(ResponseSessionRecord record)
+    {
+        var createdAt = record.CreatedAt?.ToDateTimeOffset() ?? DateTimeOffset.UtcNow;
+        var ttl = record.Ttl?.ToTimeSpan() ?? DefaultTtl.ToTimeSpan();
+        return createdAt.Add(ttl);
+    }
+
+    private static string BuildExpiredToolCallResult(string? callId) =>
```
eanzhao: **[mimo-v2.5-pro]** severity=blocker, category=arch

`System.Text.Json.JsonSerializer.Serialize` used to produce `result_json` inside a domain actor. Violates CLAUDE.md serialization constraint: "统一 Protobuf … 禁止 JSON/XML/自定义字符串格式用于 Actor State … 事实存储". The `BuildExpiredToolCallResult` helper creates a JSON string that is stored in `ResponseSessionForwardedToolCall.ResultJson` and persisted via event sourcing. Move the JSON construction to the Infrastructure adapter or define a typed protobuf message for the expired-tool-call payload and serialize only at the host boundary.

src/platform/Aevatar.GAgentService.Core/GAgents/ResponsesAgentToolStateGAgent.cs
```diff
@@ -0,0 +1,423 @@
+using System.Security.Cryptography;
+using System.Text;
+using System.Text.Json;
```
eanzhao: **[mimo-v2.5-pro]** severity=blocker, category=arch

`System.Text.Json` imported and used extensively in the domain actor for parsing `argumentsJson` (ParseTodoItems, ExtractQuery, etc.). This is the same serialization-rule violation as in `ResponseSessionGAgent`. The todo-item parse/normalize logic should either live in the Infrastructure adapter (`ResponsesAgentToolStateCommandAdapter`) or the `arguments_json` field should be typed protobuf rather than a raw JSON string that the domain actor deserializes. Same applies to `ResponseSessionGAgent.cs:7`.

src/Aevatar.Mainnet.Host.Api/Hosting/MainnetHostBuilderExtensions.cs
```diff
@@ -86,6 +88,8 @@ public static WebApplicationBuilder AddAevatarMainnetHost(
         builder.Services.AddChannelIdentityProjectionStores(builder.Configuration);
         builder.Services.AddDeviceRegistration(builder.Configuration);
         builder.Services.AddScheduledAgents(builder.Configuration);
+        builder.Services.TryAddSingleton<IResponsesCallerScopeResolver, NyxIdResponsesCallerScopeResolver>();
```
eanzhao, line 92: **[Consensus: 2 models]** severity=major, category=di

Both reviewers flag captive dependency: `IResponsesCallerScopeResolver` registered as `TryAddSingleton` here depends on `INyxIdCurrentUserResolver`, and the same Singleton-on-possibly-Scoped concern applies to `ResponsesAevatarToolProvider` -> `WebApiClient`/`WebToolOptions`. If any of those resolvers are Scoped (common for HTTP-context-bound services), the cached Singleton holds a stale reference per request. Verify lifetimes and downgrade to Scoped if any dependency is request-scoped.

<details>

- **v4-pro**: `IResponsesCallerScopeResolver` is registered as `TryAddSingleton`, but its only implementation `NyxIdResponsesCallerScopeResolver` depends on `INyxIdCurrentUserResolver`. If `INyxIdCurrentUserResolver` is registered as Scoped (common for HTTP-context-bound resolvers), this creates a captive dependency — the singleton holds a scoped reference that will be stale or disposed on subsequent requests. Verify the lifetime of `INyxIdCurrentUserResolver` and either make this registration Scoped or confirm the dependency is also Singleton.
- **glm-5.1**: `ResponsesAevatarToolProvider` is registered as `Singleton` and depends on `WebApiClient` and `WebToolOptions`. The `AddWebTools` call registers these, but their lifetimes cannot be verified from the diff. If `WebApiClient` or `WebToolOptions` are registered as `Scoped`/`Transient`, this is a captive dependency.

Similarly, `NyxIdResponsesCallerScopeResolver` (Singleton) depends on `INyxIdCurrentUserResolver` — if that resolver is scoped to the HTTP request, the cached instance will go stale or capture a disposed `HttpClient`.

**Verify:** Confirm all transitive dependencies of Singleton services are also Singleton.
</details>

src/Aevatar.Mainnet.Host.Api/Responses/ResponsesEndpoints.cs
```diff
@@ -0,0 +1,1485 @@
...
+        group.MapPost("/responses/{id}/cancel", HandleCancelResponseAsync);
+        return app;
+    }
+
+    [SuppressMessage(
+        "Maintainability",
+        "CA1506:Avoid excessive class coupling",
+        Justification = "This Minimal API adapter coordinates one external Responses endpoint across HTTP, " +
+                        "caller scope, durable session registration, and SSE shaping.")]
+    internal static async Task<IResult> HandleCreateResponseAsync(
```
eanzhao, line 39: **[Consensus: 2 models]** severity=major, category=arch

Both reviewers flag the same architectural rule violation: `HandleCreateResponseAsync` in this 1485-line file owns scope resolution, previous-response validation, session registration, multi-round LLM tool loop, forwarded-tool persistence, SSE framing, and status lifecycle. Per CLAUDE.md "API 仅做宿主与组合，不承载业务编排", extract this into an Application-layer port/service (e.g. `IResponsesOrchestrationPort` / `ResponsesOrchestrationService`) and leave the endpoint to handle only HTTP↔domain shaping (auth context extraction, SSE/JSON framing).

<details>

- **mimo-v2.5-pro**: `HandleCreateResponseAsync` (together with the private helpers in this file) orchestrates the full Responses business flow: scope resolution, session registration, tool classification, LLM message construction, tool-aware multi-round completion, tool-call persistence, SSE framing, and status lifecycle. Per CLAUDE.md "API 仅做宿主与组合，不承载业务编排", this orchestration logic belongs in an Application-layer service; the endpoint should only extract HTTP context and delegate. The 1485-line file is itself a signal. Consider extracting a `ResponsesService` in the Application layer for the orchestration.
- **codex**: `HandleCreateResponseAsync` is a Host/API endpoint but owns request normalization, previous-response validation, session registration, LLM tool-loop execution, forwarded-tool persistence, and status transitions. That violates CLAUDE.md: API only does hosting/composition and should not carry core business orchestration. Move this flow behind an application-layer port/service (for example a Responses command service) and leave the endpoint responsible only for HTTP auth/binding and SSE/JSON transport shaping.
</details>

src/Aevatar.AI.Abstractions/LLMProviders/LLMRequestMetadataKeys.cs
```diff
@@ -4,6 +4,9 @@ public static class LLMRequestMetadataKeys
 {
     public const string RequestId = "aevatar.request_id";
     public const string CallId = "aevatar.call_id";
+    public const string ScopeId = "aevatar.scope_id";
```
eanzhao, line 7: **[kimi]** severity=major, category=arch

The PR adds `ScopeId`, `OwnerSubject`, and `ResponseId` as string keys in the `LLMRequest.Metadata` bag. These are core semantic fields that drive business control flow (tool scoping, session linking, caller identity). CLAUDE.md mandates: "核心语义强类型：影响业务语义、控制流、稳定读取且仓库内可控的数据，必须建模为 `proto field / typed option / typed sub-message`，禁止塞入通用 bag。" Add typed fields to `LLMRequest` (or a typed sub-message) instead of stuffing them into the metadata dictionary.

src/Aevatar.Mainnet.Host.Api/Responses/ResponsesAevatarToolProvider.cs
```diff
@@ -0,0 +1,451 @@
+using System.Security.Cryptography;
+using System.Text;
+using System.Text.Json;
+using Aevatar.AI.Abstractions.LLMProviders;
+using Aevatar.AI.Abstractions.ToolProviders;
+using Aevatar.AI.ToolProviders.Web;
+using Aevatar.GAgentService.Abstractions;
+using Aevatar.GAgentService.Abstractions.Ports;
+
```
eanzhao, line 10: **[kimi]** severity=major, category=arch

`ResponsesAevatarToolProvider` implements `IAgentTool` substitutes (`TodoWriteTool`, `TaskTool`, `WebFetchTool`, `WebSearchTool`) directly in the Host layer. These tools contain business logic (agent-scoped todo state, task topology, web caching). Per CLAUDE.md dependency-inversion and boundary rules, business capabilities belong in Application/Domain/AI layers, not the API host. Move the provider and its tools out of Host.

src/Aevatar.Mainnet.Host.Api/Responses/ResponsesAevatarToolProvider.cs
```diff
@@ -0,0 +1,451 @@
...
+namespace Aevatar.Mainnet.Host.Api.Responses;
+
+internal sealed class ResponsesAevatarToolProvider : IResponsesToolProvider
+{
+    private readonly IResponsesAgentToolStateCommandPort _commandPort;
+    private readonly IResponsesAgentToolStateQueryPort _queryPort;
+    private readonly WebApiClient _webClient;
```
eanzhao: **[kimi]** severity=major, category=di

`ResponsesAevatarToolProvider` injects the concrete class `WebApiClient` instead of an abstraction. This violates CLAUDE.md dependency inversion: "上层依赖抽象，禁止跨层反向依赖和对具体实现的直接耦合" (upper layers depend on abstractions, no direct coupling to concrete implementations). Introduce and inject an interface such as `IWebApiClient`.

src/Aevatar.Mainnet.Host.Api/Responses/ResponsesEndpoints.cs
```diff
@@ -0,0 +1,1485 @@
...
+        [FromServices] ILoggerFactory loggerFactory,
+        CancellationToken ct)
+    {
+        ArgumentNullException.ThrowIfNull(http);
+        ArgumentNullException.ThrowIfNull(providerFactory);
+        ArgumentNullException.ThrowIfNull(callerScopeResolver);
+        ArgumentNullException.ThrowIfNull(responseSessionRegistrationPort);
+        ArgumentNullException.ThrowIfNull(responseSessionQueryPort);
+        ArgumentNullException.ThrowIfNull(toolProviders);
+        ArgumentNullException.ThrowIfNull(loggerFactory);
```
eanzhao, line 58: **[kimi]** severity=major, category=arch

This Minimal API handler (`HandleCreateResponseAsync`) and its ~1,400-line file contain the full Responses protocol orchestration: request normalization, tool classification, LLM message assembly, SSE streaming, session registration, forwarded-tool-call reconciliation, and response building. This violates CLAUDE.md strict layering: "API 仅做宿主与组合，不承载核心业务编排" (API is only for hosting and composition, no business orchestration). Extract an Application-layer port/service (e.g., `IResponsesOrchestrationPort`) so the endpoint only deserializes input, invokes the port, and shapes the HTTP response.

src/Aevatar.Mainnet.Host.Api/Responses/ResponsesEndpoints.cs
```diff
@@ -0,0 +1,1485 @@
...
+        try
+        {
+            responseSession = await responseSessionRegistrationPort.RegisterAsync(
+                BuildResponseSessionRecord(normalized, callerScope, createdAt),
+                ct);
+        }
+        catch (OperationCanceledException)
+        {
+            return Results.StatusCode(499);
+        }
```
eanzhao, line 134: **[v4-pro]** severity=major, category=security

The raw NyxID bearer access token (from `ExtractBearerToken`) is placed into the `LLMRequest.Metadata` dictionary under key `LLMRequestMetadataKeys.NyxIdAccessToken`. This dictionary is then set on `AgentToolRequestContext.CurrentMetadata` (a global ambient mutable slot) and passed into the LLM provider via `request.Metadata`. Any middleware, logging sink, or tool provider that serializes or dumps `AgentToolRequestContext.CurrentMetadata` will leak the token to logs, telemetry, or tool output. Instead, pass the token through a narrowly-scoped mechanism (e.g., only set it on `AgentToolRequestContext` inside the tight `try/finally` blocks and strip it from `LLMRequest.Metadata` before the dictionary leaves the endpoint handler).

src/Aevatar.Mainnet.Host.Api/Responses/ResponsesEndpoints.cs
```diff
@@ -0,0 +1,1485 @@
...
+        catch (Exception ex) when (ex is not OperationCanceledException)
+        {
+            await TryUpdateSessionStatusAsync(
+                responseSessionRegistrationPort,
+                logger,
+                responseSession,
+                ResponseSessionStatus.Failed,
+                CancellationToken.None);
+            return ToErrorResult(
+                StatusCodes.Status500InternalServerError,
```
eanzhao, line 299: **[kimi]** severity=major, category=security

The NyxID bearer token is placed into `LLMRequest.Metadata` with key `LLMRequestMetadataKeys.NyxIdAccessToken` and passed to the LLM provider. If the provider or its underlying transport logs the request or metadata, the bearer token may be persisted in logs or telemetry. Pass the token through a dedicated secure mechanism (e.g., a typed authentication header field on the provider contract) rather than a general metadata bag.

src/Aevatar.Mainnet.Host.Api/Responses/ResponsesEndpoints.cs
```diff
@@ -0,0 +1,1485 @@
...
+        if (visibleSnapshot.Status == ResponseSessionStatus.Expired)
+        {
+            return ToErrorResult(
+                StatusCodes.Status400BadRequest,
+                "response_expired",
+                "response id refers to an expired response session.");
+        }
+
+        var cancelledAt = DateTimeOffset.UtcNow.ToUnixTimeSeconds();
+        if (visibleSnapshot.Status != ResponseSessionStatus.Cancelled)
```
eanzhao, line 379: **[kimi]** severity=major, category=security

The generic exception handler returns `ex.Message` directly in the HTTP response (`ToErrorResult(..., ex.Message)`). This can leak internal implementation details (upstream error text, connection info, etc.) to untrusted clients. Return a generic error with a correlation ID, and log the full exception server-side.

src/Aevatar.Mainnet.Host.Api/Responses/ResponsesEndpoints.cs
```diff
@@ -0,0 +1,1485 @@
...
+                    "Failed to mark forwarded Responses tool call {CallId} as resolved for response {ResponseId}.",
+                    callId,
+                    previousSnapshot.ResponseId);
+            }
+        }
+    }
+
+    private static async Task PersistForwardedToolCallsAsync(
+        IResponseSessionRegistrationPort responseSessionRegistrationPort,
+        ILogger logger,
```
eanzhao, line 1004: **[glm-5.1]** severity=major, category=arch

The 8-round tool-call loop (`for (var round = 0; round < 8; round++)`) with classification, execution, and message accumulation is significant business orchestration living in the Host layer. Per AGENTS.md: "API 仅做宿主与组合，不承载核心业务编排". The same loop exists in both `CollectToolAwareCompletionAsync` and `StreamToolAwareCompletionAsync`.

**Fix:** Extract this into an application-level service (e.g. `IResponsesCompletionOrchestrator`) in the Application layer. The endpoint should only handle HTTP↔domain mapping. The magic number `8` should also become a named constant or configuration value.

src/Aevatar.Mainnet.Host.Api/Responses/ResponsesEndpoints.cs
```diff
@@ -0,0 +1,1485 @@
...
+
+                    if (chunk.Usage != null)
+                        usage = MapUsage(chunk.Usage);
+
+                    if (chunk.IsLast)
+                        break;
+                }
+            }
+            finally
+            {
```
eanzhao: **[v4-pro]** severity=major, category=concurrency

`AgentToolRequestContext.CurrentMetadata` is a static/ambient mutable field. The code saves it to `previousMetadata`, overwrites it with per-request data, and restores in a `finally` block. This is safe only if all operations within the `try` block are synchronous or if `AgentToolRequestContext` uses `AsyncLocal<T>`. If it uses a plain `static` field (or `ThreadStatic`), two concurrent requests sharing a thread-pool thread could observe each other's tokens. Verify `AgentToolRequestContext` is `AsyncLocal`-backed; if not, this is a cross-request data leak.

src/Aevatar.Mainnet.Host.Api/Responses/ResponsesToolClassifier.cs
```diff
@@ -0,0 +1,134 @@
...
+
+        return new ResponsesToolClassification(
+            forwarded,
+            effective,
+            substitutedNames,
+            additiveTools.Select(static tool => tool.Name).ToArray());
+    }
+
+    private static bool RequiresSubstitution(string toolName)
+    {
```
eanzhao: **[glm-5.1]** severity=major, category=bug

`StartsWith("Todo")` matches any tool starting with "Todo" (e.g. `TodoRead`, `TodoArchive`, `TodoDelete`), not just `TodoWrite`. Only `TodoWrite` is registered as a substitute in `ResponsesAevatarToolProvider.GetSubstituteTools()`, so other `Todo*` tools will silently get the `ResponsesUnavailableSubstituteTool` stub instead of being forwarded to the client.

**Fix:** Either add explicit names to `SubstituteToolNames` (e.g. `"TodoWrite"`, `"todo_write"`) or restrict the prefix match to known names.

src/platform/Aevatar.GAgentService.Abstractions/Protos/response_sessions.proto
```diff
@@ -0,0 +1,249 @@
...
+  google.protobuf.Duration ttl = 8;
+  google.protobuf.Timestamp cancelled_at = 9;
+  google.protobuf.Timestamp updated_at = 10;
+}
+
+message ResponseSessionForwardedToolCall {
+  string call_id = 1;
+  string tool_name = 2;
+  string schema_hash = 3;
+  string arguments_json = 4;
```
eanzhao: **[codex]** severity=major, category=arch

Persisting `arguments_json`/`result_json` in `ResponseSessionForwardedToolCall` makes JSON strings part of actor state and events; the same pattern is repeated for task/web traces. This violates CLAUDE.md: Serialization, which requires Actor State/domain events/internal persistence to use Protobuf rather than JSON strings. Keep raw external JSON at the Host adapter boundary, or model known payloads as typed proto submessages/`Any` plus an explicit opaque boundary contract instead of storing `*_json` fields in state/readmodels.

src/platform/Aevatar.GAgentService.Core/GAgents/ResponseSessionGAgent.cs
```diff
@@ -0,0 +1,499 @@
...
+                $"Response session actor '{Id}' is bound to response '{existing.ResponseId}' and cannot update response '{responseId}'.");
+        }
+
+        if (command.Status == ResponseSessionStatus.Unspecified)
+            return;
+
+        if (existing.Status == command.Status)
+            return;
+
+        await PersistDomainEventAsync(new ResponseSessionStatusUpdatedEvent
```
eanzhao, line 77: **[codex]** severity=major, category=bug

`HandleUpdateStatusAsync` accepts any status transition once the status differs, so a `/cancel` can be overwritten later by the still-running create path's `Completed` update. That makes cancellation non-authoritative and can also move `Failed`/`Expired` sessions back to `Completed`. Add an explicit state machine/CAS guard that treats `Cancelled`, `Expired`, `Failed`, and probably `Completed` as terminal except for idempotent repeats, and have callers handle rejected transitions.

src/platform/Aevatar.GAgentService.Infrastructure/Adapters/ResponsesAgentToolStateCommandAdapter.cs
```diff
@@ -0,0 +1,292 @@
...
+            {
+                CorrelationId = commandId,
+            },
+        };
+
+    private static IReadOnlyList<ResponsesTodoItemSnapshot> PreviewTodoItems(
+        string? argumentsJson,
+        string sourceResponseId,
+        DateTimeOffset observedAt)
+    {
```
eanzhao: **[glm-5.1]** severity=major, category=design

`PreviewTodoItems` duplicates the todo-item parsing logic from `ResponsesAgentToolStateGAgent.ParseTodoItems` (lines ~196-260 of that file). The adapter preview and the actor's authoritative parse can diverge, causing the returned `ResponsesTodoWriteResult` to disagree with what the actor actually persisted.

**Fix:** Extract a shared `ResponsesTodoParser` utility in Abstractions, or have the command port return the actor's actual persisted state instead of previewing client-side.

test/Aevatar.Hosting.Tests/MainnetResponsesEndpointsTests.cs
```diff
@@ -0,0 +1,1427 @@
...
+        provider.StreamCallCount.Should().Be(1);
+        provider.LastRequest.Should().NotBeNull();
+        provider.LastRequest!.Model.Should().Be("gpt-5.4");
+        provider.LastRequest.MaxTokens.Should().Be(128);
+        provider.LastRequest.Temperature.Should().Be(0.2);
+        provider.LastRequest.Messages.Should().ContainSingle();
+        provider.LastRequest.Messages[0].Content.Should().Be("ping");
+        provider.LastRequest.Metadata.Should().Contain(LLMRequestMetadataKeys.NyxIdAccessToken, "secret-token");
+        provider.LastRequest.Metadata.Should().ContainKey(LLMRequestMetadataKeys.RequestId);
+        provider.LastRequest.Metadata.Should().Contain("scope_id", "user-1");
```
eanzhao: **[codex]** severity=major, category=test

This assertion uses the literal key `scope_id`, but production writes the metadata under `LLMRequestMetadataKeys.ScopeId` (`aevatar.scope_id`). As written, the new test should fail even when the endpoint behaves as implemented. Change the assertion to `Contain(LLMRequestMetadataKeys.ScopeId, "user-1")`.

src/Aevatar.Mainnet.Host.Api/Responses/ResponsesAevatarToolProvider.cs
```diff
@@ -0,0 +1,451 @@
...
+    }
+
+    private sealed class WebFetchTool : ResponsesStateTool
+    {
+        private readonly string _name;
+        private readonly IResponsesAgentToolStateCommandPort _commandPort;
+        private readonly IResponsesAgentToolStateQueryPort _queryPort;
+        private readonly WebApiClient _webClient;
+
+        public WebFetchTool(
```
eanzhao, line 198: **[glm-5.1]** severity=minor, category=security

`WebFetchTool.NormalizeUrl` silently upgrades `http://` to `https://` and prepends `https://` to schemeless URLs. There is no SSRF validation — internal/private IP ranges (`127.0.0.1`, `10.x.x.x`, `169.254.x.x`) are not rejected. In a production prototype, an LLM could be prompted to fetch internal metadata endpoints (e.g. cloud instance metadata at `169.254.169.254`).

**Fix:** Add URL validation that rejects private/loopback/link-local ranges before fetching.

src/Aevatar.Mainnet.Host.Api/Responses/ResponsesEndpoints.cs
```diff
@@ -0,0 +1,1485 @@
...
+            var forwardedToolCalls = completion.ForwardedToolCalls;
+            await PersistForwardedToolCallsAsync(
+                responseSessionRegistrationPort,
+                logger,
+                responseSession,
+                toolClassification,
+                forwardedToolCalls,
+                DateTimeOffset.UtcNow,
+                ct);
+            await TryResolveIncomingToolResultsAsync(
```
eanzhao, line 220: **[mimo-v2.5-pro]** severity=minor, category=bug

`TryResolveIncomingToolResultsAsync` and `TryUpdateSessionStatusAsync` are called with `CancellationToken.None` instead of the request-scoped `ct`. If the client disconnects mid-request, these fire-and-forget operations will still run to completion — wasting resources on a response nobody will receive. Pass `ct` instead of `CancellationToken.None` (same at lines 206, 468, 522 in the streaming path).

src/Aevatar.Mainnet.Host.Api/Responses/ResponsesEndpoints.cs
```diff
@@ -0,0 +1,1485 @@
...
+        try
+        {
+            callerScope = await callerScopeResolver.ResolveAsync(bearerToken, http, ct);
+        }
+        catch (ResponsesCallerScopeUnavailableException ex)
+        {
+            return ToErrorResult(StatusCodes.Status401Unauthorized, "authentication_required", ex.Message);
+        }
+
+        var snapshot = await responseSessionQueryPort.GetByResponseIdAsync(responseId, ct);
```
eanzhao, line 360: **[kimi]** severity=minor, category=bug

`OperationCanceledException` is mapped to `Results.StatusCode(499)`. 499 is a non-standard nginx-specific code, not defined in HTTP/1.1 or standard ASP.NET Core conventions. Let the framework handle cancellation (it will typically abort the response stream), or return `StatusCodes.Status408RequestTimeout` if a status code is explicitly required.

src/Aevatar.Mainnet.Host.Api/Responses/ResponsesEndpoints.cs
```diff
@@ -0,0 +1,1485 @@
...
+    {
+        var messages = request.Messages.ToList();
+        var outputText = new StringBuilder();
+        ResponsesUsage? usage = null;
+
+        for (var round = 0; round < 8; round++)
+        {
+            var roundRequest = CloneRequestWithMessages(request, messages);
+            var toolCalls = new ResponsesToolCallAccumulator();
+            var previousMetadata = AgentToolRequestContext.CurrentMetadata;
```
eanzhao: **[kimi]** severity=minor, category=design

The endpoint sets `AgentToolRequestContext.CurrentMetadata` (an AsyncLocal ambient context) before calling the LLM provider and restores it in `finally`. While `AsyncLocal` is thread-safe, relying on ambient state in the Host layer is a service-locator anti-pattern. Prefer passing metadata explicitly through the call chain (e.g., as part of `LLMRequest` or tool call arguments).

src/Aevatar.Mainnet.Host.Api/Responses/ResponsesToolClassifier.cs
```diff
@@ -0,0 +1,134 @@
...
+    [
+        "Task",
+        "WebFetch",
+        "WebSearch",
+        "web_fetch",
+        "web_search",
+        "task",
+    ];
+
+    public static ResponsesToolClassification Classify(
```
eanzhao: **[v4-pro]** severity=minor, category=design

`SubstituteToolNames` is a hardcoded string array that must be kept in sync with the tools returned by `ResponsesAevatarToolProvider.GetSubstituteTools()`. If a new substitute tool is added to the provider but forgotten here, the tool name won't be recognized as needing substitution. Consider deriving this list dynamically from `providers.SelectMany(p => p.GetSubstituteTools()).Select(t => t.Name).ToHashSet()` instead of maintaining two parallel registries.

src/platform/Aevatar.GAgentService.Abstractions/Ports/IResponsesAgentToolStateCommandPort.cs
```diff
@@ -0,0 +1,55 @@
+using Aevatar.GAgentService.Abstractions.Queries;
```
eanzhao, line 1: **[mimo-v2.5-pro]** severity=minor, category=arch

This interface is named a *command* port but is actually mixed: `ApplyTodoWriteAsync`, `RecordTaskAsync`, `RecordWebTraceAsync` are commands, but the query methods (`GetAsync`, `GetWebCacheEntryAsync`) belong on the query port (`IResponsesAgentToolStateQueryPort`). The consumer (`ResponsesAevatarToolProvider`) correctly injects both ports separately; the recording test double conflates them. Clean the interface to contain only commands.

src/platform/Aevatar.GAgentService.Abstractions/Protos/response_sessions.proto
```diff
@@ -0,0 +1,249 @@
...
+  RESPONSES_AGENT_TOOL_TASK_STATUS_UNSPECIFIED = 0;
+  RESPONSES_AGENT_TOOL_TASK_STATUS_ACCEPTED = 1;
+  RESPONSES_AGENT_TOOL_TASK_STATUS_REJECTED = 2;
+}
+
+message ResponsesAgentToolStateRecord {
+  string scope_id = 1;
+  string owner_subject = 2;
+  google.protobuf.Timestamp created_at = 3;
+  google.protobuf.Timestamp updated_at = 4;
```
eanzhao, line 143: **[v4-pro]** severity=minor, category=design

`ResponsesAgentToolStateRecord` holds `scope_id` and `owner_subject` as plain strings, but the `actorId` is deterministically derived from these via `SHA256` in `ResponseAgentToolStateIds.BuildActorId`. This means two different `scope_id`/`owner_subject` combinations that hash-collide would map to the same actor. While SHA256 collision is practically impossible, the architecture rule says actor identity should not be cryptographically opaque — consider using a deterministic, non-hashed key format (e.g., `responses-agent-tools-{scopeId}/{ownerSubject}`) or at least document the hash truncation trade-off.

src/platform/Aevatar.GAgentService.Infrastructure/Adapters/ResponsesAgentToolStateCommandAdapter.cs
```diff
@@ -0,0 +1,292 @@
...
+                    ArgumentsJson = NormalizeOptional(argumentsJson) ?? "{}",
+                    ResultJson = resultJson,
+                    Status = ResponsesAgentToolTaskStatus.Accepted,
+                    ObservedAt = Timestamp.FromDateTime(DateTime.UtcNow),
+                }),
+                $"{sourceResponseId}:task:{taskId}"),
+            ct);
+
+        return new ResponsesTaskDispatchResult(actor.Id, taskId, childActorId, "accepted", resultJson);
+    }
```
eanzhao, line 117: **[v4-pro]** severity=minor, category=design

`EnsureActorAsync` dispatches a `RegisterResponsesAgentToolStateRequested` command on every call to ensure the actor exists. The GAgent's `HandleRegisterAsync` is idempotent (returns early if already registered), but this means every `TodoWrite`/`Task`/`WebTrace` operation performs an extra redundant dispatch. Consider separating actor creation from command dispatch: only register on first access, and use an `EnsureCreated` pattern that skips the dispatch when the actor state already indicates registration.

src/Aevatar.Mainnet.Host.Api/Responses/ResponsesToolClassifier.cs
```diff
@@ -0,0 +1,134 @@
...
+        IReadOnlyList<ResponsesToolDeclaration> declaredTools,
+        IEnumerable<IResponsesToolProvider> providers,
+        ILogger logger)
```
eanzhao: **[v4-pro]** severity=nit, category=style

Missing `TodoWrite` and `Todo` variants in `SubstituteToolNames` — `RequiresSubstitution` matches `toolName.StartsWith("Todo", ...) || toolName.StartsWith("todo", ...)` but the static list only has `Task`/`WebFetch`/`WebSearch` variants. This is inconsistent: the list doesn't include all de-facto substitute tool name prefixes.

PR #657, opened 2026-05-14
Fix GAgent durable terminal completion

### LINKED ISSUE #370: Implement durable terminal completion resolution for draft-run and approval interactions
## Problem
`GAgentDraftRunInteraction` and `GAgentApprovalInteraction` still use placeholder durable completion resolvers that return `Incomplete` instead of reading a durable terminal source.

## Why this matters
Happy-path live streaming works, but reconnect / dropped-stream / delayed-finalization scenarios cannot reliably recover terminal state from durable data. The result is user-visible sessions that can remain stuck in `Unknown` / `Incomplete` when the live sink misses the terminal event.

## Scope
Implement a real durable terminal completion path for draft-run and approval interactions, backed by a stable projection/read-model contract rather than a placeholder resolver.

## Acceptance criteria
- draft-run durable completion resolver reads a durable terminal source
- approval durable completion resolver reads a durable terminal source
- reconnect / missed-live-event flows can resolve `Completed` / `Failed` from durable state
- no query-time replay or ad hoc fallback logic is introduced
- targeted tests cover terminal recovery after live-stream interruption

louis4li:
## Summary

Fixes #370 by replacing the placeholder durable completion resolvers for GAgent draft-run and approval interactions with a committed-fact -> projection readmodel -> query-port path.

This PR adds a GAgent run terminal readmodel and durable materialization scope that observes committed `RoleChatSessionCompletedEvent` facts, then maps the resulting terminal snapshot back into draft-run / approval durable completion. It also fixes the approval live observation session key so `CommandId != CorrelationId` still receives the continuation terminal event.

## What changed

- Added `GAgentRunTerminalSnapshot`, `GAgentRunTerminalStatus`, `GAgentRunTerminalInteractionKind`, `IGAgentRunTerminalQueryPort`, and `IGAgentRunTerminalProjectionPort`.
- Added `GAgentRunTerminalReadModel`, metadata provider, projector, query reader, and durable projection activation port.
- Activated terminal durable materialization before draft-run / approval dispatch.
- Updated draft-run and approval durable resolvers to query by `actorId + correlationId`, then fallback to `actorId + sessionId` where applicable.
- Extended approval receipts with `SessionId` for fallback lookup.
- Persisted approval failure terminal facts for denied / timeout / remote failure paths so missed-live-event recovery has a committed source.
- Made durable materialization release session-aware through `IProjectionSessionScopedMaterializationContext`, so session/correlation-scoped terminal materialization can be released with the same key used for activation.
- Added design notes under `docs/history/2026-05/` documenting the architecture and follow-up decisions.

## Explicitly deferred

Two review points are intentionally not included in this PR:

1. **Replacing legacy approval / LLM failure markers with a typed terminal event**
   - Current implementation treats `RoleChatSessionCompletedEvent.Content` markers as a legacy fallback only.
   - A proper typed terminal event or typed `status/reason_code/reason_message` proto fields would change the committed event contract and should be designed/migrated separately.
   - The PR documents that live-only terminal events and marker strings are not the long-term durable protocol.

2. **Changing `ReasonCode` from string to enum**
   - `ReasonCode` is currently used for durable readmodel reason recording and completion mapping, not as a stable cross-module control-flow or filtering contract.
   - Keeping it as string avoids freezing an enum while legacy marker compatibility still exists.
   - If terminal reason later becomes API/SDK/filter/control-flow surface, it should move together with the typed terminal event work to proto enum / typed code.

## Validation

Previously run on this branch:

```bash
dotnet test test/Aevatar.CQRS.Projection.Core.Tests/Aevatar.CQRS.Projection.Core.Tests.csproj --nologo --filter "ProjectionRuntimeRegistrationTests"
dotnet test test/Aevatar.GAgentService.Tests/Aevatar.GAgentService.Tests.csproj --nologo --filter "GAgentDraftRunInteraction|GAgentApprovalInteraction|GAgentRunTerminal"
dotnet test test/Aevatar.GAgentService.Integration.Tests/Aevatar.GAgentService.Integration.Tests.csproj --nologo --filter "ScopeServiceEndpointsStreamTests|GAgentServiceHostingServiceCollectionExtensionsTests"
dotnet test test/Aevatar.AI.Tests/Aevatar.AI.Tests.csproj --nologo --filter "RoleGAgentStateCoverageTests"
bash tools/ci/query_projection_priming_guard.sh
bash tools/ci/projection_state_version_guard.sh
bash tools/ci/projection_state_mirror_current_state_guard.sh
bash tools/ci/projection_route_mapping_guard.sh
git diff --check
dotnet build aevatar.slnx --nologo
```

Also started the GAgent service locally with in-memory projection/document/graph providers and auth disabled, then exercised related scope service endpoints. The RoleGAgent draft-run failure path committed `RoleChatSessionCompletedEvent` and wrote `GAgentRunTerminalReadModel` successfully.

## Notes

Full build succeeds with existing warnings from NuGet source configuration, vulnerability advisories, obsolete APIs, complexity, and nullability warnings.


src/platform/Aevatar.GAgentService.Application/ScopeGAgents/GAgentDraftRunInteraction.cs
```diff
@@ -416,12 +435,44 @@ public Task EmitAsync(
...
     {
         ArgumentNullException.ThrowIfNull(receipt);
-        _ = ct;
-        return Task.FromResult(CommandDurableCompletionObservation<GAgentDraftRunCompletionStatus>.Incomplete);
+
+        try
+        {
+            var snapshot = await _queryPort.GetByCorrelationIdAsync(receipt.ActorId, receipt.CorrelationId, ct);
+            if (snapshot == null && !string.IsNullOrWhiteSpace(receipt.SessionId))
+                snapshot = await _queryPort.GetBySessionIdAsync(receipt.ActorId, receipt.SessionId, ct);
```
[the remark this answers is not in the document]
louis4li: Fixed in aee3b139. The durable resolvers now only accept correlation or session fallback snapshots when the snapshot matches the active receipt actor/correlation and the expected interaction kind (draft-run vs approval), so reused sessions cannot resolve from stale terminal documents. Added unit coverage for stale correlation, wrong interaction kind, and valid fallback for both draft-run and approval.

src/platform/Aevatar.GAgentService.Projection/Projectors/GAgentRunTerminalProjector.cs
```diff
@@ -0,0 +1,136 @@
...
+        var (status, reasonCode, reasonMessage) = ResolveTerminal(completed);
+        var document = new GAgentRunTerminalReadModel
+        {
+            Id = documentId,
+            ActorId = context.RootActorId,
+            StateVersion = stateVersion,
+            LastEventId = eventId,
+            SessionId = completed.SessionId,
+            CorrelationId = correlationId,
+            InteractionKind = (int)ResolveInteractionKind(completed.SessionId, correlationId),
```
eanzhao: **[Consensus: 3 models]** severity=major, category=bug

`ResolveInteractionKind` infers `Approval` whenever `sessionId != correlationId`, but `GAgentDraftRunCommandTargetBinder.ResolveSessionId` returns `command.SessionId.Trim()` when the caller supplies one — so any draft-run with an explicit `SessionId` distinct from `CorrelationId` will be written into the readmodel as `Approval`. This makes `InteractionKind` unreliable for any downstream consumer (filtering, dashboards, future typed terminal-event mapping).

Fix: carry the interaction kind through the projection scope/context (e.g. add `InteractionKind` to `GAgentRunTerminalProjectionContext` from the activation site at `GAgentDraftRunInteraction.cs` / `GAgentApprovalInteraction.cs`) and read it from the context here, instead of inferring from session/correlation equality.

<details>

- **codex**: `ResolveInteractionKind` infers approval whenever `SessionId != CorrelationId`, but draft-run supports an explicit `SessionId`, so those draft-run terminal snapshots will be mislabeled as `Approval`. Carry the interaction kind through the activation context/port or store separate projection kinds instead of deriving it from the session/correlation equality.
- **gemini**: The `ResolveInteractionKind` logic assumes `DraftRun` always has `SessionId == CorrelationId`. If a user provides a custom `SessionId` in the `GAgentDraftRunCommand`, this will incorrectly label the interaction as `Approval`. While non-critical for completion logic, it makes the `InteractionKind` field in the read model potentially misleading.
- **kimi**: `ResolveInteractionKind` uses the heuristic `sessionId == correlationId` to distinguish DraftRun from Approval. A draft-run with an explicit `SessionId` (different from `CorrelationId`) would be misclassified as Approval. Consider documenting this assumption or deriving the kind from explicit event metadata when a typed terminal event is introduced.
</details>

src/platform/Aevatar.GAgentService.Abstractions/ScopeGAgents/GAgentRunTerminalModels.cs
```diff
@@ -0,0 +1,49 @@
...
+        string correlationId,
+        CancellationToken ct = default);
+
+    Task<GAgentRunTerminalSnapshot?> GetBySessionIdAsync(
+        string actorId,
+        string sessionId,
+        CancellationToken ct = default);
+}
+
+public interface IGAgentRunTerminalProjectionPort
```
eanzhao, line 52: **[Consensus: 2 models]** severity=major, category=arch

`IGAgentRunTerminalProjectionPort` only exposes `EnsureProjectionAsync` (correlation-scoped) with no corresponding `ReleaseProjectionAsync` / lease return. Each draft-run / approval command currently activates a new correlation-keyed durable materialization scope (`GAgentApprovalInteraction.cs:181`, `GAgentDraftRunInteraction.cs:236`) and never releases it. Since correlation is per-command, these scopes accumulate until Orleans deactivation — defeating the point of the new `IProjectionSessionScopedMaterializationContext` release-aware design described in the PR summary.

Fix: return a lease (analogous to `IGAgentDraftRunProjectionLease`) from `EnsureProjectionAsync`, and have the interaction cleanup path (`GAgentApprovalCommandTarget.ReleaseAsync` / draft-run equivalent) release it. Alternatively, make this an actor-scoped (not correlation-scoped) materialization.

<details>

- **codex** (GAgentRunTerminalProjectionPort.cs:33): This activates a durable materialization scope keyed by correlation id, then discards the lease and exposes no release path. Because draft-run/approval create a new correlation per command, these projection actors/subscriptions can accumulate without cleanup; return a terminal projection lease from the port and release it from the interaction cleanup path, or make this an actor-scoped materialization instead of a per-correlation scope.
- **gemini** (GAgentRunTerminalModels.cs:43): The interface is missing a corresponding `ReleaseProjectionAsync` or `ReleaseIfIdleAsync` method. Given the PR summary explicitly mentions making durable materialization 'release session-aware', the port should expose this capability to allow the interaction service to clean up command-scoped materializers.
- **gemini** (GAgentApprovalInteraction.cs:157): The terminal projection is ensured here but never released. Since these materialization actors are now partitioned by `CorrelationId` (one per command), they will leak until Orleans deactivates them. The `GAgentApprovalCommandTarget` should hold a reference to the `IGAgentRunTerminalProjectionPort` and call a (yet to be exposed) release method in its `ReleaseAsync` implementation, similar to how the live `_projectionPort` is handled.
</details>

src/platform/Aevatar.GAgentService.Application/ScopeGAgents/GAgentDraftRunInteraction.cs
eanzhao: **[codex]** severity=major, category=bug

The `if (snapshot == null && !string.IsNullOrWhiteSpace(receipt.SessionId))` fallback accepts whatever `GetBySessionIdAsync` returns without checking that the returned snapshot's `CorrelationId` matches `receipt.CorrelationId`. If a session/thread is reused (which is the whole point of `command.SessionId`) and the correlation document is temporarily missing or evicted, this can complete the current command from an older terminal fact — silently delivering stale completion status.

Same pattern exists in `GAgentApprovalDurableCompletionResolver` (`GAgentApprovalInteraction.cs:343-345`).

Fix: only accept the session-fallback snapshot when its `CorrelationId == receipt.CorrelationId`, otherwise return `Incomplete`:
```csharp
if (snapshot != null &&
    !string.Equals(snapshot.CorrelationId, receipt.CorrelationId, StringComparison.Ordinal))
    snapshot = null;
```
Apply in both resolvers.

src/Aevatar.AI.Core/RoleGAgent.cs
```diff
@@ -151,6 +159,12 @@ public async Task HandleToolApprovalDecision(ToolApprovalDecisionEvent evt)
         else
```
eanzhao, line 152: **[kimi]** severity=major, category=bug

The approval-denied `else` branch calls `PersistApprovalTerminalFailureAsync` without a `try / catch`. If this throws after `ClearPendingApprovalEvent` is already persisted, the actor event fails and retries; on retry `State.PendingApproval` is already null so the method returns early, permanently losing the terminal failure fact — exactly what the durable readmodel needs.

The same gap exists in two more call sites in this file:
- `HandleToolApprovalTimeout` no-remote-handler branch (around line 195)
- `HandleToolApprovalTimeout` remote-failed/denied/timed-out branch (around line 249)

Only the exception path at lines 143-150 wraps it in `try { … } catch { /* best effort */ }`.

Fix: either wrap each call site in best-effort try/catch (matching the pattern at lines 143-150), or move the try/catch inside `PersistApprovalTerminalFailureAsync` itself so every caller is safe.

test/Aevatar.GAgentService.Tests/Projection/GAgentRunTerminalProjectorTests.cs
```diff
@@ -0,0 +1,227 @@
+using Aevatar.AI.Abstractions;
```
eanzhao, line 1: **[mimo-v2.5-pro]** severity=major, category=test

Missing test for monotonic version guard. CLAUDE.md projection rules require: "旧 StateVersion 不覆盖新 readmodel". Add a test that sends two committed events with the same `actorId + correlationId` but a descending `stateVersion`, then asserts the readmodel retains the higher version (or that the projector / `IProjectionWriteDispatcher` rejects the stale write).

Without this test, a future change to `_writeDispatcher.UpsertAsync` semantics (or a parallel write path) could regress the monotonic guarantee without anyone noticing — the rest of the test suite only exercises forward-version cases.

src/Aevatar.AI.Core/RoleGAgent.cs
```diff
@@ -140,6 +140,14 @@ public async Task HandleToolApprovalDecision(ToolApprovalDecisionEvent evt)
                 // Still clear pending so we don't get stuck
                 try { await PersistDomainEventAsync(new ClearPendingApprovalEvent { RequestId = pending.RequestId }); }
                 catch { /* best effort */ }
+                try
+                {
```
eanzhao: **[gemini]** severity=nit, category=other

This `try { … } catch { /* best effort */ }` swallows persistence failures silently. If `PersistApprovalTerminalFailureAsync` fails here, the durable resolver will stay `Incomplete` indefinitely with no operator signal. Consider logging at warning level inside the catch (include `pending.SessionId`, `pending.RequestId`, the reason code, and `ex.Message`) so this is observable.

Same applies to the symmetric catch the kimi comment proposes adding at the denied / timeout sites — when those wraps are added, give them logging too.

src/platform/Aevatar.GAgentService.Projection/Projectors/GAgentRunTerminalProjector.cs
louis4li: Fixed in 16eccf97. The terminal projection now carries interaction kind explicitly through the activation context using separate projection kinds for draft-run and approval, and the projector writes InteractionKind from context instead of deriving it from session/correlation equality. Added regression coverage for draft-run with explicit SessionId != CorrelationId.

src/platform/Aevatar.GAgentService.Abstractions/ScopeGAgents/GAgentRunTerminalModels.cs
louis4li: Fixed in 16eccf97. IGAgentRunTerminalProjectionPort now returns an IGAgentRunTerminalProjectionLease and exposes ReleaseProjectionAsync. Draft-run and approval targets retain the terminal lease and release it during cleanup, including activation-failure cleanup after terminal materialization succeeds.

src/platform/Aevatar.GAgentService.Application/ScopeGAgents/GAgentDraftRunInteraction.cs
louis4li: Fixed in aee3b139. The fallback path now validates actor, active correlation, and expected interaction kind before accepting a session lookup result, so reused sessions cannot resolve the current command from stale terminal documents.

src/Aevatar.AI.Core/RoleGAgent.cs
louis4li: Fixed in 16eccf97. Approval denied, no-remote timeout, and remote denied/failed/timed-out paths now persist the terminal failure fact before clearing PendingApproval, preserving the durable completion source if terminal persistence fails.

test/Aevatar.GAgentService.Tests/Projection/GAgentRunTerminalProjectorTests.cs
louis4li: Fixed in 16eccf97. Added a monotonic StateVersion regression test for GAgentRunTerminalProjector using the projection write evaluator semantics so an older committed event cannot overwrite a newer terminal readmodel.

src/Aevatar.AI.Core/RoleGAgent.cs
louis4li: Fixed in 16eccf97. The approval continuation failure path now logs terminal failure persistence errors with request id, session id, and reason code, and logs pending-clear failures after the terminal fact has been persisted.

src/platform/Aevatar.GAgentService.Application/ScopeGAgents/GAgentDraftRunInteraction.cs
```diff
@@ -117,6 +129,20 @@ await _projectionPort.DetachReleaseAndDisposeAsync(
             }
         }
 
+        var terminalProjectionLease = TerminalProjectionLease;
+        if (terminalProjectionLease != null)
+        {
+            try
+            {
+                await _terminalProjectionPort.ReleaseProjectionAsync(terminalProjectionLease, ct);
```
jason-aelf, line 157: After a draft run enters pending approval, the live stream has already emitted `TextMessageEnd`, so interaction cleanup treats this interaction as `observedCompleted=true` and releases the `TerminalProjectionLease`. However, the `RoleChatSessionCompletedEvent` failure fact for approval local timeout, remote denied, or remote timeout is a later actor event and can be committed after the projection scope has already been `Released`. At that point, `ProjectionScopeGAgentBase.HandleObservedEnvelopeAsync` returns immediately, so `GAgentRunTerminalProjector` never materializes a `GAgentRunTerminalReadModel`. The actor has committed the terminal failure fact, but the durable read model layer cannot observe it, and recovery after a missed live terminal event can read an incomplete or stale state.

src/platform/Aevatar.GAgentService.Projection/Projectors/GAgentRunTerminalProjector.cs
```diff
@@ -0,0 +1,128 @@
...
+                }
+            }
+
+            return (
+                GAgentRunTerminalStatus.Failed,
+                ReasonCodeLegacyLlmError,
+                legacyMessage);
+        }
+
+        if (content.StartsWith(LegacyLlmFailedPrefix, StringComparison.Ordinal))
```
jason-aelf, line 112: The normal provider / LLM exception path in `RoleGAgent` writes `LLM request failed [tools=...]: ...` when `useWorkflowFailureMarker=false`, but `GAgentRunTerminalProjector.ResolveTerminal` only recognizes `[[AEVATAR_LLM_ERROR]]` and the exact prefix `LLM request failed:`. Failure content that includes `[tools=...]` falls through to the default branch and is materialized as `TextMessageCompleted`, so the durable resolver later recovers a real failure as completed. This path persists provider / LLM exceptions as successful terminal states.

src/platform/Aevatar.GAgentService.Application/ScopeGAgents/GAgentDraftRunInteraction.cs
louis4li: Fixed in 2898083b. Draft-run cleanup now keeps the terminal projection active only after the live AGUI stream has actually observed a TOOL_APPROVAL_REQUEST, so the later committed RoleChatSessionCompletedEvent for approval timeout/denied/remote failure can still be materialized into GAgentRunTerminalReadModel. Ordinary successful text completions without a pending approval still release the terminal projection. Added regression coverage for both the pending-approval retention path and the normal text-completion release path.

Verified with:
- dotnet test test/Aevatar.GAgentService.Tests/Aevatar.GAgentService.Tests.csproj --nologo --filter "GAgentDraftRunInteraction|GAgentApprovalInteraction|GAgentRunTerminal"
- dotnet test test/Aevatar.AI.Tests/Aevatar.AI.Tests.csproj --nologo --filter "RoleGAgent"

src/platform/Aevatar.GAgentService.Projection/Projectors/GAgentRunTerminalProjector.cs
louis4li: Fixed in 2898083b. GAgentRunTerminalProjector now recognizes the broader legacy provider-failure prefix "LLM request failed", including the tool-aware form "LLM request failed [tools=...]: ...", and materializes it as Failed with reason_code=legacy_llm_error instead of TextMessageCompleted. Added a regression test covering the tool-aware failure content.

Verified with:
- dotnet test test/Aevatar.GAgentService.Tests/Aevatar.GAgentService.Tests.csproj --nologo --filter "GAgentDraftRunInteraction|GAgentApprovalInteraction|GAgentRunTerminal"
- dotnet test test/Aevatar.AI.Tests/Aevatar.AI.Tests.csproj --nologo --filter "RoleGAgent"

jason-aelf (approved):

PR #680, opened 2026-05-19
Fix scope binding readiness observation

louis4li:
## Summary
- Remove hidden read-model polling from scope binding command upsert.
- Add explicit scope binding readiness query for service catalog and serving-set observation.
- Make Studio binding runs wait for the activated revision/deployment before dispatching success, with timeout/failure continuations.

## Test plan
- [x] dotnet test test/Aevatar.GAgentService.Tests/Aevatar.GAgentService.Tests.csproj --no-restore --nologo --filter "ScopeBindingReadinessQueryServiceTests|ScopeBindingCommandApplicationServiceTests|ApplicationServiceGuardTests"
- [x] dotnet test test/Aevatar.Studio.Tests/Aevatar.Studio.Tests.csproj --no-restore --nologo --filter "ScopeBindingStudioMemberPlatformBindingCommandServiceTests"
- [x] dotnet test test/Aevatar.GAgentService.Integration.Tests/Aevatar.GAgentService.Integration.Tests.csproj --no-restore --nologo --filter "GAgentServiceHostingServiceCollectionExtensionsTests|ScopeServiceEndpointsTests"
- [x] bash tools/ci/test_stability_guards.sh
- [x] git diff --check


src/Aevatar.Studio.Projection/CommandServices/ScopeBindingStudioMemberPlatformBindingCommandService.cs
```diff
@@ -133,6 +157,80 @@ await DispatchAsync(
...
+                    dispatchEx,
+                    "StudioMember platform binding readiness-failure continuation dispatch failed. bindingRunId={BindingRunId} platformBindingCommandId={CommandId}",
+                    request.BindingRunId,
+                    commandId);
+            }
+
+            return;
+        }
+
+        if (!readiness.InvokeReady)
```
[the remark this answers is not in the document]
louis4li: Fixed in 9198ca4b. Readiness timeout now only logs a warning and returns without dispatching a terminal continuation, so the binding run remains PlatformBindingPending and the existing watchdog/recovery path can retry. Readiness query exceptions still dispatch STUDIO_MEMBER_PLATFORM_BINDING_READINESS_FAILED because those are observation failures rather than normal projection lag.

src/platform/Aevatar.GAgentService.Application/Bindings/ScopeBindingReadinessQueryService.cs
```diff
@@ -0,0 +1,130 @@
...
+            RevisionId: eligibleTarget.RevisionId,
+            DeploymentId: eligibleTarget.DeploymentId,
+            ObservedAtUtc: observedAtUtc);
+    }
+
+    private static bool IsServiceCatalogTargetVisible(
+        ServiceCatalogSnapshot service,
+        string? expectedRevisionId,
+        string? expectedDeploymentId) =>
+        (string.IsNullOrWhiteSpace(expectedRevisionId) ||
```
eanzhao: **[codex]** severity=major, category=bug

This catalog-active check can make Studio binding runs stay pending forever when rebinding to a revision that is already active in `ServiceDeploymentManagerGAgent` but is not the catalog's current `ActiveServingRevisionId`. `HandleActivateAsync` returns early for an existing active revision and only dispatches serving targets; it does not emit `ServiceDeploymentActivatedEvent`, so `ServiceCatalogProjector` never updates `ActiveServingRevisionId`/`DeploymentId` to the requested revision even though the serving set can be switched and invocation can work. Either make the activation idempotent path republish the catalog activation fact, or relax/reshape this readiness predicate for existing-active deployments and add a test for binding back from rev-B to previously-active rev-A.

---
_Verified against `dev` HEAD: `ServiceDeploymentManagerGAgent.HandleActivateAsync` lines 64-74 early-return after `DispatchResolvedServingTargetsAsync` **without** `PersistDomainEventAsync(ServiceDeploymentActivatedEvent)`; `ServiceCatalogProjector` lines 82-93 only update `ActiveServingRevisionId`/`DeploymentId` from that event. The stale-catalog hang is real._

test/Aevatar.GAgentService.Tests/Application/ScopeBindingCommandApplicationServiceTests.cs
```diff
@@ -1245,6 +1245,36 @@ await service.UpsertAsync(new ScopeBindingUpsertRequest(
...
+            DefaultOptions);
+
+        await service.UpsertAsync(new ScopeBindingUpsertRequest(
+            ScopeId,
+            ScopeBindingImplementationKind.Workflow,
+            Workflow: new ScopeBindingWorkflowSpec([
+                "name: main\nsteps:\n  - run: echo hello",
+            ])));
+
+        typeof(ScopeBindingCommandApplicationService)
```
eanzhao, line 1416: **[v4-pro]** severity=minor, category=test

`UpsertAsync_ShouldNotPollServingReadiness` verifies that `IServiceServingQueryPort` is absent from the constructor via reflection, but does NOT behaviorally verify that `UpsertAsync` completes without calling serving-readiness methods. The design doc (section 11.1) planned a `RecordingServiceServingQueryPort` that records `GetServiceServingSetAsync` call count. Consider replacing the structural assertion with a behavioral one (inject a recording fake and assert zero calls) so that a future change adding `IServiceServingQueryPort` via property injection or service-locator is also caught.

test/Aevatar.Studio.Tests/ScopeBindingStudioMemberPlatformBindingCommandServiceTests.cs
```diff
@@ -223,14 +206,113 @@ await service.ExecuteAsync(
     }
 
     [Fact]
-    public async Task ExecuteAsync_WhenImplementationPayloadMissing_ShouldDispatchFailedContinuation()
+    public async Task ExecuteAsync_WhenReadinessBecomesReady_ShouldWaitBeforeDispatchingSucceededContinuation()
+    {
+        var readinessPort = new RecordingReadinessQueryPort([
+            NotReadySnapshot(),
+            ReadySnapshot(),
```
eanzhao, line 259: **[mimo-v2.5-pro]** severity=minor, category=test

`WaitAsync(TimeSpan.FromMilliseconds(100))` is tight for CI — on a slow runner this could flake. Consider using `Task.WhenAny` with a longer timeout or a `CancellationTokenSource` with a 2-5s deadline to assert non-completion.

test/Aevatar.Studio.Tests/ScopeBindingStudioMemberPlatformBindingCommandServiceTests.cs
```diff
@@ -223,14 +206,113 @@ await service.ExecuteAsync(
...
         var dispatchPort = new RecordingDispatchPort();
-        var service = new ScopeBindingStudioMemberPlatformBindingCommandService(
+        var now = DateTimeOffset.Parse("2026-05-19T00:00:00+00:00");
+        var service = CreateService(
             scopeBindingPort,
             dispatchPort,
-            NullLogger<ScopeBindingStudioMemberPlatformBindingCommandService>.Instance);
+            readinessPort,
+            delayAsync: (_, _) =>
+            {
```
eanzhao, line 294: **[mimo-v2.5-pro]** severity=minor, category=test

Missing test for the nested catch path: when readiness observation throws AND the subsequent `StudioMemberPlatformBindingFailed` dispatch also throws. The code handles this (lines 190-200 in the new adapter), but there is no test covering it. Add a test with `Failure` set on both readiness port and dispatch port to verify the adapter doesn't crash.

test/Aevatar.Studio.Tests/ScopeBindingStudioMemberPlatformBindingCommandServiceTests.cs
```diff
@@ -461,7 +577,37 @@ private static ScopeBindingUpsertResult BuildGAgentResult(ScopeBindingUpsertRequ
...
+        {
+            _snapshots = new Queue<ScopeBindingReadinessSnapshot>(snapshots);
+        }
+
+        public List<ScopeBindingReadinessRequest> Requests { get; } = [];
+
+        public Exception? Failure { get; init; }
+
+        public static RecordingReadinessQueryPort Ready() => new([ReadySnapshot()]);
+
```
eanzhao, line 677: **[v4-pro]** severity=minor, category=test

`RecordingReadinessQueryPort` uses `_snapshots.Peek()` when `_snapshots.Count <= 1`, which throws `InvalidOperationException` on an empty queue (`new RecordingReadinessQueryPort([])`). This can cause confusing test failures if a caller constructs it with an empty enumerable. Consider adding a guard or clarifying in the constructor that at least one snapshot is required.

test/Aevatar.GAgentService.Tests/Application/ScopeBindingReadinessQueryServiceTests.cs
```diff
@@ -0,0 +1,346 @@
...
+        snapshot.Status.Should().Be(ScopeBindingReadinessStatus.Ready);
+        snapshot.ServiceCatalogVisible.Should().BeTrue();
+        snapshot.ServingSetVisible.Should().BeTrue();
+        snapshot.EligibleServingTargetVisible.Should().BeTrue();
+        snapshot.InvokeReady.Should().BeTrue();
+        snapshot.RevisionId.Should().Be("rev-ready");
+        snapshot.DeploymentId.Should().Be("deployment-rev-ready");
+    }
+
+    [Fact]
```
eanzhao, line 273: **[mimo-v2.5-pro]** severity=nit, category=test

The `GetReadinessAsync_ShouldBuildServiceIdentityFromNormalizedRequestAndAppId` test asserts the full `ServiceIdentity` equivalence but doesn't verify the service actually queries both lifecycle and serving ports with that identity. Consider adding `servingPort.LastIdentity.Should().BeEquivalentTo(lifecyclePort.LastIdentity)` to catch identity drift between the two query calls.

src/platform/Aevatar.GAgentService.Application/Bindings/ScopeBindingReadinessQueryService.cs
```diff
@@ -0,0 +1,130 @@
...
+        }
+
+        return new ScopeBindingReadinessSnapshot(
+            normalizedScopeId,
+            normalizedServiceId,
+            ScopeBindingReadinessStatus.Ready,
+            ServiceCatalogVisible: true,
+            ServingSetVisible: true,
+            EligibleServingTargetVisible: true,
+            InvokeReady: true,
```
jason-aelf, line 127: `InvokeReady` currently does not observe the read model that invoke resolution actually uses.

`GetReadinessAsync` returns `InvokeReady = true` after service catalog and serving set show an eligible active target, but `ServiceInvocationResolutionService.ResolveAsync` first reads `IServiceTrafficViewQueryReader` and only falls back to serving set when the endpoint view is missing or has no targets. Since `ServiceServingSetProjector` and `ServiceTrafficViewProjector` are independent materializers, serving set can become visible before traffic view has caught up.

That leaves a bind-then-invoke race:

1. the new revision/deployment is visible in catalog + serving set;
2. readiness returns `InvokeReady = true`;
3. Studio dispatches `StudioMemberPlatformBindingSucceeded`;
4. an immediate invoke still resolves through the old traffic view target.

So the current `InvokeReady` contract is stronger than the state it actually observes.

src/Aevatar.Studio.Projection/CommandServices/ScopeBindingStudioMemberPlatformBindingCommandService.cs
```diff
@@ -133,6 +158,59 @@ await DispatchAsync(
...
```
jason-aelf, line 203: Readiness timeout recovery can replay a non-idempotent explicit Workflow/GAgent revision.

When readiness times out, this path returns without dispatching success or failure, leaving the run pending for watchdog recovery. The watchdog then re-drives the same platform command from durable actor state.

That is not idempotent for explicit Workflow/GAgent `revision_id` requests. For explicit revisions, `ResolveReplayRevisionId(...)` returns `null`, so Workflow and GAgent binding requests call `UpsertAsync` with `AllowExistingRevisionReplay = false`.

If the first execution already created/published/activated the revision and only readiness observation timed out, the recovery execution calls `UpsertAsync` again, sees the existing revision, and `ShouldCreateRevisionAsync` throws `Revision '<id>' already exists`.

That turns projection/readiness lag into a terminal binding failure even though the platform mutation may have succeeded.

louis4li:
@codex review

src/platform/Aevatar.GAgentService.Application/Bindings/ScopeBindingReadinessQueryService.cs
louis4li: Fixed by relaxing the catalog activation predicate: readiness no longer gates on catalog ActiveServingRevisionId/DeploymentId, and the test covers serving-set readiness when the catalog target fields differ.

test/Aevatar.GAgentService.Tests/Application/ScopeBindingCommandApplicationServiceTests.cs
louis4li: No code change here: ScopeBindingCommandApplicationService still has no IServiceServingQueryPort injection point, and this test also scans the production source for IServiceServingQueryPort/GetServiceServingSetAsync, which catches the service-locator/direct-call case the behavioral fake would target.

test/Aevatar.Studio.Tests/ScopeBindingStudioMemberPlatformBindingCommandServiceTests.cs
louis4li: Fixed: the tight 100ms non-completion wait was removed; the current timeout test waits on an explicit observation TaskCompletionSource with a 5s deadline.

test/Aevatar.Studio.Tests/ScopeBindingStudioMemberPlatformBindingCommandServiceTests.cs
louis4li: Fixed: added ExecuteAsync_WhenReadinessQueryFailsAndFailureContinuationDispatchFails_ShouldNotThrow to cover readiness failure plus failed failure-continuation dispatch.

test/Aevatar.Studio.Tests/ScopeBindingStudioMemberPlatformBindingCommandServiceTests.cs
louis4li: Fixed: RecordingReadinessQueryPort now guards empty snapshot sequences in the constructor with an ArgumentException.

test/Aevatar.GAgentService.Tests/Application/ScopeBindingReadinessQueryServiceTests.cs
louis4li: Fixed: the identity test now also asserts servingPort.LastIdentity is equivalent to lifecyclePort.LastIdentity.

src/platform/Aevatar.GAgentService.Application/Bindings/ScopeBindingReadinessQueryService.cs
louis4li: Fixed: readiness now checks the traffic view for the relevant endpoint ids and expected revision/deployment before returning InvokeReady=true; stale traffic targets return TrafficViewTargetMissing.

src/Aevatar.Studio.Projection/CommandServices/ScopeBindingStudioMemberPlatformBindingCommandService.cs
louis4li: Fixed in 35070482: explicit Workflow/GAgent requests now pass ReplayRevisionId=revisionId so watchdog recovery can replay them. Existing prepared revisions are still protected by artifact-hash validation in ScopeBindingCommandApplicationService.

src/platform/Aevatar.GAgentService.Application/Bindings/ScopeBindingCommandApplicationService.cs
```diff
@@ -155,28 +155,47 @@ private async Task<bool> ShouldCreateRevisionAsync(
...
-        var expectedArtifactHash = await ComputeScriptingArtifactHashAsync(revisionSpec, ct);
+        if (!request.AllowExistingRevisionReplay ||
+            !string.Equals(request.ReplayRevisionId, revisionId, StringComparison.Ordinal))
+        {
+            throw new InvalidOperationException(
+                $"Revision '{revisionId}' already exists for service '{ServiceKeys.Build(identity)}'.");
+        }
+
+        var expectedArtifactHash = ComputeNonScriptingArtifactHash(revisionSpec);
         if (!string.Equals(existingRevision.ArtifactHash, expectedArtifactHash, StringComparison.Ordinal))
```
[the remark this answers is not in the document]
louis4li: Fixed: replay now allows unprepared revisions with an empty ArtifactHash to continue to PrepareRevisionAsync, while prepared revisions still compare artifact hashes before replay.

src/platform/Aevatar.GAgentService.Application/Bindings/ScopeBindingReadinessQueryService.cs
```diff
@@ -0,0 +1,195 @@
...
+                ServiceCatalogVisible: true,
+                ServingSetVisible: false,
+                EligibleServingTargetVisible: false,
+                InvokeReady: false,
+                ObservedAtUtc: observedAtUtc);
+        }
+
+        var serviceEndpointIds = expectedEndpointIds.Count > 0
+            ? expectedEndpointIds
+            : GetServiceEndpointIds(service);
```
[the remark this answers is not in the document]
louis4li: Fixed: when ExpectedEndpointIds is provided, readiness now also checks the service catalog contains those endpoint ids before returning Ready; stale catalog endpoints return ServiceCatalogTargetMissing.

src/Aevatar.Studio.Projection/CommandServices/ScopeBindingStudioMemberPlatformBindingCommandService.cs
```diff
@@ -163,6 +268,68 @@ await DispatchAsync(
...
+
+    private static IReadOnlyList<string> BuildExpectedEndpointIds(StudioMemberPlatformBindingStartRequested request) =>
+        request.Request.ImplementationCase switch
+        {
+            StudioMemberBindingRequest.ImplementationOneofCase.Gagent => request.Request.Gagent.Endpoints
+                .Select(endpoint => endpoint.EndpointId?.Trim())
+                .Where(endpointId => !string.IsNullOrWhiteSpace(endpointId))
+                .Distinct(StringComparer.Ordinal)
+                .ToArray(),
+            _ => ["chat"],
```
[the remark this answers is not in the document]
louis4li: Fixed: script binding readiness now uses endpoint ids returned from BuildScriptEndpointSpecs via ScopeBindingUpsertResult.Script.EndpointIds instead of hard-coding chat.

PR #699, opened 2026-05-19
feat(ingress): ingress layer v1 — chat routing config actor + resolver + 4 text entries + /ws/voice (#691–#696)

eanzhao:
## Summary

Lands **Ingress layer v1** ([milestone #20](https://github.com/aevatarAI/aevatar/milestone/20)) as one PR. Closes #672, #674, and the per-phase trackers #691 #692 #693 #694 #695 #696.

Originally Phase 0 only (proto + ADR). After review feedback we collapsed the entire 6-PR train (#703 #704 #705 #709 #710) onto `feature/router` so the ingress layer ships in a single reviewable diff against `dev`.

## What this PR contains

- **Phase 0 — proto + ADR**. `src/Aevatar.ChatRouting.Abstractions` (`chat_route_policy.proto`, typed enums) and [ADR-0024](docs/adr/0024-chat-route-policy.md) — the form is "config actor + stateless boundary resolver + readmodel", explicitly **not** a router actor (rejected per #608 Harness review).
- **Phase 1 — config actor + projection**. `agents/Aevatar.GAgents.ChatRouting`:
  - `ChatRoutePolicyGAgent` (per-scope config aggregate; validates `default_target`, sorts rules by priority).
  - `ChatRoutePolicyCurrentStateDocument` + `ChatRoutePolicyCurrentStateProjector` (overwrite-materializes committed state, version from the committed `StateEvent`).
  - `AddChatRoutingAgents` wires materialization runtime + materializer + InMemory document store; codex caught the missing materializer registration during review and the fix is included here.
- **Phase 2 — stateless resolver + query port + env fallback**. `src/Aevatar.ChatRouting.Core`:
  - `ChatRouteResolver` — pure `Resolve(snapshot, input)`, priority-ordered rules, records `matched_rule_id`, falls through to default and env fallback. Hardened during review to skip rules with empty actions (proto3 message field can be null).
  - `IChatRoutePolicyQueryPort` + `ChatRoutePolicyQueryPort` — readmodel-only lookup, `null` snapshot when no rows.
  - `IChatRouteFallbackProvider` + env-driven `EnvChatRouteFallbackProvider`.
  - `AddChatRoutingCore()` DI, wired in the Mainnet host.
- **Phase 3 — text ingress entries**. Resolver wired into the four text paths with no new actor hop:
  - NyxIdChat create (`NyxIdChatEndpoints`) — `ForwardToGAgent` overrides the target; `Reject` → 403. Review feedback fix: rollback only destroys the actor we created locally, never a forwarded routed target.
  - Responses + Messages endpoints — resolve after caller-scope lookup; `ForwardToModel` rewrites the model.
  - Relay `ConversationGAgent` — resolves inside the turn before runner admission; decision rides the typed `NeedsLlmReplyEvent.target_ref`; `reply_token` stays actor-private. Review feedback fix: sender NyxID is now resolved at relay ingress (`INyxIdCurrentUserResolver.ResolveCurrentUserIdAsync`) and carried on a new typed `TransportExtras.nyx_sender_user_id`, so per-user channel policies actually match.
- **Phase 4 — policy-aware `/ws/voice` + dev bypass**. `PolicyAwareVoiceEndpoints`:
  - `/ws/voice` resolves through query port + resolver with typed `VoiceInput`; pre-upgrade `Reject` → 403, `ForwardToModel` → 501, `ForwardToGAgent` → attach-permission + presence-session resolution. Review feedback fix: returns 503 (retryable) instead of 404 when the routed voice GAgent's module is still warming up, matching the dev bypass.
  - First `MapVoicePresenceWebSocket` mount on the Mainnet host: `/ws/voice/{actorId}` kept behind the new `voice-dev` authorization policy (requires `voice:bypass` scope or admin).
- **Phase 5 — docs closeout**. Adds the **"router = config actor + boundary resolver"** entry to `docs/canon/architecture-vocabulary.md`, citing ADR-0024.

## Architecture invariants honored

| Invariant | How |
|---|---|
| No router actor | `ChatRoutePolicyGAgent` is config-only; resolver is a library function |
| Decision never persisted | `ChatRouteDecision` is transient; not in any state proto |
| No metadata bag | Every routing-influencing field is strong-typed (CLAUDE.md decision tree step 1) |
| Foundation does not depend on ChatRouting | `Aevatar.Foundation.VoicePresence` reverse-dep grep clean |
| `/ws/voice` not `/ws/chat` | ADR-0024 D4 spells out the binary-vs-JSON reason |
| External repos untouched | No NyxID / chrono-* surface required |

## Codex review findings — all addressed in-PR

| PR | Severity | Finding | Fix commit |
|---|---|---|---|
| #703 | P1 | `AddChatRoutingAgents` didn't register the materializer — readmodel never populated | a5352223 |
| #704 | P2 | Resolver NREed on matching rule with empty/null `action` | 7fa8de81 |
| #705 | P1 | Create rollback could destroy a pre-existing forwarded target actor | 39e430a8 |
| #705 | P2 | Relay `OwnerScope.NyxUserId` came from `registrationScopeId` — per-user channel policies always missed | 39e430a8 |
| #709 | P2 | `/ws/voice` returned 404 (not retryable) for cold voice module | 4934a9b7 |

## Test plan

- [x] `dotnet build aevatar.slnx` — green
- [x] `bash tools/docs/lint.sh` — green, ADR-0024 included
- [x] `bash tools/ci/projection_state_version_guard.sh` — green
- [x] `bash tools/ci/projection_route_mapping_guard.sh` — green
- [x] `bash tools/ci/test_stability_guards.sh` — green
- [x] Per-phase test projects green: `Aevatar.GAgents.ChatRouting.Tests` (14), `Aevatar.ChatRouting.Core.Tests` (13), `Aevatar.AI.Tests` (559, includes 4 new relay regression tests), `Aevatar.GAgents.Channel.Protocol.Tests` (138), `Aevatar.ChatRouting.Voice.Integration.Tests` (9)
- [x] `architecture_guards.sh` passes everything except the pre-existing `playground_asset_drift_guard` (unrelated; zero frontend assets touched in this milestone)
- [ ] Reviewer confirms no new guard failures introduced

## Related

- Milestone: [Ingress layer v1](https://github.com/aevatarAI/aevatar/milestone/20)
- Parent designs: #672 #674
- Phase trackers (closed by this PR): #691 #692 #693 #694 #695 #696
- Boundaries: #568 #596 #608 #560 (ADR-0024 §"Boundaries with adjacent issues")
- Collapsed PRs (merged into `feature/router`): #703 #704 #705 #709 #710


src/Aevatar.ChatRouting.Abstractions/chat_route_policy.proto
```diff
@@ -0,0 +1,221 @@
...
+  string target_actor_id = 1;
+}
+
+message ChatRouteAction {
+  oneof action {
+    ForwardToModel forward_to_model = 1;
+    ForwardToGAgent forward_to_gagent = 2;
+    ForwardToWorkflow forward_to_workflow = 3;
+    Reject reject = 4;
+    Bypass bypass = 5;
```
eanzhao: **[codex]** severity=major, category=arch

`Bypass` is part of `ChatRouteAction`, and the same type is used by `ChatRoutePolicyState.default_target` and `ChatRouteRule.action`, so a persisted policy can encode the dev-only `/ws/voice/{actorId}` bypass path. That contradicts the ADR text (D5: "the policy-aware endpoint should never produce it") and the field's own comment ("not a rule action"), and it weakens CLAUDE.md's "API 字段单一语义" by making one action type serve both policy decisions and the explicit-actor bypass contract. Consider splitting the types — e.g. keep persisted policy actions to `ForwardToModel | ForwardToGAgent | Reject`, and give the dev/explicit-entry path a separate action message if `Bypass` still needs a wire shape.

src/Aevatar.ChatRouting.Abstractions/chat_route_policy.proto
```diff
@@ -0,0 +1,221 @@
...
+  ChatRouteCallerScope owner_scope = 1;
+  ChatRouteAction default_target = 2;
+  repeated ChatRouteRule rules = 3;
+}
+
+message RemoveChatRouteRuleRequested {
+  string rule_id = 1;
+}
+
+message ResetChatRoutePolicyRequested {
```
eanzhao: **[codex]** severity=major, category=design

`ResetChatRoutePolicyRequested` explicitly clears `default_target` ("Wipes rules and clears default_target"), but `ChatRoutePolicyState.default_target` is documented as **REQUIRED** when the actor exists, and ADR D6 says the env-driven fallback is *only* for the cold-start case where the policy actor itself doesn't exist. So a reset leaves a durable, existing policy state with no `default_target` that the resolver cannot distinguish from a malformed policy — the message comment's claim that "resolver will fall back to env default until a valid one is upserted" directly contradicts D6. Either make `Reset` require a replacement `default_target`, model reset as deleting/tombstoning the policy actor, or widen ADR D6 so the env fallback also covers a reset (target-less) actor — and align all three docs.

aevatar.slnx
```diff
@@ -79,6 +79,7 @@
     <Project Path="src\Aevatar.AI.ToolProviders.Workflow\Aevatar.AI.ToolProviders.Workflow.csproj" />
     <Project Path="src\Aevatar.Bootstrap.Extensions.AI\Aevatar.Bootstrap.Extensions.AI.csproj" />
     <Project Path="src\Aevatar.Bootstrap\Aevatar.Bootstrap.csproj" />
+    <Project Path="src/Aevatar.ChatRouting.Abstractions/Aevatar.ChatRouting.Abstractions.csproj" />
```
eanzhao: **[Consensus: 4 models]** severity=minor, category=style

glm-5.1, mimo-v2.5-pro, kimi, and gemini all independently flagged this: the new `<Project Path>` entry uses forward slashes (`src/Aevatar.ChatRouting.Abstractions/...`) while its immediate neighbors — and the large majority of entries in this file — use backslashes (`src\...`). MSBuild accepts both, so there is no functional impact; the lone forward-slash entry wedged between two backslash entries is purely a consistency / diff-noise nit. Switch to `src\Aevatar.ChatRouting.Abstractions\Aevatar.ChatRouting.Abstractions.csproj` to match the surrounding block.

<details>

- **glm-5.1**: Forward-slash path separator (`src/Aevatar.ChatRouting.Abstractions/...`) is inconsistent with every other `<Project Path>` entry in this file (all use backslashes). Both work in MSBuild, but the mismatch creates unnecessary diff noise. Change to `src\Aevatar.ChatRouting.Abstractions\Aevatar.ChatRouting.Abstractions.csproj` for consistency.
- **mimo-v2.5-pro**: Path separator inconsistency: new entry uses forward slashes (`src/Aevatar.ChatRouting.Abstractions/...`) while both immediate neighbors at lines 81 and 82 use backslashes (`src\Aevatar.Bootstrap\...`, `src\Aevatar.Configuration\...`). The slnx already has mixed separators elsewhere (lines 5–8 use `/`), but the inserted position between two `\` entries makes it visually jarring. Recommend using `src\Aevatar.ChatRouting.Abstractions\Aevatar.ChatRouting.Abstractions.csproj` to match the surrounding block.
- **kimi**: The new `<Project Path="src/Aevatar.ChatRouting.Abstractions/..." />` uses forward slashes while every other entry in this file uses backslashes (`src\...`). Prefer backslashes for consistency with the existing 80+ entries.
- **gemini**: Inconsistent path separator. Neighbors in this section (e.g., lines 79-81) use backslashes (`\`), but this entry uses forward slashes (`/`). Suggest using `src\Aevatar.ChatRouting.Abstractions\Aevatar.ChatRouting.Abstractions.csproj` for consistency.
</details>

src/Aevatar.ChatRouting.Abstractions/chat_route_policy.proto
```diff
@@ -0,0 +1,221 @@
...
+
+// ─── Caller identity ───
+//
+// Mirrors Aevatar.GAgents.Scheduled OwnerScope but stays local to
+// ChatRouting.Abstractions so this project doesn't depend on the Scheduled
+// agent package. Phase 2 wires a thin mapper at the resolver boundary
+// (CompositeCallerScopeResolver output → ChatRouteCallerScope). Moving
+// OwnerScope to Foundation.Abstractions is a deferred refactor — out of
+// scope for the ingress-layer milestone.
+message ChatRouteCallerScope {
```
eanzhao, line 17: **[gemini]** severity=nit, category=arch

Mirroring `ChatRouteCallerScope` field-for-field with `Aevatar.GAgents.Scheduled.OwnerScope` is a sound temporary measure to avoid a circular dependency. The doc comment and ADR-0024 (§D7 + "Out of scope") already record this as a deferred refactor; consider also adding a tracking issue reference here so the duplication has a concrete consolidation owner rather than only prose. (Low priority — the ADR already documents the intent.)

src/Aevatar.ChatRouting.Abstractions/chat_route_policy.proto
```diff
@@ -0,0 +1,221 @@
...
+}
+
+message RemoveChatRouteRuleRequested {
+  string rule_id = 1;
+}
+
+message ResetChatRoutePolicyRequested {
+  // Wipes rules and clears default_target. Caller responsibility to set
+  // a sane default_target after — resolver will fall back to env default
+  // until a valid one is upserted.
```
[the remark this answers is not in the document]
eanzhao: Addressed in a297c1d2. `ResetChatRoutePolicyRequested` removed entirely — `default_target` is REQUIRED whenever the actor exists, and the env fallback (ADR D6) only covers actor-absent / readmodel-unavailable cold start. Callers that want to wipe and replace should use `UpsertChatRoutePolicyRequested` with the new `default_target` and an empty `rules` list — atomic, single committed event, no transient invalid window.

src/Aevatar.ChatRouting.Abstractions/chat_route_policy.proto
eanzhao: Fixed in a297c1d2. `Bypass` removed from `ChatRouteAction` (deleted `message Bypass`; tag 5 + name `bypass` are `reserved` on `ChatRouteAction` to prevent silent reuse). The dev `/ws/voice/{actorId}` endpoint never produces a `ChatRouteAction` — it reads `actorId` from the route and short-circuits the resolver — so no replacement wire shape is needed. ADR-0024 §D5 rewritten to spell this out.

src/Aevatar.ChatRouting.Abstractions/chat_route_policy.proto
eanzhao: Fixed in a297c1d2 — same change as your major/arch comment on Bypass. Chose option (b) from your list: dropped `ResetChatRoutePolicyRequested` entirely. ADR D6 stays as-is (env fallback only for actor-absent). Reset semantics now expressed as `Upsert(default_target, rules=[])` — atomic, never produces an invalid persisted state. §D5 text added explaining the removal.

aevatar.slnx
eanzhao: Fixed in a297c1d2. Switched `<Project Path>` to backslashes (`src\Aevatar.ChatRouting.Abstractions\...`) to match the surrounding `/src/` block.

src/Aevatar.ChatRouting.Abstractions/chat_route_policy.proto
eanzhao: Addressed in a297c1d2. Filed #700 to track the OwnerScope → Foundation.Abstractions move and updated the `ChatRouteCallerScope` doc comment to reference it inline, so the deferred consolidation has a concrete owner rather than only the ADR prose.

src/Aevatar.ChatRouting.Core/Aevatar.ChatRouting.Core.csproj
```diff
@@ -0,0 +1,23 @@
...
+        <ImplicitUsings>enable</ImplicitUsings>
+        <Nullable>enable</Nullable>
+        <AssemblyName>Aevatar.ChatRouting.Core</AssemblyName>
+        <RootNamespace>Aevatar.ChatRouting.Core</RootNamespace>
+    </PropertyGroup>
+
+    <ItemGroup>
+        <ProjectReference Include="..\Aevatar.ChatRouting.Abstractions\Aevatar.ChatRouting.Abstractions.csproj" />
+        <ProjectReference Include="..\Aevatar.CQRS.Projection.Stores.Abstractions\Aevatar.CQRS.Projection.Stores.Abstractions.csproj" />
+        <ProjectReference Include="..\..\agents\Aevatar.GAgents.ChatRouting\Aevatar.GAgents.ChatRouting.csproj" />
```
eanzhao: **[Consensus: 2 models]** severity=blocker, category=arch

Two reviewers independently flag this exact line. `src/Aevatar.ChatRouting.Core` declares a `ProjectReference` to `agents/Aevatar.GAgents.ChatRouting`, so a `src/` project depends *outward* on an `agents/` project. Both trace the root cause to `ChatRoutePolicyCurrentStateDocument` (the readmodel document type) living in the agents project while `ChatRoutePolicyQueryPort` in Core consumes it. Per CLAUDE.md 严格分层 / 禁止跨层反向依赖, the readmodel document (+ its proto) should move into a `src/` project so the dependency direction is `agents/` → `src/`, not the reverse. Severity split: kimi=blocker, mimo=major.

<details>

- **kimi**: `src/Aevatar.ChatRouting.Core.csproj` references `agents/Aevatar.GAgents.ChatRouting.csproj`, which is a reverse dependency: `src/` (foundation/core layer) must not depend on `agents/` (application layer). Per CLAUDE.md “严格分层” and “上层依赖抽象，禁止跨层反向依赖”. The root cause is `ChatRoutePolicyCurrentStateDocument` (the readmodel DTO) living in the agents project while `ChatRoutePolicyQueryPort` in Core needs to query it. Move the readmodel document (and its proto) into `Aevatar.ChatRouting.Abstractions` or `Aevatar.ChatRouting.Core`, then have the agents project reference Core.
- **mimo-v2.5-pro**: **Layering violation**: `src/` (core/inner layer) has a `<ProjectReference>` to `agents/Aevatar.GAgents.ChatRouting` (outer/application layer). CLAUDE.md: "严格分层" and "依赖反转：上层依赖抽象，禁止跨层反向依赖". The inward dependency exists because `ChatRoutePolicyQueryPort` in `src/Aevatar.ChatRouting.Core` references `ChatRoutePolicyCurrentStateDocument` defined in `agents/Aevatar.GAgents.ChatRouting`. Fix: move `ChatRoutePolicyCurrentStateDocument` (and its partial class + readmodel proto) into `src/Aevatar.ChatRouting.Core` or a shared `src/Aevatar.ChatRouting.ReadModel` project, so `agents/` depends on `src/` and not vice versa.
</details>

src/Aevatar.ChatRouting.Core/ChatRoutePolicyQueryPort.cs
```diff
@@ -0,0 +1,71 @@
...
+        if (document?.DefaultTarget is null ||
+            document.DefaultTarget.ActionCase == ChatRouteAction.ActionOneofCase.None)
+        {
+            return null;
+        }
+
+        return new ChatRoutePolicySnapshot(document.DefaultTarget, document.Rules);
+    }
+
+    private static IReadOnlyList<ProjectionDocumentFilter> BuildCallerScopeFilters(OwnerScope callerScope) =>
```
eanzhao, line 83: **[Consensus: 2 models]** severity=blocker, category=bug

codex and gemini independently flag `BuildCallerScopeFilters`: it ANDs four exact-equality (`Eq`) filters across every `OwnerScope` field (NyxUserId, Platform, RegistrationScopeId, SenderId). A scope-only / 'generic channel' policy stored with empty NyxUserId/SenderId can therefore never match a caller whose scope has those fields populated — which `ConversationGAgent` does once relay ingress resolves the sender — so `LookupForCallerAsync` silently returns no policy and the default/rule is ignored. Both recommend trying the most-specific tuple first, then falling back to the scope-only tuple, making the precedence explicit instead of requiring an exact 4-field match. Severity split: gemini=blocker, codex=major.

<details>

- **gemini**: `BuildCallerScopeFilters` uses strict equality (`Operator.Eq`) on all 4 fields of `OwnerScope`. In channel scenarios (e.g., Relay), `ConversationGAgent` always populates `SenderId` and `NyxUserId` based on incoming traffic. If a generic scope-wide policy was upserted (with empty user/sender fields), this exact-match query will return zero documents, effectively ignoring the bot's default policy. Implement fallback queries in `LookupForCallerAsync` (e.g., try the exact match, then try `OwnerScope.ForChannel("", callerScope.Platform, callerScope.RegistrationScopeId, "")`).
- **codex**: The exact `OwnerScope.NyxUserId` filter means a channel policy written as scope-only (`nyx_user_id == ""`, which `ConversationGAgent` explicitly says should be a fallback shape) will never match once relay ingress successfully resolves `TransportExtras.NyxSenderUserId`. `LookupForCallerAsync` should try the most-specific per-user tuple first and then fall back to the scope-only tuple, or encode that precedence explicitly in the readmodel query.
</details>

agents/Aevatar.GAgents.ChatRouting/DependencyInjection/ChatRoutingServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,65 @@
...
+        services.AddProjectionMaterializationRuntimeCore<
+            ChatRoutePolicyMaterializationContext,
+            ChatRoutePolicyMaterializationRuntimeLease,
+            ProjectionMaterializationScopeGAgent<ChatRoutePolicyMaterializationContext>>(
+            static scopeKey => new ChatRoutePolicyMaterializationContext
+            {
+                RootActorId = scopeKey.RootActorId,
+                ProjectionKind = scopeKey.ProjectionKind,
+            },
+            static context => new ChatRoutePolicyMaterializationRuntimeLease(context));
```
eanzhao, line 56: **[codex]** severity=blocker, category=di

`AddCurrentStateProjectionMaterializer` only registers the projector; it does not activate a projection scope or subscribe to `ChatRoutePolicyGAgent` committed events. Existing materialized readmodels such as `UserAgentCatalog` and `DeviceRegistration` add a projection port plus startup/write-path `EnsureProjectionForActorAsync`; without the same activation path here, `ChatRoutePolicyQueryPort` will keep returning `null` even after policy commands commit. Add a `ChatRoutePolicyProjectionPort` and invoke it from the policy write path or a startup/discovery service before relying on the readmodel.

agents/Aevatar.GAgents.ChatRouting/DependencyInjection/ChatRoutingServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,65 @@
...
+        services.AddCurrentStateProjectionMaterializer<
+            ChatRoutePolicyMaterializationContext,
+            ChatRoutePolicyCurrentStateProjector>();
+
```
eanzhao, line 60: **[codex]** severity=major, category=arch

This unconditionally wires the chat-route policy readmodel to `InMemory`, and `MainnetHostBuilderExtensions` calls `AddChatRoutingAgents()` on the production host path. That violates CLAUDE.md's `InMemory` rule (`InMemory` only for dev/tests) and makes route policy visibility node-local/non-durable. Follow the existing `AddScheduledAgents(configuration)` / `AddChannelIdentityProjectionStores(configuration)` pattern and choose Elasticsearch when the host configuration enables it, keeping in-memory only as the local/test fallback.

src/Aevatar.Mainnet.Host.Api/Responses/ResponsesEndpoints.cs
```diff
@@ -93,6 +99,36 @@ internal static async Task<IResult> HandleCreateResponseAsync(
...
+        if (routeDecision.Action.Reject is not null)
+            return ToErrorResult(
+                StatusCodes.Status403Forbidden,
+                "chat_route_rejected",
+                string.IsNullOrWhiteSpace(routeDecision.Action.Reject.Reason)
+                    ? "The chat route policy rejected this request."
+                    : routeDecision.Action.Reject.Reason);
+        if (!string.IsNullOrWhiteSpace(routeDecision.Action.ForwardToModel?.ModelName))
+        {
+            routedModel = routeDecision.Action.ForwardToModel.ModelName.Trim();
```
eanzhao, line 133: **[codex]** severity=major, category=bug

A policy decision of `ForwardToGAgent` is silently ignored here and the request continues through model dispatch. That bypasses an explicit route policy: a default/rule targeting a GAgent will still call the LLM model path. Return an honest unsupported-action response (for example 501/403) or implement the GAgent dispatch for this source; apply the same behavior to `/v1/messages`.

src/Aevatar.ChatRouting.Core/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Aevatar.ChatRouting.Core;
+
+public static class ServiceCollectionExtensions
+{
+    public static IServiceCollection AddChatRoutingCore(this IServiceCollection services)
+    {
+        ArgumentNullException.ThrowIfNull(services);
+
+        services.AddOptions<ChatRoutingOptions>();
```
eanzhao: **[gemini]** severity=major, category=di

`services.AddOptions<ChatRoutingOptions>()` registers the options type but does not bind it to any configuration section. As a result, `_options.Value.Defaults.FallbackModel` in `EnvChatRouteFallbackProvider` will always be empty (unless configured externally by a consumer before or after this call). Append `.BindConfiguration("ChatRouting")` (or take `IConfiguration` to `Configure<ChatRoutingOptions>`) to map the settings from `appsettings.json`.

agents/Aevatar.GAgents.NyxidChat/NyxIdChatEndpoints.cs
```diff
@@ -85,16 +87,45 @@ private static async Task<IResult> HandleCreateConversationAsync(
...
-        var actorId = NyxIdChatServiceDefaults.GenerateActorId();
-        await actorRuntime.CreateAsync<NyxIdChatGAgent>(actorId, ct);
+        var forwardedActorId = decision.Action.ForwardToGagent?.ActorId;
+        var actorId = !string.IsNullOrWhiteSpace(forwardedActorId)
+            ? forwardedActorId.Trim()
+            : NyxIdChatServiceDefaults.GenerateActorId();
+        // We only own the actor's lifecycle when we created it in this request.
+        // A forwarded ChatRoute decision reuses an actor that an earlier request
+        // (possibly under a different scope) created; destroying it on a
+        // registry rollback would delete unrelated traffic's target actor.
```
eanzhao, line 125: **[kimi]** severity=major, category=bug

When `decision.Action.ForwardToGagent` is non-null but its `ActorId` is empty/whitespace, the code generates a new actor ID (`NyxIdChatServiceDefaults.GenerateActorId()`) yet sets `createdLocally = false` because `ForwardToGagent is null` evaluates to false. This skips `actorRuntime.CreateAsync`, so the actor is never created before registration is attempted. On rollback, `destroyActor: false` prevents cleanup of the never-created actor. Fix: `createdLocally` should reflect whether this request actually created the actor, e.g. `createdLocally = decision.Action.ForwardToGagent is null || string.IsNullOrWhiteSpace(decision.Action.ForwardToGagent.ActorId);`.

src/Aevatar.Mainnet.Host.Api/Voice/PolicyAwareVoiceEndpoints.cs
```diff
@@ -0,0 +1,342 @@
...
+        var sessionResolver = http.RequestServices.GetRequiredService<IVoicePresenceSessionResolver>();
+        var moduleName = FirstNonEmpty(action.ForwardToGagent.VoiceModuleName, routeInput.Voice?.VoiceModuleName);
+        var session = await sessionResolver.ResolveAsync(
+            new VoicePresenceSessionRequest(actorId, moduleName),
+            http.RequestAborted);
+        if (session is null)
+        {
+            http.Response.StatusCode = StatusCodes.Status404NotFound;
+            await http.Response.WriteAsync("Voice session not found for this agent.", http.RequestAborted);
+            return;
```
eanzhao, line 112: **[mimo-v2.5-pro]** severity=minor, category=di

Service-locator pattern via `GetRequiredService<IChatRoutePolicyQueryPort>()` and `GetRequiredService<ChatRouteResolver>()` inside the handler delegate. While at a host boundary, this bypasses constructor injection which makes the dependency surface implicit and harder to test via unit tests (only integration tests via `TestServer` are possible). Consider making these endpoint methods accept the services as parameters via ASP.NET's `[FromServices]` attribute, matching the pattern already used by `NyxIdChatEndpoints` and `ResponsesEndpoints` elsewhere in this PR.

src/Aevatar.ChatRouting.Core/ChatRouteResolver.cs
```diff
@@ -0,0 +1,119 @@
...
+    {
+        ArgumentNullException.ThrowIfNull(input);
+
+        if (snapshot is null)
+        {
+            return _fallbackProvider.GetFallbackDecision();
+        }
+
+        foreach (var rule in snapshot.Rules
+                     .OrderByDescending(static rule => rule.Priority)
```
eanzhao: **[gemini]** severity=minor, category=perf

The policy actor already pre-orders the rules before persisting them, so the readmodel supplies them in priority order (as documented in `ChatRoutePolicyGAgent.cs`). Re-sorting `snapshot.Rules` with `.OrderByDescending(...).ThenBy(...)` here on the hot path during every resolve call is redundant and wastes CPU cycles. Evaluate `snapshot.Rules` directly.

docs/README.md
```diff
@@ -53,6 +53,7 @@ Immutable records of architectural choices and their rationale.
 - [Lark Reply Chain Completion Semantics](adr/0021-lark-reply-chain-completion-semantics.md)
 - [0022 — OpenTelemetry semantic conventions for aevatar.* activities](adr/0022-otel-aevatar-semantic-conventions.md)
 - [0023 — Two-tier Inspector architecture (canonical readmodel vs observation OTel)](adr/0023-two-tier-inspector-architecture.md)
+- [Chat Route Policy — Config Actor + Boundary Resolver](adr/0024-chat-route-policy.md)
```
eanzhao, line 57: **[kimi]** severity=minor, category=style

Manual edit to `docs/README.md` violates CLAUDE.md docs rule: “`docs/README.md` 由 `tools/docs/build-index.sh` 自动生成，不手动编辑。” The ADR and audit-scorecard entries should be picked up by the index generator; manual insertion risks drift on subsequent doc builds. Run `bash tools/docs/build-index.sh` and revert the hand-edited lines.

src/Aevatar.ChatRouting.Core/OwnerScope.cs
```diff
@@ -0,0 +1,37 @@
+namespace Aevatar.ChatRouting.Core;
+
+/// <summary>
+/// Caller ownership tuple used by chat routing read ports.
+/// </summary>
+public sealed class OwnerScope
```
eanzhao, line 8: **[mimo-v2.5-pro]** severity=nit, category=arch

Duplicate of `Aevatar.GAgents.Scheduled.OwnerScope` (acknowledged in ADR-0024 D7 and proto comment). The two types have identical fields and factory methods. Track expedited consolidation into `Aevatar.Foundation.Abstractions` to avoid drift — two independent `OwnerScope` types will silently diverge as new fields are added to one but not the other.

PR #740, opened 2026-05-20
Fix draft-run SSE accepted timing

louis4li:
## Summary

- emit the draft-run accepted/runStarted frame after command preparation and before awaiting prepared dispatch
- keep stream pumping after dispatch admission so long-running actor/LLM execution no longer leaves SSE clients with an open zero-frame response
- map committed RoleChatSessionCompletedEvent terminal facts through the draft-run projection session so SSE receives real downstream terminal frames
- return terminal provider/LLM failures as AG-UI runError frames instead of falling back to a generic endpoint timeout

## Issue #736 scope

This PR fixes issue #736 problem 2. I also tried to reproduce problem 1 locally in distributed mode, including the PR #702 GAgent member payload, but the member binding path returned 202 and the binding run succeeded; missing members were asynchronously rejected with STUDIO_MEMBER_NOT_FOUND. No 500 was reproduced locally, so this PR intentionally does not claim to fix problem 1.

## Verification

- dotnet test test/Aevatar.Workflow.Application.Tests/Aevatar.Workflow.Application.Tests.csproj --nologo --filter "FullyQualifiedName~WorkflowApplicationLayerTests"
- dotnet test test/Aevatar.Studio.Tests/Aevatar.Studio.Tests.csproj --nologo --filter "FullyQualifiedName~StudioMemberEndpointsTests|FullyQualifiedName~ScopeBindingStudioMemberPlatformBindingCommandServiceTests"
- dotnet test test/Aevatar.GAgentService.Tests/Aevatar.GAgentService.Tests.csproj --nologo --filter "FullyQualifiedName~ScopeGAgentAguiEventMapperTests|FullyQualifiedName~GAgentDraftRunInteractionCoverageTests"
- dotnet test test/Aevatar.GAgentService.Integration.Tests/Aevatar.GAgentService.Integration.Tests.csproj --nologo --filter "FullyQualifiedName~ScopeServiceEndpointsStreamTests|FullyQualifiedName~ScopeGAgentEndpointsTests"
- dotnet build aevatar.slnx --nologo

## Local runtime check

Started Mainnet Host in distributed mode against local Kafka/Garnet/Elasticsearch with Orleans/Kafka/Garnet runtime configuration and called POST /api/scopes/{scopeId}/gagent/draft-run. Neo4j still reported an auth failure in Development and was ignored by startup, matching the earlier local setup.

The SSE response now produces a real downstream terminal event, not just an initial runStarted frame and not a generic timeout. With no local NyxID auth token, the observed frames were:

```text
data: { "runStarted": { "threadId": "Role:ce480c92", "runId": "27699f500f124d3c877d9db5f2ce20a4" } }

data: { "runError": { "message": "LLM request failed [tools=...]: NyxID authentication required for provider 'nyxid'. Please sign in." } }\n```\n\nSo the local unauthenticated environment cannot prove a successful `pong` text response, but it does prove the actor/LLM path reaches a terminal provider result and that the SSE chain returns that terminal data promptly instead of hanging until timeout.

### CODE REVIEW

File: src/Aevatar.CQRS.Core/Interactions/DefaultCommandInteractionService.cs [hunk 1]
```diff
@@ -61,6 +61,8 @@ public async Task<CommandInteractionResult<TReceipt, TError, TCompletion>> Execu
             if (onAcceptedAsync != null)
                 await onAcceptedAsync(receipt, ct);
 
+            await _dispatchPipeline.DispatchPreparedAsync(execution, ct);
```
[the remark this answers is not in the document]
louis4li: Thanks for calling this out. For this endpoint the early frame is intentional: after PrepareAsync succeeds we have a stable command/run id and can start the SSE stream before the potentially long actor/LLM dispatch path. Moving onAcceptedAsync back after DispatchPreparedAsync would reintroduce the zero-frame/late-accepted behavior this PR is fixing. If DispatchPreparedAsync fails after that point, the honest outcome should be a stream error/terminal frame rather than a pre-stream start error. I also rechecked the follow-up projector change so committed completions no longer replay content when ContentEmitted=true; they only synthesize the missing terminal frames.

jason-aelf: This moves accepted/runStarted before dispatch. For GAgent draft-run, `onAcceptedAsync` starts the SSE response and sets `ResponseStarted` to true. If `DispatchPreparedAsync` or the actor handler throws afterward, the endpoint exception path skips prepared actor rollback because `ResponseStarted == true`.

This affects draft-run requests that create a new actor. `GAgentDraftRunActorPreparationService` marks newly created actors with `RequiresRollbackOnFailure: true`, but after early `runStarted`, the failure path may no longer unregister/destroy that temporary actor, leaving the actor and registry entry behind. Reusing an existing actor is not affected.

louis4li: Good catch — this was a real issue. I pushed dbafc053 to address it.

What changed:
- draft-run exception/timeout/client-disconnect paths now call RollbackPreparedActorAsync even after the SSE response has started
- rollback remains gated by RequiresRollbackOnFailure, so existing actor reuse is not affected
- added coverage for the case where runStarted is emitted first and a later failure still rolls back the prepared temporary actor

Validation:
- dotnet test test/Aevatar.GAgentService.Integration.Tests/Aevatar.GAgentService.Integration.Tests.csproj --nologo --no-restore --filter "FullyQualifiedName~ScopeServiceEndpointsStreamTests" — passed 19/19
- git diff --check — passed

PR #1039, opened 2026-05-25
iter98 cluster-743: delete P4 query-time ES mapping reader;alias+fingerprint lifecycle 唯一 schema-drift 权威

### LINKED ISSUE #743: ES projection index lifecycle: schema-drift gap silently breaks by-field queries (Lark bot outage 2026-05-20)
## TL;DR

PR #665 ("Stabilize Elasticsearch projection index mappings", merged 2026-05-18) silently broke every by-field ES filter on `_id`-suffix string fields whose index was created **before** that date. The Lark bot has been 100% unable to reply since the change reached production: `NyxIdRelayScopeResolver` cannot resolve `apiKeyId → scopeId`, the relay endpoint returns 401, the user message never enters business logic.

The bug itself is one structural property: `BuildExactMatchFieldPathResolver` trusts in-memory augmented metadata as ground truth, while pre-existing ES indices keep their dynamic-default mapping forever — code and ES diverge permanently. This issue captures the full chain and proposes a long-term lifecycle fix that does not rely on operators running runbooks manually.

## 1. Production symptom (2026-05-20)

`POST /api/webhooks/nyxid-relay` returns 401 on every Lark callback:

```
warn: Aevatar.NyxId.Chat.Relay
  Relay callback authentication succeeded but did not resolve a canonical scope id:
  message=<uuid>, apiKeyId=801667e9-772d-4bdf-8000-717ce331746c
→ HTTP 401
```

Cross-checks:

- `GET /api/channels/registrations` returns the row with the same apiKey → readmodel itself is healthy.
- Direct ES probe on `aevatar-mainnet-channel-bot-registrations` with the exact code-generated query body returns **1 hit** if the term targets `nyx_agent_api_key_id.keyword`, **0 hits** if it targets `nyx_agent_api_key_id` (the plain text field).
- `_mapping/field/nyx_agent_api_key_id` shows ES has `type: text` with `fields.keyword: type:keyword` (dynamic-default multi-field, as created on 2026-04-22).

## 2. Mechanism

`ElasticsearchProjectionDescriptorMappingSupport.AugmentMetadata` (added by #665) walks the proto descriptor and, for any string field whose name matches `IsStableKeywordFieldName` (anything ending in `_id`, `_key`, `_hash`, `_status`, `_kind`, `_type`, `_type_url`, `_revision`, `_revision_id`, `_actor_id`, plus literal `id`/`actor_id`/`last_event_id`), registers `{"type": "keyword"}` in the in-memory augmented `indexMetadata.Mappings`.

`BuildExactMatchFieldPathResolver` consults that augmented metadata:

```csharp
if (TryGetFieldMapping(indexMetadata.Mappings, resolvedFieldPath, out var explicitMapping)) {
    if (IsKeywordFieldMapping(explicitMapping))  return resolvedFieldPath;            // ← chosen here
    if (HasKeywordMultiField(explicitMapping))   return $"{resolvedFieldPath}.keyword";
    return resolvedFieldPath;
}
// fallthrough: descriptor String field → append .keyword
```

For `nyx_agent_api_key_id`, the augmented mapping says `keyword`, so the resolver returns the field **without** `.keyword`. The term query targets the `text` field. ES doesn't match a UUID-like value against an analyzed text field (no token equals the full string), so the query returns 0 hits even though the document is right there.

`EnsureIndexAsync` only creates the index if it is missing — it never reconciles existing mappings against the augmented metadata, never reindexes, never alerts. Code intent and ES reality diverge permanently the moment the schema in code changes.

## 3. Originating change

| Field | Value |
|---|---|
| PR | #665 "Stabilize Elasticsearch projection index mappings" |
| Author | @louis4li (Claude Opus 4.6 co-author) |
| Merged | 2026-05-18 09:29 UTC by @potter-sun |
| Fixes | #533 |
| Design doc | `docs/design/2026-05-15-elasticsearch-projection-index-mapping-blueprint.md` |
| Commit | 7ac9aa72bd0561354d347d3fe6b344361cf7b0f9 |
| `reviewDecision` | `(none)` |
| Formal reviews | 0 |
| Non-bot comments | 0 |

The PR description was self-aware: *"Note: mapping changes are not backward-compatible with old Elasticsearch projection indexes; clear or rebuild affected projection indexes."* The blueprint formalized the same decision in §5 and §8:

> 5. 若变更影响旧 index，PR 描述中写清"不兼容旧 index，需清空 / 重建 projection index"；不要设计历史兼容层。
> 8. 任何实现都不得把 read path 变成 index repair / rebuild path.

That is a defensible architectural stance — until you observe that **no operator ran the clear/rebuild step**, no CI gate enforces it, and `feature/router` (the live deploy branch) auto-deploys on push. Four review/gate layers failed at once:

1. **0 formal reviews** on PR #665 — `reviewDecision` was empty when it merged.
2. **Codex automated review did not trigger** for #665 (same author's PR #657 from the same week did trigger it).
3. **Deploy-time gate missing** — auto-deploy of the merged branch to mainnet does not require a "runbook executed" checkbox.
4. **CI does not include a real-ES integration test against a pre-existing dynamic-mapped index**. Local tests all build fresh indices, which by definition can't see schema drift.

## 4. Architectural gap

> Default 路径须定义资源语义：任何"缺失即创建"策略须同时定义归属、复用规则和清理责任。 — CLAUDE.md

`ElasticsearchProjectionDocumentStore` implements "缺失即创建" but has no "schema drift → migrate" or "stale → cleanup" counterpart. Augmented metadata is the code's intent; ES is the source of physical truth; **nothing reconciles them**. The query path silently trusts intent over truth and returns 0 hits when they differ.

A correct framework cannot rely on operators executing migration runbooks faithfully every time a `_id`-shape field appears in a readmodel proto — that is precisely the kind of偶然 correctness CLAUDE.md says is undone designs.

## 5. Long-term solution — index versioning + alias indirection + schema-fingerprint reconciliation

Treat ES schema as a first-class evolution unit. Move every read/write through aliases; let `IndexLifecycleManager` own version transitions; let the resolver consult ES truth, not code intent.

### 5.1 Alias indirection

For every projection readmodel:

- **read alias** = current public index name (e.g., `aevatar-mainnet-channel-bot-registrations`)
- **write alias** = `<read>-write` (same target during steady state; differs only during migration)
- **physical index** = `<read>-v{N}` where N is a monotonic version derived from the schema fingerprint

All `QueryAsync` / `UpsertAsync` calls target aliases. Physical index churn is invisible to business code.

### 5.2 Schema fingerprint

`DocumentIndexMetadata` exposes `ComputeFingerprint()` over the canonicalized augmented `Mappings` dictionary (sorted keys, normalized number/string forms, SHA-256 truncated to 8 hex). The fingerprint is the physical-index version suffix.

When proto changes or `IsStableKeywordFieldName` rules change, the fingerprint changes deterministically; otherwise it is stable.

### 5.3 `IndexLifecycleManager.EnsureIndexAsync` upgraded to reconcile + migrate

On store construction (or at host startup, behind one `IHostedService` per provider):

1. Look up which physical index the read alias currently points at.
2. Pull `_mapping` and compute its actual fingerprint.
3. Compare against expected fingerprint (from augmented metadata).
4. **Equal** → no-op.
5. **Different** →
   - Create `<read>-v{newN}` with the expected mapping payload.
   - Issue `POST _reindex` from old → new (small indices < 1 s; large ones run as background ES tasks tracked by a registered task id).
   - Atomic `_aliases` swap: read alias now points to v{newN}; write alias too.
   - Publish a `ProjectionScopeRebuildSignal` so the authoritative actor re-emits its `CommittedStateEventPublished` envelope — readmodel idempotent overwrite catches up. Required because reindex copies docs but does not reconcile any since-divergence drift; the actor-side state is authoritative.
   - Mark v{oldN} for cleanup; background job DELETEs after N days (configurable rollback window).
6. **Failure during migration** → fail loud at startup; host does not accept requests.

### 5.4 Reader trusts ES, not augmented metadata

`BuildExactMatchFieldPathResolver` reads the **actual** ES mapping of the current physical index (cached, invalidated when alias swaps or fingerprint changes). Augmented metadata's only remaining role is fingerprint computation + v{N} template generation. The query path's keyword/text decision is sourced from physical reality, not code's projected intent.

### 5.5 Fail-loud at host startup

The lifecycle manager must refuse to bring the host online if:
- read alias is missing AND first-time creation fails;
- alias points to a physical index whose mapping diverges from expected AND migration is blocked (ES auth failure, reindex stalled, conflicting in-flight migration);
- a previous migration left an inconsistent state (v{newN} created but alias not swapped).

CLAUDE.md "本地可用不等于分布式正确：依赖本地 runtime 偶然细节才成立的实现视为未完成设计" — silent 0-hit queries are the textbook偶然 correctness this rule rejects.

### 5.6 Test coverage gate

Add a `Aevatar.CQRS.Projection.Providers.Elasticsearch.IntegrationTests` project against a real ES (Testcontainers) that exercises:

- Build index with dynamic mapping → run new code → verify migration happens → verify by-field query hits.
- Build index with explicit mapping → run new code → verify no migration (fingerprint matches).
- Migration mid-failure (kill reindex) → restart store → verify recovery resumes or fails loud.
- Add this test as a CI guard required for any PR touching `src/Aevatar.CQRS.Projection.Providers.Elasticsearch/`.

This is the regression door that PR #665 walked through; closing it permanently is part of the fix.

## 6. How this recovers the Lark bot

The first host startup after the new lifecycle ships:

1. `IndexLifecycleManager` enumerates `aevatar-mainnet-*` projection indices.
2. For `channel-bot-registrations`, it discovers the read alias points to a v0 physical index whose fingerprint mismatches the expected one (because of the `_id`-suffix → keyword augmentation from #665).
3. It creates `aevatar-mainnet-channel-bot-registrations-v1` with the expected explicit-keyword mapping, reindexes the 1 existing doc (~14 KB), atomically swaps the read+write aliases, dispatches a `ProjectionScopeRebuildSignal` to `channel-bot-registration-store`.
4. The authoritative `ChannelBotRegistrationGAgent` re-emits its committed state events, the projection scope re-materializes the doc into v1, fingerprint matches, ✅.
5. Next Lark callback: resolver hits the now-v1 index with the right `.keyword` field path (or, post-5.4, with the ES-truth-driven path); resolver returns `ce646b72-...`; relay continues into `ConversationGAgent`; bot replies.

**Not a bypass. The first invocation of the new framework is its own validation.** All other affected indices created before 2026-05-18 (likely `aevatar-mainnet-user-agent-catalog-nyx-credentials`, `-aevatar-oauth-clients`, `-studio-*`, `-script-read-model-documents`) get the same fix in the same migration window, silently fixing every latent variant of this bug.

## 7. Implementation slices

| Phase | Scope | Risk |
|---|---|---|
| P1 | `DocumentIndexMetadata.ComputeFingerprint()` + alias naming convention + `IndexLifecycleManager` interface | internal API, no on-line impact |
| P3 | Migration path inside `EnsureIndexAsync` — **handles both first-time bare-index wrapping and aliased fingerprint drift** as a single state machine: create `<name>-v{fingerprint}` with expected mapping → reindex from current physical (bare or v{old}) → atomic `_aliases` swap (one ES call that adds the alias to the new physical and either removes the alias from the old physical or removes the bare index outright); subsequent host startups become no-ops; background cleanup of old v{N} physical indices after grace window | the core new code; needs reindex-failure rollback tests; ships **before** P2 so fail-loud only triggers on cases the migration can't handle |
| P2 | After P3 ships, tighten `EnsureIndexAsync` to fail loud when a physical index's mapping diverges from expected fingerprint and the migration cannot resolve it (ES auth failure, reindex stalled, conflicting in-flight migration) | breaking change on host startup if infra is unhealthy — staged through staging |
| P4 | Reader-side: resolver reads ES actual mapping; augmented metadata demoted to fingerprint + v{N} template only | changes query path; e2e coverage required |
| P5 | Real-ES integration tests via Testcontainers — bare-index wrap, fingerprint drift, recovery, reindex-failure rollback — wired as a CI guard for `src/Aevatar.CQRS.Projection.Providers.Elasticsearch/` | new CI dependency, longer CI time |

**Ordering note**: P3 ships before P2. The previous draft put P2 ("fail-loud") before P3 with a manual one-time alias annotation step for existing prod indices; that's a bypass we don't need. With P3 handling bare-index wrapping automatically, no manual annotation is required — the framework's first execution against an unaliased pre-2026-05-18 index IS the migration.

## 8. Non-goals (explicitly rejected to prevent scope drift)

- **No manual `DELETE` of any prod index.** Recovery must come through the new lifecycle, not operator ops.
- **No revert of #665.** The "stabilize index mappings" intent is correct; what was missing is the migration counterpart.
- **No heuristic patches to `BuildExactMatchFieldPathResolver`** that try to recognize text+keyword multi-field and append `.keyword`. That deepens the implicit-convention debt; the architectural fix above replaces it.
- **No historical compatibility shim** preserving dynamic-mapping behavior for pre-2026-05-18 indices. Per blueprint §5, but now backed by an automated migration path instead of an unenforced runbook.

## 9. Acceptance criteria

- [ ] All `aevatar-mainnet-*` projection indices have a `<name>` read alias pointing at a `<name>-v{fingerprint}` physical index. No bare-index reads/writes from code.
- [ ] `IndexLifecycleManager.EnsureIndexAsync` performs fingerprint compare + migration on startup; host fails loud if migration is blocked.
- [ ] `BuildExactMatchFieldPathResolver` decides `.keyword` from ES actual mapping, not augmented metadata.
- [ ] Real-ES integration test suite (Testcontainers) exists and runs in CI as a guard for `src/Aevatar.CQRS.Projection.Providers.Elasticsearch/`.
- [ ] Lark relay returns 200 for inbound callbacks; `NyxIdRelayScopeResolver` log shows `Resolved relay callback scope id from relay scope resolver` for live apiKeys.
- [ ] No operator runbook is required to deploy a future PR that changes proto field shapes for any projection readmodel.

## 10. References

- PR #665 — the originating change
- Issue #533 — the motivating problem statement
- `docs/design/2026-05-15-elasticsearch-projection-index-mapping-blueprint.md` — the design that deliberately omitted the migration path
- Commit `7ac9aa72bd0561354d347d3fe6b344361cf7b0f9`
- CLAUDE.md sections: "权威状态 / ReadModel / Projection（强制）", "Default 路径须定义资源语义", "正确架构优先"

loning:
## 摘要

iter98 cluster-743。ES projection index lifecycle:删除 P4 query-time ES actual mapping reader,改 alias+fingerprint lifecycle 作为唯一 schema-drift 权威。补 lifecycle fail-loud + 文档收口 + ES 回归门禁。

## Phase 9 r2 consensus(3/3 unanimous)

`META_JUDGE_DONE:consensus:delete-p4-lifecycle-fail-loud-only:三票一致删除 P4 query-time ES actual mapping reader,保留 alias+fingerprint lifecycle 为唯一 schema-drift 权威,剩余只做 lifecycle fail-loud、文档收口与真实 ES 回归门禁`

## 改动

- **删** P4 query-time ES mapping reader from `ElasticsearchIndexLifecycleManager.cs`
- **保留** alias+fingerprint lifecycle 唯一权威;simplified `ElasticsearchProjectionSchemaFingerprint.cs`
- **fail-loud**:drift detected → throw
- **加 ES 回归 test** `ProjectionProviderE2EIntegrationTests.cs`(+112 LOC dynamic-mapping drift coverage)
- **docs** `docs/canon/cqrs-projection.md` + design blueprint + README:schema-drift authority 收口

## 约束遵循

- ❌ 不加新 actor / envelope / pipeline / ADR
- ❌ 不读 query-time ES mapping(query resolver 改回 readmodel / static schema)
- ❌ 不破坏 NyxID/chrono-* 外部仓库
- ✅ 真实 ES provider e2e 覆盖 drift

## 验证

- `dotnet build aevatar.slnx --nologo` ✅
- `dotnet test ElasticsearchProjection` ✅
- guards ✅

净 +146 / -61 LOC。

closes #743

⟦AI:AUTO-LOOP⟧

### DIFF
--- a/docs/canon/cqrs-projection.md
+++ b/docs/canon/cqrs-projection.md
@@ -111,6 +111,7 @@ CQRS 不应只提供零散 helper，而应定义所有 capability 复用的标�
 6. 禁止 `Projection:ReadModel:Bindings` 与任何 BindingResolver 路由；投影存储路由统一由 `IProjectionStoreDispatcher` + Store Binding (`IProjectionDocumentStore` / `IProjectionGraphStore`) 决策。
 7. Host 组合层按配置仅注册所需 provider 组合，不允许无条件并列注册 InMemory/Elasticsearch/Neo4j。
 8. 持久投影 scope 的激活由 committed-state publication owner hook 触发：Actor 完成 domain event commit 并构造 `EventEnvelope<CommittedStateEventPublished>` 前，Foundation 调用 `ICommittedStatePublicationHook`，Projection Core 根据精确的 actor type 与 state event descriptor 生成 `ProjectionScopeStartRequest` 并分发给已有 `IProjectionScopeActivationService<TLease>`。命令入口不得同步调用 projection activation facade，也不得新增 actor/lifecycle phase 来“预热”读模型。
+9. Elasticsearch projection schema-drift 的唯一权威是 provider 生成的 augmented mapping fingerprint 与稳定 alias lifecycle。query resolver / query reader 不得读取 live ES mapping 作为第二真相；alias 指向非当前 fingerprint 的 physical index 时，lifecycle 必须 fail loud，projection 拒绝继续，而不是在 query-time 或 projection turn 内自动 repair / reindex。
 
 ## 5.1 编排减重落地（当前实现）
 
--- a/docs/design/2026-05-15-elasticsearch-projection-index-mapping-blueprint.md
+++ b/docs/design/2026-05-15-elasticsearch-projection-index-mapping-blueprint.md
@@ -43,8 +43,8 @@ owner: aevatar-core
 2. 稳定 read model 字段 mapping 应由 Elasticsearch provider 基于 protobuf descriptor 统一补齐，而不是由每个业务 metadata provider 重复手写。
 3. 显式 metadata provider mapping 优先。统一 helper 只补缺失的稳定字段，不覆盖业务 provider 已声明的 mapping。
 4. open / native / plugin document 区域必须显式保留开放边界，禁止为了标准化把动态脚本字段收窄。
-5. 不背历史 index 兼容包袱。当前尚未进入生产环境，mapping 变更允许要求清空 / 重建 Elasticsearch projection index；不做在线兼容、双轨 query fallback 或旧 index repair。
-6. 当前 `GetAsync` / `QueryAsync` 会调用 `EnsureIndexAsync` 处理缺失 index，这是既有 lifecycle 行为；本设计不新增 query-time mapping mutation / repair，也不引入 query-time replay 或 priming。
+5. 不背历史 index 兼容包袱。mapping 变更以 alias + augmented mapping fingerprint lifecycle 为唯一 schema-drift 权威；alias 指向非当前 fingerprint 的 physical index 时 fail loud，不做在线兼容、双轨 query fallback 或旧 index repair。
+6. 当前 `GetAsync` / `QueryAsync` 会调用 `EnsureIndexAsync` 处理缺失 index，这是既有 lifecycle 行为；本设计不新增 query-time ES actual mapping reader、query-time mapping mutation / repair，也不引入 query-time replay 或 priming。
 
 ## 3. 重构目标
 
@@ -78,7 +78,7 @@ owner: aevatar-core
 - 不新增 read model 类型。
 - 不引入第二套 projection envelope。
 - 不在 query path 中刷新 read model、修复 index mapping 或重建 index；缺失 index 的创建继续沿用当前 `EnsureIndexAsync` 行为。
-- 不自动修复已有 Elasticsearch index 的错误 mapping；需要时直接清空 / 重建 projection index。
+- 不自动修复已有 Elasticsearch index 的错误 mapping；需要时直接清空 / 重建 projection index，或通过显式运维迁移重建 alias 指向。
 - 不改变 native script materialization compiler 的 schema-driven mapping 语义。
 - 不为了兼容历史 index 增加双轨 query fallback。
 
@@ -86,7 +86,7 @@ owner: aevatar-core
 
 1. Elasticsearch mapping augmentation 必须位于 provider 边界，不得泄露到 Domain / Application / Host 业务编排层。
 2. 统一 helper 只能基于 protobuf read model contract 和显式 `DocumentIndexMetadata` 工作，不能依赖运行时 query 样本或已写入 document。
-3. query path 禁止执行 mapping mutation / repair、projection lifecycle 操作或 event replay；不得在查询路径补写 read model 数据。
+3. query path 禁止执行 live ES mapping 读取、mapping mutation / repair、projection lifecycle 操作或 event replay；不得在查询路径补写 read model 数据。
 4. 显式业务 mapping 优先；helper 不得覆盖 metadata provider 已声明的字段 mapping。
 5. provider-owned `ProjectionDocumentId` 必须保持 `keyword`；冲突时应启动期失败，而不是降级兼容。
 6. `google.protobuf.Any`、`google.protobuf.Struct`、map 字段、repeated complex message 默认不得递归展开为静态 mapping。
--- a/src/Aevatar.CQRS.Projection.Providers.Elasticsearch/README.md
+++ b/src/Aevatar.CQRS.Projection.Providers.Elasticsearch/README.md
@@ -32,8 +32,9 @@ Elasticsearch Document Provider。
 - 新建索引时，provider 会基于 read model 的 protobuf descriptor 补齐低风险稳定字段映射：root-level `google.protobuf.Timestamp` 映射为 `date`，root-level 稳定字符串标识字段（如 `id`、`actor_id`、`last_event_id`、`*_id`、`*_key`、`*_hash`、`*_status`、`*_kind`、`*_type`、`*_type_url`）映射为 `keyword`
 - `DocumentIndexMetadata` 中显式声明的 mapping 优先，provider 不覆盖自定义 `text`、analyzer、object、nested 或其他业务 mapping
 - `google.protobuf.Any`、`google.protobuf.Struct`、map、repeated message 与 repeated scalar 字段默认保持开放，不由通用 helper 递归展开
-- mapping 契约变更不兼容旧 Elasticsearch index 时，直接清空或重建 projection index；provider 不做旧索引在线修复、双读 fallback 或 query-time mapping repair
-- `AutoCreateIndex=true` 只会在缺失 index 时按当前契约创建新 index；如果需要保留数据，应通过 projection 重放或外部重建流程恢复数据
+- schema-drift 权威源只有 alias + augmented mapping fingerprint：alias 必须指向 `{alias}-v{fingerprint}` 物理索引
+- alias 指向不同 fingerprint 时视为配置错误，provider 在 projection 写入/读取入口 fail loud；不会读取 live ES mapping 作为第二真相，也不会在 query-time 做 mapping repair、双读 fallback 或自动 reindex
+- `AutoCreateIndex=true` 只会在缺失 index 或 legacy bare index 包装时按当前契约创建新 physical index；如果 drift 需要保留数据，应通过显式 projection 重放、外部重建或运维迁移流程恢复数据
 
 参考：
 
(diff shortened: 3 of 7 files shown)


loning:
## ✅ 共识卡片 — Phase 8 r1 consensus(2 approve + 1 comment,非阻塞)

| 维度 | 值 |
|---|---|
| PR | #1039 iter98 cluster-743: ES lifecycle schema-drift |
| Round | r1 |
| 角色 verdict | architect: **approve** · tests: **approve** · quality: **comment**(non-blocking) |
| 后续 | 等 CI 绿 → auto squash merge + close #743 |
| **是否需要人介入** | **❌ 否** |

⟦AI:AUTO-LOOP⟧

PR #1061, opened 2026-05-26
Tool calls: Ornn skills, n8n migration, and Aevatar core tools

eanzhao:
## Purpose

This branch is the work branch for Aevatar tool-call capabilities. The main goals are:

1. Make Aevatar proactively query Ornn for skills and invoke the matching skills.
2. Port selected n8n implementations into Aevatar.
3. Continue debugging Aevatar core capabilities exposed to the LLM as tools.

## Current Scope

- Adds runtime-enforced Ornn skill discovery and `use_skill` invocation for Lark slash-command flows such as `/daily` and `/goal`.
- Adds a `SkillRecoveryOrchestrator` / `SkillRecoveryPlanner` split so recovery can execute real tools instead of relying on prompt-only final-answer nudges.
- Preserves DeepSeek thinking-mode continuity by extracting OpenAI-compatible streaming `reasoning_content` from raw SSE patches and replaying it in assistant history before tool follow-up rounds.
- Adds reasoning replay for synthetic recovery-generated assistant tool-call messages before Ornn skill follow-up rounds.
- Keeps the branch open for the next tool-call work: n8n migration and Aevatar core capability tool injection.

## Production Finding

- Prod deployment `docker.io/aelfdevops/aevatar-console-backend:5106fe09` was confirmed in Kubernetes after the prior push.
- Prod logs showed `/daily` was no longer skipping Ornn: it called `skill-search`, loaded `chrono-ai-daily/json`, then searched again.
- The remaining failure was still DeepSeek thinking-mode continuity: HTTP 400 `The reasoning_content in the thinking mode must be passed back to the API.`
- The missing piece was the synthetic assistant tool-call messages generated by `SkillRecoveryOrchestrator`; they were not produced by the streaming provider path, so they did not receive the raw SSE `reasoning_content` extraction fix from the previous commit.
- Commit `fcd71b864` adds non-empty reasoning content to those synthetic recovery tool-call messages and covers it with tests.

## Verification

- `dotnet test test/Aevatar.AI.Tests/Aevatar.AI.Tests.csproj --filter "FullyQualifiedName~SkillRecoveryPlannerTests|FullyQualifiedName~AIComponentCoverageTests|FullyQualifiedName~ChatRuntimeStreamingBufferTests" --nologo` — passed, 63 tests.
- `dotnet test test/Aevatar.GAgents.ChannelRuntime.Tests/Aevatar.GAgents.ChannelRuntime.Tests.csproj --filter "FullyQualifiedName~ConversationReplyGeneratorTests" --nologo` — passed, 28 tests.
- `bash tools/ci/test_stability_guards.sh` — passed.
- `git diff --check` — passed.


src/platform/Aevatar.GAgentService.Application/Responses/LlmSessionCompletionObserver.cs
```diff
@@ -0,0 +1,39 @@
...
+        WaitForCompletionAsync(sessionQueryPort, responseId, DefaultTimeout, ct);
+
+    internal static async Task<LlmSessionCompletionSnapshot?> WaitForCompletionAsync(
+        ILlmSessionQueryPort sessionQueryPort,
+        string responseId,
+        TimeSpan timeout,
+        CancellationToken ct)
+    {
+        ArgumentNullException.ThrowIfNull(sessionQueryPort);
+        ArgumentException.ThrowIfNullOrWhiteSpace(responseId);
```
eanzhao, line 24: **[Consensus: 2 models]** severity=major, category=arch

Both reviewers flag the same architectural violation: the observer polls a read model with `Task.Delay(50ms)` up to 2s inside the request path waiting for committed completion to materialize. CLAUDE.md forbids query-time priming / replay — projection freshness should be guaranteed by the projection-scope activation lifecycle (already added in this PR for workflow/GAgent runs) or by an event-driven continuation, not by polling. Consider removing this observer in favour of the explicit activation contract, or moving the wait into an async background materializer.

<details>

- **mimo**: Polling a read model in a request path with `Task.Delay(50ms)` up to 2s violates CLAUDE.md: "禁止 query-time replay/priming". This waits for the committed completion to become visible in the read model. The race window is narrow and the timeout is bounded, but the pattern sets a precedent. Consider an event-driven notification (e.g. `Channel<T>` or `TaskCompletionSource` wired through the projection pipeline) instead of polling.
- **kimi**: This component introduces a synchronous `Task.Delay` polling loop inside a command/query facade to wait for read-model visibility. This violates AGENTS.md "禁止 query-time priming" and "read model 物化必须脱离 query 调用栈". Projection freshness should be guaranteed by the lease/activation contract (already added in this PR for workflow/GAgent runs) or by event-driven continuation, not by polling the read model from the request path. Consider removing this observer and relying on the explicit projection-scope activation lifecycle, or moving the wait to an async background materializer.
</details>

agents/Aevatar.GAgents.Channel.Runtime/Conversation/ConversationGAgent.cs
```diff
@@ -277,12 +279,16 @@ private async Task HandleInboundActivityCoreAsync(
...
             RestoreRuntimeTransportCredentials(runCopy.Activity, runtimeContext);
+            runCopy.PriorHistory.Clear();
+            runCopy.PriorHistory.AddRange(State.RetainedHistory.Select(entry => entry.Clone()));
             var persistedCopy = runCopy.Clone();
             persistedCopy.ReplyToken = string.Empty;
             persistedCopy.ReplyTokenExpiresAtUnixMs = 0;
             persistedCopy.Activity = CloneForDurableState(persistedCopy.Activity);
             persistedCopy.TargetRef = null;
             persistedCopy.LlmControl = null;
+            persistedCopy.PriorHistory.Clear();
```
eanzhao, line 290: **[codex]** severity=major, category=bug

`persistedCopy.PriorHistory.Clear()` drops the newly-added transcript window from the durable pending request. That means any deferred redispatch path (`HandleDeferredLlmReplyDispatchRequestedAsync`, process restart, dispatcher unavailable retry) will re-run the LLM turn without the retained conversation context, which defeats the feature you just added. Keep `PriorHistory` on the persisted copy and only scrub runtime-only credentials/tokens.

src/Aevatar.AI.ToolProviders.Lark/Tools/LarkApprovalsActTool.cs
```diff
@@ -63,6 +63,10 @@ protected override async Task<string> ExecuteAsync(Parameters parameters, Cancel
         if (!string.IsNullOrWhiteSpace(userIdType) && !AllowedUserIdTypes.Contains(userIdType))
             return LarkProxyResponseParser.Serialize(new { success = false, error = "user_id_type must be one of: user_id, union_id, open_id" });
 
+        var userId = ResolveUserId(parameters.UserId);
```
eanzhao, line 66: **[codex]** severity=major, category=bug

This now always prefers `channel.lark.operator_user_id` regardless of `user_id_type`. If the caller sets `user_id_type="open_id"` or `"union_id"`, the tool will still send the operator's `user_id`, which is the wrong identifier type and will cause Lark approval actions to fail. Resolve the operator ID from the matching metadata key (`channel.lark.operator_open_id` / `channel.lark.operator_union_id` / `channel.lark.operator_user_id`) based on `user_id_type`, then fall back to the explicit parameter only if that matching key is absent.

src/platform/Aevatar.GAgentService.Application/ScopeGAgents/GAgentDraftRunInteractionService.cs
```diff
@@ -61,19 +84,39 @@ public async Task<CommandInteractionResult<GAgentDraftRunAcceptedReceipt, GAgent
...
+            }
 
             return result;
         }
         catch
         {
-            await RollbackAsync(actor, CancellationToken.None);
+            if (!accepted)
+            {
+                await _observationScopeActivationPort.ReleaseAsync(activation, CancellationToken.None);
```
eanzhao, line 116: **[kimi]** severity=major, category=bug

If `_observationScopeActivationPort.ReleaseAsync` throws inside this catch block, `RollbackAsync` is skipped and the original exception is lost. This can leak created actors and obscure the root cause. Wrap each cleanup step in an individual try/catch (or use `ExceptionDispatchInfo`) so that rollback is attempted even when release fails, and the original exception is preserved.

src/workflow/Aevatar.Workflow.Application/Runs/WorkflowChatRunInteractionService.cs
```diff
@@ -0,0 +1,193 @@
...
+                return await ExecuteAttemptAsync(
+                    attempt.Value!,
+                    emitAsync,
+                    onAcceptedAsync,
+                    ct).ConfigureAwait(false);
+            }
+            catch (Exception ex) when (!attempt.Value!.Accepted &&
+                                       _fallbackPolicy.TryCreateFallbackCommand(currentRequest, ex, out var fallbackRequest))
+            {
+                await CleanupAttemptAsync(attempt.Value!, CancellationToken.None).ConfigureAwait(false);
```
eanzhao, line 67: **[kimi]** severity=major, category=bug

In the fallback catch block, if `CleanupAttemptAsync` throws, the original exception (`ex`) is lost because the filter has already returned true and the catch body is executing. This makes debugging production failures difficult. Consider wrapping cleanup in its own try/catch (logging any cleanup failure) so the original exception is always preserved and rethrown, or use `ExceptionDispatchInfo.Capture`.

agents/Aevatar.GAgents.Channel.Runtime/Conversation/ConversationGAgent.cs
```diff
@@ -338,6 +344,45 @@ private async Task HandleInboundActivityCoreAsync(
...
+        var durableContext = AgentToolExecutionContextMapper.FromPayload(request.ToolContext) with
+        {
+            Credentials = AgentToolCredentials.Empty,
+        };
+        request.ToolContext = HasDurableToolContext(durableContext)
+            ? durableContext.ToPayload()
+            : null;
+    }
+
+    private static bool HasDurableToolContext(AgentToolExecutionContext context) =>
```
eanzhao, line 361: **[mimo]** severity=minor, category=di

`HasDurableToolContext` manually enumerates every field of `AgentToolExecutionContext` with 24 `string.IsNullOrWhiteSpace` checks. When new fields are added to context sub-records (e.g. a future `AgentSkillRecoveryContext` field or new `AgentToolRoutingContext` field), this predicate will silently drop them. Consider adding a `bool HasContent` property on `AgentToolExecutionContext` itself, or using reflection, to avoid the maintenance drift risk.

src/Aevatar.AI.Core/Chat/SkillRecoveryPlanner.cs
```diff
@@ -0,0 +1,497 @@
...
+{
+    private const string OrnnSearchSkillsToolName = "ornn_search_skills";
+    private const string UseSkillToolName = "use_skill";
+
+    public readonly record struct RecoveryDirective(
+        ToolCall? ToolCall,
+        bool ConsumesOrnnSearchAttempt,
+        string? Nudge);
+
+    private static readonly string[] BlockerPhrases =
```
eanzhao, line 18: **[mimo]** severity=minor, category=design

`BlockerPhrases` includes extremely broad substrings like `"error"`, `"failed"`, `"missing"`, `"unavailable"`. Any tool result containing "the error rate is fine" or "we failed over successfully" will incorrectly trigger a blocker recovery round. The impact is bounded by `MaxOrnnSearchAttempts` (causes an unnecessary re-search, not data corruption), but consider narrowing to phrases that are more specific to actual failure states, or requiring the phrase to appear in a tool-result message (not the final LLM content) to reduce false positives.

PR #1830, opened 2026-06-05
Fix stream topology storage conflicts

louis4li:
## Summary
- Fix #1828 by removing the duplicate projection observation relay write from activation service startup.
- Keep relay/topology mutation owned by the projection scope actor command/lifecycle path instead of masking the conflict with stream-topology storage retry.
- Remove the previously added stream topology retry tests and keep focused coverage that activation does not write relays directly.

## Root cause
The version conflict was caused by two lifecycle paths writing the same stream-topology grain for hot roots like `agent-registry-store`:
1. `ProjectionScopeActivationService.EnsureAsync()` created the projection scope and directly upserted the observation relay.
2. The projection scope actor then handled `EnsureProjectionScopeCommand` and upserted the same relay again as part of its own lifecycle.

This violated the actor/lifecycle ownership boundary and increased stale ETag writes against Orleans/Garnet storage. The fix makes activation only ensure the actor and dispatch the command; the actor owns relay attachment.

## Test plan
- [x] `dotnet test test/Aevatar.CQRS.Projection.Core.Tests/Aevatar.CQRS.Projection.Core.Tests.csproj --nologo --no-restore`
- [x] `dotnet test test/Aevatar.Foundation.Runtime.Hosting.Tests/Aevatar.Foundation.Runtime.Hosting.Tests.csproj --nologo --no-restore`
- [x] `git diff --check -- src/Aevatar.Foundation.Runtime.Implementations.Orleans.Streaming/Streaming/Topology/StreamTopologyGrain.cs test/Aevatar.Foundation.Runtime.Hosting.Tests/OrleansDistributedCoverageTests.cs src/Aevatar.CQRS.Projection.Core/Orchestration/ProjectionScopeActivationService.cs test/Aevatar.CQRS.Projection.Core.Tests/ProjectionRuntimeRegistrationTests.cs`


src/Aevatar.CQRS.Projection.Core/Orchestration/ProjectionActivationPlanDispatcher.cs
```diff
@@ -33,6 +40,41 @@ public Task DispatchAsync(ProjectionActivationPlan plan, CancellationToken ct =
...
+        var scopeKey = new ProjectionRuntimeScopeKey(
+            plan.StartRequest.RootActorId,
+            plan.StartRequest.ProjectionKind,
+            plan.StartRequest.Mode,
+            plan.StartRequest.SessionId);
+        var targetScopeActorId = ProjectionScopeActorId.Build(scopeKey);
+        var envelope = new EventEnvelope
+        {
+            Id = Guid.NewGuid().ToString("N"),
+            Timestamp = Timestamp.FromDateTime(DateTime.UtcNow),
```
eanzhao: **[Consensus: 2 models]** severity=major, category=bug

The new 3-arg `DispatchAsync` builds a **synthetic** observation envelope (fresh `Id = Guid.NewGuid()`, no `Propagation`) and dispatches it directly to the scope actor — but this is **not equivalent** to the real committed-state publication, and the observation relay is still attached by the scope actor itself (`ProjectionScopeGAgentBase.HandleEnsureAsync` line 61), so both delivery paths are live. Two verified consequences:

1. **Silently dropped for propagation-keyed projectors.** `GAgentRunTerminalProjector.ProjectAsync` (lines 51–56) returns early unless `envelope.Propagation.CorrelationId == context.CorrelationId`. This envelope carries no `Propagation`, so the bootstrap observation is discarded for run-terminal scopes — the draft-run current-state then materializes only via the eventually-consistent relay, which is exactly why this PR has to switch `ScopeDraftRunWorkflowRunCurrentStateIntegrationTests` to polling and add it to the polling allowlist.
2. **Double-processed for propagation-agnostic projectors.** `ProjectionMaterializationScopeGAgentBase.ProcessObservationCoreAsync` has no version/event-id dedup (it always runs materializers and advances the watermark), and `HandleObservedEnvelopeAsync` filters only by route. So the same `StateEvent` arriving via both the direct forward **and** the relay is materialized twice and persists two `ProjectionScopeWatermarkAdvancedEvent`s — redundant work and event-store writes on the very path this PR is meant to de-conflict.

**Fix:** build the bootstrap envelope with the same propagation/dedup origin as the real publication (copy `context.SourceEnvelope?.Propagation`), or drop the synthetic second path and rely on the relay delivering the triggering event exactly once. (deepseek separately noted the forwarding also couples this dispatcher to stream-topology internals — `StreamForwardingRules` / `ProjectionScopeActorId` / `EnvelopeRouteSemantics` — and would read better extracted into the activation hook or a dedicated forwarding port.)

<details>

- **codex**: This synthetic envelope drops the propagation from `context.SourceEnvelope` and gives the same committed state event a fresh envelope id. `GAgentRunTerminalPlans` derives the scope `SessionId` from `context.SourceEnvelope.Propagation.CorrelationId`, while `GAgentRunTerminalProjector` later requires `envelope.Propagation.CorrelationId` to match that session, so the bootstrap observation can be skipped; if the normal relay also forwards the real publish, the different envelope id also defeats runtime dedup and can process the same `StateEvent` twice.
- **kimi**: Direct-dispatching the triggering `CommittedStateEventPublished` to the scope actor duplicates delivery whenever the scope is already active: the existing observation relay (still upserted by the actor in `HandleEnsureAsync`) will forward the same event via the stream. The scope actor does not dedupe by `(sourceActorId, eventId)`, so the same committed event will be processed twice. Either gate the direct forward to scopes actually created by this activation, or add idempotent deduplication inside `HandleObservedEnvelopeAsync`.
</details>

src/Aevatar.CQRS.Projection.Core/Orchestration/ProjectionActivationPlanDispatcher.cs
```diff
@@ -16,10 +21,12 @@ public sealed class ProjectionActivationPlanDispatcher
         ?? throw new MissingMethodException(nameof(ProjectionActivationPlanDispatcher), nameof(DispatchCoreAsync));
 
     private readonly IServiceProvider _services;
+    private readonly IActorDispatchPort? _dispatchPort;
 
     public ProjectionActivationPlanDispatcher(IServiceProvider services)
     {
         _services = services ?? throw new ArgumentNullException(nameof(services));
+        _dispatchPort = services.GetService<IActorDispatchPort>();
```
eanzhao: **[Consensus: 2 models]** severity=minor, category=di

`_dispatchPort = services.GetService<IActorDispatchPort>()` resolves a dependency from the container inside the constructor — a service-locator pattern (CLAUDE.md: 不用 Service Locator 隐藏依赖——依赖一律构造注入、显式可见). Prefer an explicit constructor parameter so the dependency is visible and lifetime-checked. The likely motive is *optional* resolution: a non-nullable ctor param would make every `TryAddSingleton<ProjectionActivationPlanDispatcher>()` site throw when `IActorDispatchPort` isn't registered. If optionality is intended, register via a factory that keeps the dependency explicit: `TryAddSingleton(sp => new ProjectionActivationPlanDispatcher(sp, sp.GetService<IActorDispatchPort>()))`.

<details>

- **kimi**: `services.GetService<IActorDispatchPort>()` deepens the service-locator pattern in a singleton. Prefer adding `IActorDispatchPort? dispatchPort = null` as an explicit constructor parameter so the dependency is visible and lifetime-checked by the container.
- **mimo**: Resolving it via `services.GetService<IActorDispatchPort>()` here is a service-locator anti-pattern that CLAUDE.md prohibits. Consider adding `IActorDispatchPort?` as a constructor parameter and updating the `TryAddSingleton` registrations to a factory.
</details>

src/Aevatar.CQRS.Projection.Core/Orchestration/ProjectionScopeActivationService.cs
```diff
@@ -54,7 +54,6 @@ public async Task<TLease> EnsureAsync(
             request.SessionId);
 
         await _scopeRuntime.EnsureExistsAsync(scopeKey, ct).ConfigureAwait(false);
```
eanzhao, line 60: **[kimi]** severity=minor, category=arch

Removing the `_scopeRuntime.EnsureObservationRelayAsync(scopeKey, ct)` call (the deleted line just below) leaves `ProjectionScopeActorRuntime<TScopeAgent>.EnsureObservationRelayAsync` — and its only consumer, the `_streams` field plus the `IStreamProvider? streams` ctor param — with no callers. Per 删除优先, delete the dead method/field/param and drop the now-unused stream-provider argument at the `new ProjectionScopeActorRuntime<TScopeAgent>(...)` construction site, rather than leaving a dead relay path. (`_scopeRuntime` is a concrete type, not an interface, so this is an internal cleanup with no contract impact.)

src/Aevatar.CQRS.Projection.Core/Orchestration/ProjectionActivationPlanDispatcher.cs
```diff
@@ -33,6 +40,41 @@ public Task DispatchAsync(ProjectionActivationPlan plan, CancellationToken ct =
...
+            plan.StartRequest.RootActorId,
+            plan.StartRequest.ProjectionKind,
+            plan.StartRequest.Mode,
+            plan.StartRequest.SessionId);
+        var targetScopeActorId = ProjectionScopeActorId.Build(scopeKey);
+        var envelope = new EventEnvelope
+        {
+            Id = Guid.NewGuid().ToString("N"),
+            Timestamp = Timestamp.FromDateTime(DateTime.UtcNow),
+            Payload = Any.Pack(context.Published.Clone()),
```
eanzhao: **[v4-pro]** severity=minor, category=perf

`Any.Pack(context.Published.Clone())` — `Any.Pack` already serializes the message into the `Any`'s byte buffer, so the `.Clone()` deep-copy is a redundant throwaway allocation on every committed-state publication (a hot path). Use `Any.Pack(context.Published)`. (If the envelope is reworked per the major comment on line 64, this line goes away anyway.)

src/Aevatar.CQRS.Projection.Core/Orchestration/ProjectionActivationPlanDispatcher.cs
louis4li: Fixed in 0dc9a0517.

I removed the synthetic committed-observation direct dispatch from `ProjectionActivationPlanDispatcher`. Activation now only ensures the projection scope before the normal committed-state publication proceeds, so the committed fact is delivered through the existing publication/relay path once, with the normal propagation/origin semantics instead of a freshly-created envelope.

Verification run:
- `git diff --check`
- `bash tools/ci/projection_state_version_guard.sh`
- `bash tools/ci/projection_state_mirror_current_state_guard.sh`
- `bash tools/ci/projection_route_mapping_guard.sh`
- `dotnet test test/Aevatar.CQRS.Projection.Core.Tests/Aevatar.CQRS.Projection.Core.Tests.csproj --nologo --filter "FullyQualifiedName~CommittedStateProjectionActivationHookTests|FullyQualifiedName~ProjectionRuntimeRegistrationTests|FullyQualifiedName~ProjectionScopeGAgentBaseTests|FullyQualifiedName~ProjectionScopeActorRuntimeTests"`
- `dotnet test test/Aevatar.Scripting.Core.Tests/Aevatar.Scripting.Core.Tests.csproj --nologo --filter FullyQualifiedName~ScriptingCommittedStateProjectionActivationPlanProviderTests`
- `dotnet test test/Aevatar.GAgentService.Integration.Tests/Aevatar.GAgentService.Integration.Tests.csproj --nologo --filter FullyQualifiedName~ScopeDraftRunWorkflowActorCurrentStateIntegrationTests`

src/Aevatar.CQRS.Projection.Core/Orchestration/ProjectionActivationPlanDispatcher.cs
louis4li: Fixed in 0dc9a0517.

The optional `IActorDispatchPort` service-locator dependency was removed together with the synthetic forwarding path. `ProjectionActivationPlanDispatcher` no longer resolves or stores `IActorDispatchPort`; it only dispatches activation plans to the existing activation service path.

src/Aevatar.CQRS.Projection.Core/Orchestration/ProjectionScopeActivationService.cs
louis4li: Fixed in 0dc9a0517.

Deleted the dead `ProjectionScopeActorRuntime.EnsureObservationRelayAsync` helper, removed its `_streams` field / constructor parameter, and dropped the `streams: sp.GetService<IStreamProvider>()` plumbing from the runtime registrations. The actor-owned relay attach/remove path remains in `ProjectionScopeGAgentBase`, where it is still used by scope activation/ensure/release.

src/Aevatar.CQRS.Projection.Core/Orchestration/ProjectionActivationPlanDispatcher.cs
louis4li: Fixed in 0dc9a0517.

The `Any.Pack(context.Published.Clone())` allocation site was removed by deleting the synthetic envelope construction entirely. The triggering committed fact now goes through the normal committed-state publication path rather than being repacked into a second envelope.

PR #1867, opened 2026-06-08
Improve team member entry actions

potter-sun:
## What changed

- Added direct Team member handoffs from the Teams homepage so users can edit the entry member, edit the only member, create a member, or manage members based on roster state.
- Added Create member access from the Team Members tab and polished member action buttons, including right-aligned table actions.
- Unified the Teams list/card bottom actions into a rounded text action group with separators while preserving Team detail and members-tab navigation.
- Added locale strings and focused tests for the new entry paths.

## Why

The old flow forced users through Team detail and the Members tab before they could create or edit members. The new entry points keep Team detail available while making common member work reachable in one step.

## Validation

- pnpm --dir apps/aevatar-console-web jest src/pages/teams/home.test.tsx --runInBand
- pnpm --dir apps/aevatar-console-web jest src/pages/teams/detail.test.tsx --runInBand
- pnpm --dir apps/aevatar-console-web tsc
- git diff --check


apps/aevatar-console-web/src/pages/teams/home.tsx
```diff
@@ -564,6 +583,57 @@ function buildTeamRosterPreview(input: {
...
+              mode: "create-member",
+              returnTo: teamHomeHref,
+              scopeId: input.scopeId,
+              teamId: input.team.teamId,
+            }),
+            kind: "create-member",
+            label: t("teams.home.actions.createMember", "Create member"),
+          }
+        : {
+            href: buildTeamDetailHref({
```
AbigailDeng, line 629: Non-blocking question: in the multi-member/no-entry fallback, `memberQuickAction.href` points to the same members-tab route as `membersHref` below, so the action group can show both “Manage members” and “View members” with identical navigation. That may be intentional for this roster state, but if the first CTA is meant to be a distinct quick action, consider either hiding the duplicate “View members” button for this branch or making the primary label/navigation more clearly different.

PR #1934, opened 2026-06-10
Fix workflow draft id resolution

potter-sun:
## Summary

- Fix Studio workflow-member routing so display names are resolved to stable workflow draft ids before loading drafts.
- Stop treating member display names such as `Untitled member` as direct `/api/workspace/workflow-drafts/{workflowId}` path ids.
- Add a regression test for scoped workflow members whose display name differs from the normalized draft id.

## Root Cause

The Studio member-first flow could fall back from a workflow member display name directly to a workflow draft id. That made the frontend request paths like `/api/workspace/workflow-drafts/Untitled%20member`, while the backend stores the stable draft id as a normalized value such as `untitled-member`.

## Validation

- `pnpm --dir apps/aevatar-console-web test src/pages/studio/index.test.tsx --runInBand`
- `pnpm --dir apps/aevatar-console-web build`
- `git diff --check`

### DIFF
--- a/apps/aevatar-console-web/src/pages/studio/index.test.tsx
+++ b/apps/aevatar-console-web/src/pages/studio/index.test.tsx
@@ -3600,6 +3600,68 @@ describe("StudioPage", () => {
     expect(studioApi.getMember).toHaveBeenCalledWith("scope-1", "workspace-demo");
   });
 
+  it("resolves workflow member display names to stable draft ids before loading drafts", async () => {
+    mockStudioMembers = [
+      {
+        ...mockStudioMembers[0],
+        memberId: "untitled-member",
+        displayName: "Untitled member",
+        publishedServiceId: "",
+        lastBoundRevisionId: null,
+      },
+    ];
+    mockWorkflowSummaries = [
+      {
+        ...mockWorkflowSummaries[0],
+        workflowId: "untitled-member",
+        name: "Untitled member",
+        fileName: "untitled-member.yaml",
+        filePath: "scope://scope-1/untitled-member.yaml",
+      },
+    ];
+    mockWorkflowFile = {
+      ...mockWorkflowFile,
+      workflowId: "untitled-member",
+      name: "Untitled member",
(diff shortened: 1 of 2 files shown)


AbigailDeng (changes requested):
# Review Summary

Overall:
NEEDS_CHANGES

## P1

The new summary-only workflow-member resolution can return the backend `memberId` as a workflow id before it tries to match the member `displayName` against the workflow list.

In `resolveWorkflowIdForMemberSummary`, the code passes only `memberId` and `displayName` into `resolveWorkflowIdFromMemberWorkflowReference`. Inside that helper, the `memberId` branch uses `allowDirectIdFallback: true`, so a member like `m-1` can resolve to workflow id `m-1` even when the actual draft is represented by a workflow summary such as `{ workflowId: "wf-1", name: "Workflow One" }`.

That identity split is supported by the backend contract: `StudioMemberImplementationRefResponse.WorkflowId` is separate from `MemberId`, and existing backend tests cover member/detail cases like `memberId = m-1` with workflow ref `wf-1` / `workflow-main`. With this PR, summary-only paths such as lifecycle Build navigation and unpublished member rail clicks can request the wrong draft, for example `/api/workspace/workflow-drafts/m-1`, or fall back away from the intended workflow.

Suggested fix: for summary-only resolution, try known workflow matches before direct fallback. A safer order is:

1. Try `memberId` against `visibleWorkflowSummaries` with `allowDirectIdFallback: false`.
2. Try `displayName` against `visibleWorkflowSummaries` with `allowDirectIdFallback: false`.
3. Only then allow direct fallback to `memberId`, not to `displayName`.

Please also add a regression test where `memberId = "m-1"`, `displayName = "Workflow One"`, and the workflow summary is `{ workflowId: "wf-1", name: "Workflow One" }`; assert the page calls `getWorkflow("wf-1", scopeId)`, never `getWorkflow("m-1", scopeId)`.

## Merge Recommendation

Changes Required

PR #1965, opened 2026-06-10
Fix Teams runtime run sampling

AbigailDeng:
## 问题

Teams Home 和 Team Detail 会把 Team member id / roster 上的 member-like `publishedServiceId` 用来请求 runtime runs。部分场景会打到 member 私有接口或不存在的 service runtime，例如：

- `/api/scopes/{scopeId}/members/{memberId}/runs?take=1` 返回 `MEMBER_ACCESS_DENIED`
- `/api/scopes/{scopeId}/services/member-untitled-member1/runs?take=12` 因为 service id 不存在而 500

## 方案

- Teams Home 改为按成员的 `publishedServiceId` 聚合 service-scoped runs，不再批量请求 member-scoped runs。
- Team Detail 的 runtime lens 只在 service catalog 中确认 service id 存在后，才请求 service revisions / service runs。
- 移除 Team Detail 对 `listMemberRuns` 的 fallback，避免页面用 member 私有运行接口作为团队运行态来源。
- 增加回归测试覆盖 member-like service id 缺失时不请求 runs/revisions 的场景。

## 影响路径

- `apps/aevatar-console-web/src/pages/teams/home.tsx`
- `apps/aevatar-console-web/src/pages/teams/runtime/useTeamRuntimeLens.ts`
- `apps/aevatar-console-web/src/pages/teams/home.test.tsx`
- `apps/aevatar-console-web/src/pages/teams/detail.test.tsx`

## 验证

- `npm test -- src/pages/teams/home.test.tsx src/pages/teams/detail.test.tsx --runInBand`：71 passed
- `npm run tsc`：passed
- `bash tools/ci/test_stability_guards.sh`：passed
- `git diff --check`：passed

## 备注

之前误提到 `dev` 的 PR #1962 已关闭；这个 PR 的 base 是 `feat/2026-06-08_workflow-studio`。

PR #1967, opened 2026-06-10
Fix workflow member invoke identity

AbigailDeng:
## Summary
- add the Team workflow member invoke page at `/teams/:scopeId/:teamId/members/:memberId/invoke`
- wire workflow-type team members from Team detail/member surfaces into the new invoke route
- reuse the Studio member invoke panel for member-target workflow execution and show run output/log evidence in the invoke experience
- keep frontend `memberId` identity separate from workflow draft `workflowId` when creating and loading workflow members

## Scope
- Frontend only: this PR changes `apps/aevatar-console-web/...` files.
- No backend/API implementation changes are included. The frontend now avoids passing or guessing workflow-derived member ids; it relies on the remote backend to return the authoritative member id.

## Validation
- `npm --prefix apps/aevatar-console-web run jest -- src/pages/team-member-workflow-studio/index.test.tsx src/shared/navigation/teamRoutes.test.ts src/shared/studio/api.test.ts --runInBand`
- `npm --prefix apps/aevatar-console-web run jest -- src/pages/team-member-invoke/index.test.tsx src/pages/teams/detail.test.tsx src/routesConfig.test.ts src/app.navigation.test.ts --runInBand`

## Notes
- This PR is split from #1965 because #1965 uses a different head branch.


apps/aevatar-console-web/src/pages/studio/index.tsx
```diff
@@ -5749,7 +5749,7 @@ const StudioPage: React.FC = () => {
           setInventoryBusyKey('create');
           setInventoryBusyAction('create');
           try {
-            createdScriptMember = await studioApi.createMember({
+            createdScriptMember = await studioApi.createMemberWithId({
```
potter-sun, line 5752: This call now uses `createMemberWithId`, but the Studio page Jest mock still only defines `createMember` (`src/pages/studio/index.test.tsx`). The script-member create path throws before it can open the script build panel, which is why CI currently fails at `StudioPage › creates a named Script member authority and opens its draft before bind`. Please add the new helper to the mock, or alias it to the existing mock implementation, and update the expectations to assert `createMemberWithId` for explicit-id member creation.

apps/aevatar-console-web/src/pages/team-member-invoke/index.tsx
```diff
@@ -0,0 +1,416 @@
...
+  padding: "10px 12px",
+};
+
+const invokeStageStyle: React.CSSProperties = {
+  minHeight: 520,
+};
+
+const TeamMemberInvokePage: React.FC = () => {
+  const { token } = theme.useToken();
+  const route = React.useMemo(() => readTeamMemberInvokeRouteState(), []);
```
potter-sun: `route` is memoized with an empty dependency list from `window.location.pathname`, so this page never re-reads the member identity after the component mounts. Umi/React Router can reuse the same page instance when navigating from `/teams/scope/team/members/member-a/invoke` to `/teams/scope/team/members/member-b/invoke`, leaving the queries, back links, displayed member, and `StudioMemberInvokePanel` pointed at the old member. Since this page is explicitly fixing stable member invoke identity, it needs to derive the route from router params/location, or subscribe to location changes the way other pages in this app do, and cover member-to-member navigation with a test.

potter-sun (changes requested):
Reviewed the identity-routing changes. I found two blocking issues: the console-web test suite is red because the Studio test mock was not updated for createMemberWithId, and the new member invoke page snapshots the route only once, so client-side navigation between member invoke URLs can keep invoking the previous member.

PR #1968, opened 2026-06-10
Unify scheduling on ScheduleGAgent

louis4li:
## Summary
- Move workflow-chat schedule lifecycle onto the actor-owned `ScheduleGAgent` scheduling path and expose the unified `/schedules` API.
- Remove the legacy `/workflow-schedules` route and workflow schedule store/poller facade from production registration.
- Preserve accepted-only schedule command semantics and keep read/query coverage through the existing schedule readmodel path.

## Test plan
- [x] `dotnet test test/Aevatar.Workflow.Core.Tests/Aevatar.Workflow.Core.Tests.csproj --filter ScheduleGAgentTests --nologo`
- [x] `dotnet test test/Aevatar.GAgentService.Integration.Tests/Aevatar.GAgentService.Integration.Tests.csproj --filter ScheduledDispatchEndpointsTests --nologo`
- [x] `dotnet test test/Aevatar.GAgentService.Tests/Aevatar.GAgentService.Tests.csproj --filter "ScheduledDispatchApplicationServiceTests|ScheduledDispatchServiceInvocationTests" --nologo`
- [x] `dotnet test test/Aevatar.Workflow.Host.Api.Tests/Aevatar.Workflow.Host.Api.Tests.csproj --filter "WorkflowInfrastructureCoverageTests|WorkflowScheduleProjectionTests" --nologo`
- [x] `dotnet build aevatar.slnx --nologo`

Closes #1963


src/workflow/Aevatar.Workflow.Infrastructure/CapabilityApi/ChatEndpoints.cs
```diff
@@ -26,8 +25,6 @@ public static IEndpointRouteBuilder MapWorkflowCapabilityEndpoints(this IEndpoin
     {
         var group = app.MapGroup("/api").WithTags("Chat");
         ChatQueryEndpoints.Map(group);
-        if (HasWorkflowScheduleDependencies(app.ServiceProvider))
-            WorkflowScheduleEndpoints.Map(group);
 
         return app;
```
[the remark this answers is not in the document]
louis4li: 确认，这个点有效：workflow host 只注册 workflow scheduling capability 时，删除旧的 conditional map 会导致替代的 `/api/schedules` 没有被暴露。

我会把 workflow host 路径补到新的 scheduled dispatch endpoint 映射上，并确保不会重新引入旧的双轨 schedule endpoint。

⟦AI:AUTO-LOOP⟧

src/platform/Aevatar.GAgentService.Hosting/Endpoints/Schedules/ScheduledDispatchEndpoints.cs
```diff
@@ -207,8 +208,7 @@ public sealed record ScheduledDispatchConfigurationHttpRequest
     public string? Timezone { get; init; }
     public bool Enabled { get; init; } = true;
     public IReadOnlyDictionary<string, string>? Headers { get; init; }
-    public ScheduledDispatchEnvelopeTargetHttpRequest? Envelope { get; init; }
-    public ScheduledDispatchServiceInvocationTargetHttpRequest? ServiceInvocation { get; init; }
+    public ScheduledWorkflowChatTargetHttpRequest? WorkflowChatTarget { get; init; }
 
     public ScheduledDispatchConfiguration ToConfiguration(string? fallbackScheduleId) =>
         new(
```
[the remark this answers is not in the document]
louis4li: 确认，这个点有效：workflow-chat schedule 现在应显式记录为 `ScheduleKind.Workflow`，否则后续 projection/query 的 workflow-kind filter 会漏掉这些记录。

我会在 workflow-chat request 到 configuration 的映射里补上 workflow kind，并补对应测试覆盖。

⟦AI:AUTO-LOOP⟧

src/platform/Aevatar.GAgentService.Hosting/Endpoints/Schedules/ScheduledDispatchEndpoints.cs
```diff
@@ -218,52 +240,54 @@ public ScheduledDispatchConfiguration ToConfiguration(string? fallbackScheduleId
...
+public sealed record ScheduledWorkflowChatTargetHttpRequest
 {
     public required ServiceIdentity Identity { get; init; }
-    public required string EndpointId { get; init; }
-    public required Any Payload { get; init; }
+    public required string Prompt { get; init; }
+    public string? SessionId { get; init; }
     public string? RevisionId { get; init; }
     public ServiceInvocationCaller? Caller { get; init; }
+    public LLMControlContextPayload? LlmControl { get; init; }
```
eanzhao: **[P1] Caller-supplied raw credentials get persisted into durable schedule state.**

This newly exposes `LlmControl` on the public schedule API and clones it verbatim into the packed `ChatRequestEvent` (line 275). `LLMControlContextPayload` carries raw bearer tokens (`nyx_id_access_token` / `nyx_id_org_token` / `sender_nyx_id_access_token`, ai_messages.proto:44-46), and `NyxIdLLMProvider.ResolveAccessToken` (NyxIdLLMProvider.cs:355) falls back to `LlmControl.NyxIdAccessToken` as the NyxID bearer.

Net effect: a raw credential is persisted into the schedule actor's state + committed event store (at rest, forever) and replayed on every fire — impersonation of the token owner for the token's TTL on each fire. `ScheduledDispatchTargetPreparationService` does no stripping. The deleted /workflow-schedules path never exposed this; it only had the typed `Auth` credential-source, which is exchanged server-side at fire time and is the intended channel for exactly this.

Suggestion: drop `LlmControl` from the HTTP target (or whitelist non-credential fields like model override only), and have target preparation strip token fields before anything is persisted.

src/platform/Aevatar.GAgentService.Projection/Queries/ScheduledDispatchQueryPort.cs
```diff
@@ -43,14 +43,21 @@ public async Task<ScheduledDispatchListResult> ListAsync(
...
             result.NextCursor,
             result.TotalCount);
     }
 
     private static ProjectionDocumentFilter[] BuildFilters(ScheduledDispatchListQuery query)
     {
         var filters = new List<ProjectionDocumentFilter>();
+        filters.Add(new ProjectionDocumentFilter
+        {
+            FieldPath = nameof(ScheduledDispatchDocument.Deleted),
```
eanzhao: **[P1] On Elasticsearch this filter hides every schedule document indexed before this deploys.**

`Eq` translates to a `term` query (ElasticsearchProjectionDocumentStorePayloadSupport.cs:144), and `term` does not match documents that lack the field. Pre-existing `ScheduledDispatchDocument`s have no `deleted` field, so LIST silently drops all of them (the actors keep firing — they just vanish from the API) until each actor happens to commit a new event and re-project.

The InMemory store evaluates filters against the deserialized document where proto3 defaults `deleted` to false, so tests stay green; only the ES provider regresses.

Suggestion: express the filter as "not (deleted == true)" so missing counts as not-deleted, or pair the rollout with a one-shot re-projection of the `scheduled-dispatches` index. A test with a `Deleted=true` document would also give the read-side filtering its first coverage.

src/platform/Aevatar.GAgentService.Application/Schedules/ScheduledDispatchApplicationService.cs
```diff
@@ -80,6 +80,17 @@ public async Task<ScheduledDispatchMutationReceipt> DisableAsync(
         return CreateMutationReceipt(normalizedScheduleId, actorId, admission);
     }
 
+    public async Task<ScheduledDispatchMutationReceipt> DeleteAsync(
```
eanzhao, line 87: **[P2] Doomed mutations are accepted with 202 and fail invisibly.**

`ScheduledDispatchConflictException` is defined but never thrown anywhere in src/, so the endpoints' 409 mapping is dead code. Combined with async-ack dispatch: create on an existing id, create on a deleted (tombstoned) id, and update/enable/disable/run-now/delete on a deleted id all return 202 — resolve only checks the runtime actor directory, and a tombstoned actor still exists — then throw inside the actor where the caller can't see it. Since GET/LIST also hide deleted documents, the failure is unobservable end to end.

202 = accepted-not-committed is fine per the ACK rules, but accepting commands that are statically guaranteed to fail makes the receipt meaningless. Suggestion: consult the readmodel (exists / deleted) here and throw NotFound/Conflict before dispatch for the statically-doomed cases.

src/platform/Aevatar.GAgentService.Core/Schedules/ScheduleGAgent.cs
```diff
@@ -112,6 +113,8 @@ private async Task HandleConfigureAsync(
         bool isCreate)
     {
         ArgumentNullException.ThrowIfNull(command);
+        if (State.Deleted)
```
eanzhao, line 118: **[P2] Delete permanently burns the scheduleId — is that intended?**

After `ScheduledDispatchDeletedEvent`, this guard blocks every future configure (create included), the actor and its event stream stay around forever, and the API answers 202 to the doomed create (see app-service comment) while GET 404s — so the id is unusable and the reason is invisible. If irreversible tombstone is the intended semantic, it should be documented on the DELETE endpoint and surfaced as a 409 on create; otherwise consider letting create resurrect a deleted id with fresh config. Either way this is the kind of resource-lifecycle decision the repo rules ask to make explicit (ownership / reuse / cleanup).

src/platform/Aevatar.GAgentService.Core/Schedules/ScheduleGAgent.cs
```diff
@@ -548,6 +564,7 @@ private ScheduledDispatchTargetKindState ResolveTargetKind() =>
             : ScheduledDispatchTargetKindState.Envelope;
 
     private bool IsConfigured() =>
+        !State.Deleted &&
```
eanzhao, line 575: **[P2] Late-fire vs delete race now throws instead of no-opping.**

A disabled schedule handles a stale non-manual fire gracefully via the `!command.Manual && !State.Enabled` early-return. Adding `!State.Deleted` to `IsConfigured()` means a deleted schedule hits `EnsureConfiguredForWrite("fire")` and throws before reaching that branch — an in-flight durable callback racing the delete (lease cancel runs after the event persist and can itself fail) surfaces as actor errors + runtime retry noise instead of a clean no-op.

Suggestion: treat `State.Deleted` like the disabled branch for non-manual fires (logged no-op).

src/platform/Aevatar.GAgentService.Core/Schedules/ScheduleGAgent.cs
```diff
@@ -11,14 +11,14 @@
 
 namespace Aevatar.GAgentService.Core.Schedules;
 
-public sealed class ScheduledDispatchGAgent : GAgentBase<ScheduledDispatchState>
+public sealed class ScheduleGAgent : GAgentBase<ScheduledDispatchState>
```
eanzhao, line 16: **[P2] Class rename has no migration story for already-persisted actors.**

On dev, `OrleansActorRuntime.CreateAsync` persists the CLR `AssemblyQualifiedName` via `InitializeAgentAsync`, and this class carries no `[GAgent]` stable kind. After the rename, every already-persisted schedule actor still stores `...ScheduledDispatchGAgent` as its identity and won't resolve to the new type on re-activation — existing schedules break on their next fire after deploy.

Suggestion: pin the stable kind as part of this rename — `[GAgent("gagent.service.scheduled-dispatch")]` — and wire the legacy CLR-name mapping (`ILegacyAgentClrTypeResolver` exists for exactly this) for actors created before the pin. That exact kind string is already what feature/integrate pins for this actor (952abf141), so using the same token here also makes the eventual dev -> integrate merge converge instead of conflict.

src/workflow/Aevatar.Workflow.Host.Api/Program.cs
```diff
@@ -23,7 +23,7 @@
         options.EnableWebSockets = true;
     });
 builder.AddAevatarPlatform();
-builder.Services.AddScheduledDispatchCapability(builder.Configuration);
+builder.AddGAgentServiceCapabilityBundle();
```
eanzhao: **[P3] Full capability bundle expands the workflow host surface beyond the PR's intent.**

Swapping `AddScheduledDispatchCapability` for `AddGAgentServiceCapabilityBundle` exposes the whole platform surface on the standalone workflow host (`/api/services`, `/api/scopes/{scopeId}/binding|workflows|scripts`, ...) and also registers `GAgentServiceDemoBootstrapHostedService`, which auto-creates demo services on startup in Development. The PR goal is "expose the unified /schedules API", and the narrow capability added in `AevatarPlatformHostBuilderExtensions` (`AddScheduledDispatchCapability` + `MapScheduledDispatchEndpoints`) already does exactly that. Suggestion: use the narrow capability here too.

src/platform/Aevatar.GAgentService.Hosting/Endpoints/Schedules/ScheduledDispatchEndpoints.cs
```diff
@@ -13,38 +14,43 @@ public static class ScheduledDispatchEndpoints
...
             .Produces(StatusCodes.Status400BadRequest);
-        group.MapPut("/scheduled-dispatches/{scheduleId}", Update)
-            .WithTags("Scheduled dispatches")
+        group.MapPut("/schedules/{scheduleId}", Update)
+            .WithTags("Schedules")
             .Produces<ScheduledDispatchMutationReceipt>(StatusCodes.Status202Accepted)
             .Produces(StatusCodes.Status400BadRequest);
-        group.MapPost("/scheduled-dispatches/{scheduleId}/enable", Enable)
-            .WithTags("Scheduled dispatches")
+        group.MapPost("/schedules/{scheduleId}:enable", Enable)
```
eanzhao, line 25: **[P3] `:action` route style collides with the schedule-id charset.**

`ScheduleIdAllowedCharacters` still includes `:` (ScheduledDispatchApplicationService.cs:8, ScheduledDispatchActorId.cs:6), so an id like `nightly:enable` is legal while the routes use `{scheduleId}:enable` — ambiguous for humans, and some proxies/gateways normalize or reject literal colons in path segments in front of the router. Suggestion: drop `:` from the allowed id charset, or keep the `/{id}/enable` sub-path style.

src/platform/Aevatar.GAgentService.Hosting/Endpoints/Schedules/ScheduledDispatchEndpoints.cs
```diff
@@ -109,7 +115,24 @@ internal static async Task<IResult> Disable(
...
+        }
+        catch (Exception ex) when (TryMapScheduleMutationError(ex, out var result))
+        {
+            return result;
+        }
+    }
+
+    internal static async Task<IResult> Delete(
+        string scheduleId,
+        [FromBody] ScheduledDispatchStateChangeHttpRequest? input,
```
eanzhao, line 129: **[P3] DELETE with a request body is routinely stripped by clients/proxies/Swagger tooling**, so the audit `reason` will often arrive empty without anyone noticing. Suggestion: accept `reason` as a query parameter on DELETE.

eanzhao (commented):
Reviewed `c7903f33f` against dev (merge-base `3cde812c9`). Direction is right — one scheduling spine on the actor-owned path, the legacy store/poller facade deleted, the `Deleted/DeletedAt` projection mirror is complete across Document/Projector/QueryPort, canon docs updated, and keeping the `scheduled-dispatches` index name stable through the rename is the correct call.

Two findings I'd treat as merge blockers, details inline:

1. **[P1] Raw NyxID bearer tokens persisted into durable schedule state** via the new caller-supplied `LlmControl` field.
2. **[P1] The unconditional `Deleted == false` list filter hides every pre-existing schedule document on Elasticsearch** (term query vs missing field).

Not inline-able, also worth addressing:

- **Lifecycle test coverage is thin.** Delete ships with 2 tests (lease-cancel + endpoint 404 mapping). Untested: delete -> LIST exclusion (exactly where the ES regression hides), create-after-delete, mutate-after-delete, and the late-fire-vs-delete race. The ~1.9k deleted test lines covered the removed facade (fine), but the replacement surface got ~180 lines.
- **Breaking rename for external callers**: `/api/scheduled-dispatches` -> `/api/schedules` plus the create-contract narrowing (Envelope/ServiceInvocation target shapes removed). No in-repo consumers are left behind (checked src/agents/tools/demos/test), but aevatar-cli / Studio / external automation need a migration note in the PR description.
- **Heads-up for the eventual dev -> feature/integrate merge**: integrate's copy of this actor already carries `[GAgent("gagent.service.scheduled-dispatch")]` (stable persisted kind, 952abf141), `HandleEnsureAsync` + `IScheduledDispatchApplicationService.EnsureAsync` (#1841), and a caller-credential projection flag; integrate's `WorkflowSelfRescheduleModule` also consumes the `WorkflowScheduleConfiguration` records this PR deletes with `WorkflowScheduleModels.cs`. A merge-tree dry run shows 5 content + 3 modify/delete conflicts. Pinning the same kind token in this PR (see inline comment on the class) is what makes the rename converge instead of conflict.

src/platform/Aevatar.GAgentService.Hosting/Endpoints/Schedules/ScheduledDispatchEndpoints.cs
louis4li: Fixed in 8462526b5. I removed `LlmControl` from the public scheduled workflow-chat HTTP target and stopped copying it into the persisted `ChatRequestEvent`; scheduled credentials now stay on the typed `Auth` credential-source path and are exchanged at fire time. Added endpoint coverage asserting the DTO no longer exposes `LlmControl` and the configured chat payload has no `LlmControl`.

src/platform/Aevatar.GAgentService.Projection/Queries/ScheduledDispatchQueryPort.cs
louis4li: Fixed in 8462526b5. I removed the store-side `Deleted == false` filter so Elasticsearch documents that predate the `deleted` field are not silently hidden. The query still post-filters returned tombstones with `!x.Deleted`, and the projection/query test now asserts no `Deleted` filter is sent to the reader while `Deleted=true` results are excluded.

src/platform/Aevatar.GAgentService.Application/Schedules/ScheduledDispatchApplicationService.cs
louis4li: Fixed in 8462526b5. The application service now preflights the read model before dispatch: create conflicts on existing/tombstoned ids, and update/enable/disable/run-now/delete return not found for missing or deleted schedules. That keeps `202` for commands that may actually progress instead of accepting statically doomed mutations.

src/platform/Aevatar.GAgentService.Core/Schedules/ScheduleGAgent.cs
louis4li: Handled in 8462526b5. We kept the irreversible tombstone semantics and made them observable before dispatch: create on an existing/tombstoned id now returns conflict instead of an accepted doomed command, while mutations on deleted schedules return not found. Added lifecycle coverage for tombstoned create/mutation behavior.

src/platform/Aevatar.GAgentService.Core/Schedules/ScheduleGAgent.cs
louis4li: Fixed in 8462526b5. Non-manual late fires on a deleted schedule now no-op before `EnsureConfiguredForWrite`, matching the stale/disabled callback behavior and avoiding actor retry noise. Manual run-now on a deleted schedule remains rejected. Added actor coverage for this race.

src/platform/Aevatar.GAgentService.Core/Schedules/ScheduleGAgent.cs
louis4li: Addressed in 8462526b5 by pinning `ScheduleGAgent` with `[GAgent("gagent.service.scheduled-dispatch")]`, matching the integrate token. I did not add a legacy CLR-name alias because I could not identify an exact stable legacy kind token in this branch; the new stable kind is now pinned for forward compatibility.

src/workflow/Aevatar.Workflow.Host.Api/Program.cs
louis4li: Addressed in 8462526b5. I removed the full `AddGAgentServiceCapabilityBundle()` from the standalone workflow host; `AddAevatarPlatform()` already registers the narrow scheduled-dispatch capability and maps `/api/schedules`, so the host no longer expands to the full GAgent service surface or demo bootstrap just for schedules. Route coverage still passes.

src/platform/Aevatar.GAgentService.Hosting/Endpoints/Schedules/ScheduledDispatchEndpoints.cs
louis4li: Fixed in 8462526b5. I removed `:` from the allowed schedule id character set in both the application service validation and `ScheduledDispatchActorId`, and added coverage for the rejection path.

src/platform/Aevatar.GAgentService.Hosting/Endpoints/Schedules/ScheduledDispatchEndpoints.cs
louis4li: Fixed in 8462526b5. DELETE now accepts `reason` as an optional query parameter while preserving body compatibility; query `reason` takes precedence when both are present. Added endpoint coverage for the query reason path.

eanzhao (commented):
Verified the fixes in `8462526b5` + `e2446f3eb` — all nine findings are addressed, each with a test pinning the behavior (`ScheduleGAgent_ShouldDeclareStableAgentKind` locking the kind token is a nice touch, and matching integrate's `gagent.service.scheduled-dispatch` collapses the future dev -> integrate merge conflict on this file). CI is green.

Two small follow-ups below — neither blocking.

src/platform/Aevatar.GAgentService.Application/Schedules/ScheduledDispatchApplicationService.cs
```diff
@@ -157,6 +175,20 @@ private static ScheduledDispatchMutationReceipt CreateMutationReceipt(
...
+    {
+        var existing = await _queryPort.GetAsync(scheduleId, ct);
+        if (existing != null)
+            throw new ScheduledDispatchConflictException(scheduleId, $"Scheduled dispatch '{scheduleId}' already exists.");
+    }
+
+    private async Task EnsureMutableAsync(string scheduleId, CancellationToken ct)
+    {
+        var existing = await _queryPort.GetAsync(scheduleId, ct);
+        if (existing?.Schedule.Deleted != false)
```
eanzhao: **[P3] This treats read-model lag the same as a tombstone, which regresses create-then-mutate.**

`existing?.Schedule.Deleted != false` returns 404 both when the schedule is deleted AND when the document simply hasn't materialized yet. On ES the projection lags by up to seconds, so `POST /schedules` followed immediately by `PUT`/`:enable` now 404s inside that window — before this preflight, the same call succeeded because `ResolveScheduleActorAsync` checks the actor directory, which is synchronously consistent with create.

Suggestion: only short-circuit on a positive tombstone (`existing?.Schedule.Deleted == true`) and let the null case fall through to `ResolveScheduleActorAsync` on the next line — the actor directory stays authoritative for existence, the read model for deletion. The only case that then slips through is mutate-racing-delete inside the projection window, where the actor's own guard still wins (202 + actor-side rejection), which is the pre-existing best-effort semantics.

`Mutations_ShouldPreflightReadModelLifecycle` covers deleted/missing but not the lagging-document case, which is exactly the one InMemory projection can't reproduce.

src/platform/Aevatar.GAgentService.Application/Schedules/ScheduledDispatchTargetPreparationService.cs
```diff
@@ -103,6 +109,19 @@ private static EventEnvelope CreateAdapterEnvelope(
             },
         };
 
+    private static Any StripCredentialBearingLlmControl(Any payload)
```
eanzhao, line 118: **[P3] The strip only runs for ServiceInvocation targets — Envelope targets still persist `LlmControl` verbatim.**

The commit message says non-HTTP callers can no longer persist credential-bearing chat control payloads, but `PrepareEnvelopeTarget` (line 30) clones `target.Envelope` without passing its payload through this method. A programmatic caller creating an `Envelope`-kind schedule whose payload is a `ChatRequestEvent` (or any message embedding `LLMControlContextPayload`) still gets raw tokens into durable state + committed events.

The HTTP surface can't reach this branch anymore, so it's internal-callers-only — but since the point of `e2446f3eb` was exactly defense-in-depth for non-HTTP callers, suggest applying the same strip to the envelope payload when `payload.Is(ChatRequestEvent.Descriptor)`, or narrowing the commit-message claim.

src/platform/Aevatar.GAgentService.Application/Schedules/ScheduledDispatchApplicationService.cs
louis4li: Fixed in `2c6d1b535`. `EnsureMutableAsync` now only short-circuits on a positive tombstone (`Deleted == true`); a missing/lagging read model falls through to `ResolveScheduleActorAsync`, so create-then-mutate can still dispatch when the actor exists. I also broadened the test to cover update/enable/disable/delete/run-now under read-model lag.

Validation: `dotnet test test/Aevatar.GAgentService.Tests/Aevatar.GAgentService.Tests.csproj --nologo --filter "FullyQualifiedName~ScheduledDispatchApplicationServiceTests|FullyQualifiedName~ScheduledDispatchServiceInvocationTests"` — 48 passed.

⟦AI:AUTO-LOOP⟧

src/platform/Aevatar.GAgentService.Application/Schedules/ScheduledDispatchTargetPreparationService.cs
louis4li: Fixed in `2c6d1b535`. `PrepareEnvelopeTarget` now runs the envelope payload through the same `ChatRequestEvent` LLM-control stripping path before building the durable prepared trigger envelope, and the prepared target descriptor is written back with the sanitized envelope clone. The original caller configuration remains unmodified.

Added `PrepareAsync_ShouldStripLlmControlBeforePersistingEnvelopeTarget` covering both the trigger envelope and descriptor payload.

Validation: `dotnet test test/Aevatar.GAgentService.Tests/Aevatar.GAgentService.Tests.csproj --nologo --filter "FullyQualifiedName~ScheduledDispatchApplicationServiceTests|FullyQualifiedName~ScheduledDispatchServiceInvocationTests"` — 48 passed.

⟦AI:AUTO-LOOP⟧

PR #2019, opened 2026-06-12
Preserve team member workflow draft hints

AbigailDeng:
## 问题与方案

This PR brings the paste YAML fixes from `fix/2026-06-11_paste-yaml` into `feat/2026-06-08_workflow-studio`.

It preserves team member workflow draft hints, adds the paste/import YAML studio panels, strengthens navigation into team member workflow studio, and adds backend repair/materialization support for workflow draft member data.

## 影响路径

- `apps/aevatar-console-web/src/pages/team-member-workflow-studio/*`
- `apps/aevatar-console-web/src/shared/navigation/*`
- `apps/aevatar-console-web/src/pages/teams/detail.*`
- `src/Aevatar.Studio.Application/*`
- `src/Aevatar.Studio.Hosting/*`
- `src/Aevatar.Studio.Projection/*`
- `test/Aevatar.Studio.Tests/*`

## 验证命令与结果

- Not run in this PR creation step.

## 相关文档更新

- No documentation updates included.


potter-sun (approved):

apps/aevatar-console-web/src/pages/team-member-workflow-studio/hooks/useTeamMemberWorkflowStudio.ts
```diff
@@ -1304,21 +1365,6 @@ export function useTeamMemberWorkflowStudio(): TeamMemberWorkflowStudioState {
...
-      await studioApi.bindMemberWorkflow({
-        scopeId: route.scopeId,
-        memberId: createdMemberId,
-        displayName: savedDraft.title,
-        workflowId: savedWorkflowId,
-        workflowYamls: [serialized.yaml],
-      });
-
       return {
         memberId: createdMemberId,
```
potter-sun, line 1029: [P1] This path now saves the workflow draft and creates the member, but it no longer writes any authoritative member -> workflow link. The only connection retained is the `workflowId` query/session hint from the redirect. `createMember` does not accept a workflow id, and scoped draft creation is explicitly draft-only (`CreateDraftAsync_ShouldPersistScopedDraftWithoutCreatingMemberAuthority`). After a refresh/new tab/shared URL without the query, the member read model still has no `implementationRef`, so the editor falls back to "No workflow draft is linked". We need to persist this through a real member contract/read model path, e.g. the existing implementationRef patch or binding flow, rather than browser state.

apps/aevatar-console-web/src/pages/team-member-workflow-studio/hooks/useTeamMemberWorkflowStudio.ts
```diff
@@ -998,6 +1013,15 @@ export function useTeamMemberWorkflowStudio(): TeamMemberWorkflowStudioState {
     publishedServiceId: memberPublishedServiceId,
     workflowId: route.workflowId,
   });
+  const routeDraftWorkflowId =
+    route.mode === "existing" &&
```
potter-sun, line 772: [P1] A same-member `workflowId` hint from the route/session now wins over the authoritative binding facts. Because `shouldResolveWorkflowFromRevision` is disabled whenever `route.workflowId` is any reloadable-looking draft id, we skip recovery from `lastBoundRevisionId` / published service even for already-bound members. Team detail can rehydrate this value from `sessionStorage` and put it back into the Studio URL, so an old hint can make a bound member open and publish the wrong draft. Route/session hints should be treated as candidates only after the backend member binding facts fail or explicitly match, not as higher-priority truth.

apps/aevatar-console-web/src/pages/team-member-workflow-studio/hooks/useTeamMemberWorkflowStudio.ts
```diff
@@ -1610,6 +1734,19 @@ export function useTeamMemberWorkflowStudio(): TeamMemberWorkflowStudioState {
       (workflowDraftReloadIds.length > 0 &&
         (workflowQuery.isLoading || parseQuery.isLoading)));
   const workflowHasSteps = Boolean(editableDocument?.steps?.length);
+  const canCreateUnlinkedMemberDraft = Boolean(
+    route.mode === "existing" &&
```
potter-sun, line 1406: [P2] A stale `workflowId` hint leaves the editor unable to recover. When the hinted draft 404s, `linkedWorkflowMissing` becomes true, but `canCreateUnlinkedMemberDraft` also requires `!routeDraftWorkflowId`, so even after the user pastes/adds steps Save remains disabled and the reason says "Load the workflow draft before saving." Since these hints can come from route/session state, the 404 path needs to clear/ignore the invalid hint or allow creating a replacement reusable draft.

potter-sun (commented):
Codex review: focused on member/workflow identity and draft recovery semantics.

PR #2040, opened 2026-06-12
Stream workflow draft run logs

AbigailDeng:
## Problem
- Workflow Studio draft run logs looked like they arrived all at once because the Run log tab only created node cards after a step completed.
- Blank draft-run input was still treated like an implicit default prompt in the UI flow.

## Solution
- Render step run cards as soon as step request SSE frames arrive, then update the same card when completion/failure frames arrive.
- Keep blank draft-run input blank instead of synthesizing a default prompt, and update the panel/localized copy accordingly.
- Keep workflow draft title semantics separate from member display names.
- Add a controlled SSE test that emits frames one by one and asserts intermediate UI states.

## Impacted Paths
- apps/aevatar-console-web/src/pages/team-member-workflow-studio/components/WorkflowStudioExecutionPanel.tsx
- apps/aevatar-console-web/src/pages/team-member-workflow-studio/hooks/useTeamMemberWorkflowStudio.ts
- apps/aevatar-console-web/src/pages/team-member-workflow-studio/index.test.tsx
- apps/aevatar-console-web/src/pages/team-member-workflow-studio/components/WorkflowStudioDraftRunPanel.tsx
- apps/aevatar-console-web/src/shared/studio/api.ts
- apps/aevatar-console-web/src/locales/en-US.ts
- apps/aevatar-console-web/src/locales/zh-CN.ts

## Verification
- pnpm --dir apps/aevatar-console-web jest src/pages/team-member-workflow-studio/index.test.tsx --runInBand
- pnpm --dir apps/aevatar-console-web tsc
- git diff --check


apps/aevatar-console-web/src/pages/team-member-workflow-studio/hooks/useTeamMemberWorkflowStudio.ts
```diff
@@ -972,6 +1081,15 @@ export function useTeamMemberWorkflowStudio(): TeamMemberWorkflowStudioState {
         throw new Error("Resolve a Team workspace before creating a workflow member.");
       }
 
+      if (pendingCreatedWorkflowMemberLink) {
```
potter-sun, line 1084: This retry branch links the previously saved draft without first saving the current editor state. If the first create flow gets past `saveWorkflow`/`createMember` but fails while waiting for member materialization or updating the implementation ref, `pendingCreatedWorkflowMemberLink` remains set. On the next Save, any edits the user made after the failure are skipped because we return the old `savedDraft` here; `onSuccess` then applies that stale draft and navigates away. The new test even asserts `saveWorkflow` is still called only once on retry, which locks in the data-loss path. Please either save the current document/title before retrying the link, or make the retry explicit in a way that does not discard unsaved editor changes.

apps/aevatar-console-web/src/pages/team-member-workflow-studio/components/WorkflowStudioExecutionPanel.tsx
```diff
@@ -66,6 +80,73 @@ function trimConsoleText(value: string, maxLength: number): string {
...
+        outputText: "",
+        stepId: log.stepId,
+      });
+      activeCardIndexByStepId.set(log.stepId, cards.length - 1);
+      return;
+    }
+
+    if (log.tone === "pending" || log.tone === "run") {
+      if (activeCard && !isTerminalStepLog(activeCard.statusLog)) {
+        activeCard.statusLog = log;
```
potter-sun, line 123: This collapses pending/run step logs into the node card status but drops the pending log text (`log.clipboardText`) and interaction data. For a workflow that pauses on `aevatar.human_input.request` / approval, the old Logs view rendered the prompt (`approve waiting for approval`, `Need approval`); after this change the card only shows the original input/output blocks, or blank placeholders when there was no started card. That hides the exact prompt the user needs to respond to. Please preserve/render the pending prompt separately, or keep pending human-input/approval logs visible in the default log view.

apps/aevatar-console-web/src/pages/team-member-workflow-studio/hooks/useTeamMemberWorkflowStudio.ts
AbigailDeng: Fixed in d97871dcc.

The pending-created-member retry now saves the current editor document/title back into the existing saved draft workflow before retrying the member implementation link. The retry still reuses the previously created member, but it no longer returns/applies the stale saved draft. I also updated the regression test so the retry edits the title and asserts a second save to the same workflowId before linking.

Verified with:
- pnpm --dir apps/aevatar-console-web tsc
- pnpm --dir apps/aevatar-console-web test --runInBand src/pages/team-member-workflow-studio/index.test.tsx

apps/aevatar-console-web/src/pages/team-member-workflow-studio/components/WorkflowStudioExecutionPanel.tsx
AbigailDeng: Fixed in d97871dcc.

Node run cards now retain pending prompt text separately from status, and render it as a Prompt block in the default Logs view. Interaction/run text is also kept separately so a later run event does not wipe out the pending prompt. The draft-run test now includes a human approval request and asserts the approval prompt is visible in the default log card.

Verified with:
- pnpm --dir apps/aevatar-console-web tsc
- pnpm --dir apps/aevatar-console-web test --runInBand src/pages/team-member-workflow-studio/index.test.tsx

PR #2078, opened 2026-06-15
Fix member binding completion semantics

AbigailDeng:
## Summary
- Treat `publishedServiceId` as member-owned service identity, not a completed binding fact.
- Gate Workflow Studio published state, Team member invoke, Team entry resolution, and member contract/revision reads on current completed binding facts (`bind_ready` plus revision fact).
- Add regressions for identity-only members, stale previous bindings, invoke blocking, and backend contract admission.

## Verification
- `pnpm --dir apps/aevatar-console-web test --runInBand src/pages/team-member-invoke/index.test.tsx src/pages/team-member-workflow-studio/index.test.tsx`
- `bash tools/ci/workflow_binding_boundary_guard.sh`
- `git diff --check`

## Not run
- `dotnet test ...` because `dotnet` is not available in the local PATH (`zsh:1: command not found: dotnet`).

### DIFF
--- a/apps/aevatar-console-web/src/pages/team-member-workflow-studio/hooks/useTeamMemberWorkflowStudio.ts
+++ b/apps/aevatar-console-web/src/pages/team-member-workflow-studio/hooks/useTeamMemberWorkflowStudio.ts
@@ -35,6 +35,7 @@ import {
   STUDIO_GRAPH_CATEGORIES,
 } from "@/shared/studio/graph";
 import { isStudioApiStatus, studioApi } from "@/shared/studio/api";
+import { normalizeStudioMemberLifecycleStage } from "@/shared/studio/models";
 import type {
   StudioExecutionDetail,
   StudioExecutionFrame,
@@ -211,6 +212,41 @@ const CREATED_MEMBER_MATERIALIZATION_DELAY_MS = 450;
 
 function trimOptional(value: string | null | undefined): string {
   return value?.trim() ?? "";
+}
+
+function hasCurrentCompletedMemberBinding(
+  detail: StudioMemberDetail | null | undefined,
+): boolean {
+  if (
+    normalizeStudioMemberLifecycleStage(detail?.summary.lifecycleStage) !==
+    "bind_ready"
+  ) {
+    return false;
+  }
+
+  return Boolean(
+    trimOptional(detail?.summary.lastBoundRevisionId) ||
+      trimOptional(detail?.lastBinding?.revisionId),
+  );
+}
+
+function hasActiveMemberBindingRun(
+  detail: StudioMemberDetail | null | undefined,
+): boolean {
+  return Boolean(
+    detail?.currentBindingRun &&
+      !isTerminalBindingRun(detail.currentBindingRun),
+  );
+}
+
+function readActiveMemberBindingRunId(
+  detail: StudioMemberDetail | null | undefined,
+): string {
+  if (!hasActiveMemberBindingRun(detail)) {
+    return "";
+  }
+
+  return trimOptional(detail?.currentBindingRun?.bindingRunId);
 }
 
 function readPathSegments(): {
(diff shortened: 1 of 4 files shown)


apps/aevatar-console-web/src/pages/team-member-workflow-studio/hooks/useTeamMemberWorkflowStudio.ts
```diff
@@ -1673,22 +1737,33 @@ export function useTeamMemberWorkflowStudio(): TeamMemberWorkflowStudioState {
...
+    authoritativeBindingRunInProgress ||
+      (publishBindingRun &&
+        publishBindingRun.status !== "succeeded" &&
+        !publishBindingRunTerminal),
   );
   const publishHasDraftChanges = Boolean(dirty && workflowHasSteps);
   const publishPending = publishMutation.isPending;
   const memberPublished = memberIsPublished;
+  const publishVisibleError =
+    memberIsPublished || publishStatusStillInProgress ? "" : publishError;
```
potter-sun: [P1] This masks failed re-publishes for members that already have an older successful binding.

A dirty published member can still publish a new draft version because `publishDisabled` only blocks `memberIsPublished && !publishHasDraftChanges`. If that new binding run returns `failed` or `rejected`, `mutationFn` throws and `onError` stores `publishError`, but `memberIsPublished` is still true from the previous completed binding. This line then blanks `publishVisibleError`, so the header falls through to the success tone and `Published member workflow is serviceable.` The latest publish failure disappears from the persistent UI.

Please only suppress the specific stale-preflight/already-published case, or track the latest attempted binding run separately and surface terminal failure even when an older binding remains serviceable.

apps/aevatar-console-web/src/pages/team-member-workflow-studio/hooks/useTeamMemberWorkflowStudio.ts
```diff
@@ -1719,40 +1796,87 @@ export function useTeamMemberWorkflowStudio(): TeamMemberWorkflowStudioState {
...
+    const memberResult = await memberQuery.refetch();
     if (routeDraftWorkflowId) {
-      void workflowQuery.refetch();
+      await workflowQuery.refetch();
+    }
+
+    const refreshedMember = memberResult.data;
+    const activeRunId = readActiveMemberBindingRunId(refreshedMember);
+    if (activeRunId) {
+      const refreshedRun = await studioApi.getMemberBindingRun(
```
potter-sun: [P2] Refresh can fail silently while the binding-run read model is not materialized yet.

`memberQuery.refetch()` returns a hydrated member detail whose `currentBindingRun` may come from the member read model fallback when the run-owned read model is missing. The backend does that intentionally in `ResolveCurrentBindingRunAsync`: it returns `run ?? memberCurrentRun`. But this refresh path immediately calls `getMemberBindingRun(...)`, whose endpoint throws `STUDIO_MEMBER_BINDING_RUN_NOT_FOUND` when that run document is still absent. There is no catch here, so clicking Refresh status can reject the callback with no UI message and leave the user stuck on the stale Publishing state during normal projection lag.

Please handle 404 by keeping/using `refreshedMember.currentBindingRun` as the fallback status, matching the member detail API semantics, or show a real retryable message instead of letting the promise escape.

apps/aevatar-console-web/src/pages/team-member-workflow-studio/hooks/useTeamMemberWorkflowStudio.ts
AbigailDeng: Fixed in e4d1916ae. The publish error is no longer suppressed just because an older completed binding still makes the member serviceable. A terminal failed/rejected latest binding run now keeps the header in Error and surfaces the latest failure message. Added regression coverage with `surfaces failed republish even when an older member binding remains serviceable`.

Verified with:
- `pnpm --dir apps/aevatar-console-web test --runInBand --runTestsByPath src/pages/team-member-workflow-studio/index.test.tsx`

apps/aevatar-console-web/src/pages/team-member-workflow-studio/hooks/useTeamMemberWorkflowStudio.ts
AbigailDeng: Fixed in e4d1916ae. Refresh now catches the binding-run 404 case and falls back to the refreshed member detail currentBindingRun when it matches the active run id, preserving the member detail API semantics during projection lag instead of letting the callback reject silently. Added regression coverage with `falls back to member current binding run when refresh read model is not materialized`.

Verified with:
- `pnpm --dir apps/aevatar-console-web test --runInBand --runTestsByPath src/pages/team-member-workflow-studio/index.test.tsx`

PR #2089, opened 2026-06-15
Add member automation scheduling UI

potter-sun:
## Summary
- add the Team Automations tab for member-owned recurring schedule management
- wire the frontend to the schedules API for list, create, update, preview, run-now, pause/resume, and delete
- add Workflow Studio and Team member entry points into recurring work while preserving member/workflow/service identity boundaries
- add locale copy and tests for schedule API payloads and Team automation flows

## Related issues
- Refs #1963

## Related backend PR
- Depends on #2087

## Backend contract
Targets the schedule API introduced by backend PR #2087:
- `GET /api/schedules`
- `POST /api/schedules`
- `PUT /api/schedules/{scheduleId}`
- `POST /api/schedules/preview`
- `POST /api/schedules/{scheduleId}:run-now`
- `POST /api/schedules/{scheduleId}:enable`
- `POST /api/schedules/{scheduleId}:disable`
- `DELETE /api/schedules/{scheduleId}`

## Identity boundary
The UI treats automations as member-owned in the Team surface, but schedule payloads resolve the selected member to its published service identity. Tests assert the schedule payload does not send `memberId` or `workflowId` to `/api/schedules`.

## Validation
- `pnpm --dir apps/aevatar-console-web test --runInBand -- scheduledDispatchApi`
- `pnpm --dir apps/aevatar-console-web test --runInBand -- pages/teams/detail`
- `pnpm --dir apps/aevatar-console-web test --runInBand -- hardcodedCopyAudit catalog`


AbigailDeng (changes requested):
Thanks for the work here. I re-reviewed the current head (`3d9ed15303ddede04cdf12a8dace2a831ab1987e`) and I think this still needs changes before merge.

1. [P0] `apps/aevatar-console-web/src/shared/api/scheduledDispatchApi.ts:398`

`workflowChatTarget.prompt` is only trimmed at line 379 and then encoded directly into the scheduled `ChatRequestEvent` payload here, while the UI textarea has no `maxLength`. That makes this a persisted, model-visible recurring prompt with no hard bound, replayed into every scheduled chat run. Please define and enforce a bounded prompt contract before encoding/storing the schedule payload, and surface the same limit in the form. A UI counter alone would not be enough; oversized prompts should be rejected at the client/API contract boundary and covered by tests.

2. [P1] `apps/aevatar-console-web/src/shared/api/scheduledDispatchApi.ts:397`

The client posts `serviceInvocation.payloadTypeUrl` and `serviceInvocation.payloadBase64`, but the current `/api/schedules` app-server contract binds `serviceInvocation.payload` as a required protobuf `Any` (`ScheduledDispatchEndpoints.cs:275`). Unlike the normal service invocation endpoint, this schedule endpoint does not pack the flattened base64 shape into `Payload`, so create/update requests from this UI will fail binding/validation against the current server API. Please either send the accepted `payload` object shape or land an explicit backend adapter for `payloadTypeUrl`/`payloadBase64` before shipping this UI.

3. [P1] `apps/aevatar-console-web/src/pages/teams/tabs/TeamAutomationsTab.tsx:714`

The Team automations tab calls the broad schedules list and then keeps schedules by matching only bare `schedule.serviceId` against the selected team members. `serviceId` is only one component of the service identity; two scopes/apps/namespaces can reuse the same value, and the schedule summary already carries `serviceKey` for the full identity. This can show or mutate another schedule that happens to share the same service id. Please filter by the full scoped identity (`serviceKey` / tenant + app + namespace + serviceId) or add a scoped backend query.

4. [P1] `apps/aevatar-console-web/src/pages/teams/tabs/TeamAutomationsTab.tsx:625`

This PR is too large and non-mechanical to review safely as one unit: 17 files, `+5026/-3`, with a new 2,569-line interactive `TeamAutomationsTab.tsx` plus API contract/encoding, routing, workflow-studio entry points, locales, and large interaction tests. A reviewable split would be: schedule API/client + protobuf/error helpers first, route/member identity handoff second, Team tab shell third, then the interactive automation UI split into list/read-only, create/preview, and edit/pause/resume/delete/run-now behavior.

5. [P2] `apps/aevatar-console-web/src/shared/api/scheduledDispatchApi.ts:387`

Workflow automations are created through the generic scheduled-dispatch request with no way to set `ScheduleKind.Workflow`; the backend `ScheduledDispatchConfiguration` defaults to `Generic`, while workflow schedule consumers list/filter only workflow schedules. After the payload-shape issue is fixed, these records can still be invisible to workflow schedule APIs and any workflow-specific behavior keyed off `ScheduleKind.Workflow`. Please use/extend a workflow schedule contract that writes the workflow schedule kind.

6. [P2] `apps/aevatar-console-web/src/shared/navigation/teamRoutes.ts:247`

`buildTeamMemberAutomationsHref` still builds `/scopes/:scopeId/teams/:teamId?memberId=...&tab=automations`. The repo route contract requires member surfaces to preserve `scope -> team -> member` ownership in the path, and the new tests currently assert the query-string shape. Please make the member automation handoff use a canonical member path instead of carrying member identity in query state.

7. [P2] `apps/aevatar-console-web/src/shared/runs/protobufPayload.ts:79`

`encodeChatRequestEventBase64` has no independent golden test for the actual protobuf bytes/field numbers. The new `scheduledDispatchApi` tests build the expected value by calling the same helper, so a regression such as encoding `scopeId` on the wrong field, omitting `sessionId`, or changing the type URL could still pass. Please add a direct `protobufPayload.test.ts` golden case with fixed expected base64 for prompt/session/scope.

8. [P2] `apps/aevatar-console-web/src/pages/teams/tabs/TeamAutomationsTab.tsx:2525`

The “Preview next runs” modal action is introduced, but the page tests only assert static preview copy and never click this button or verify rendered `nextFireTimes`. A broken API call, wrong cron/timezone/count payload, or missing render of returned fire times would stay green. Please add a page-level test that opens the automation modal, clicks preview, asserts the API input, and verifies the returned next run appears.

9. [P3] `apps/aevatar-console-web/src/shared/api/scheduledDispatchApi.ts:521`

The schedule delete helper is not covered at the API-client layer. Page tests mock `scheduledDispatchApi.delete`, while `scheduledDispatchApi.test.ts` covers run-now/enable/disable route shapes but not `DELETE /api/schedules/{id}`. Please add an API unit test for the delete URL, reason handling, and receipt decode so a route/body regression is caught below the UI layer.

PR #2090, opened 2026-06-15
chore: prune low-value frontend tests

AbigailDeng:
## 问题

前端测试套件里有一批低价值、重复或过度绑定实现细节的测试，会增加维护成本，但对真实产品行为保护有限。

## 方案

- 删除 7 个低价值前端测试文件，主要是静态配置、纯 mock 调用、重复覆盖或过浅的结构性断言。
- 精简 `StudioShell` 和 `primitives` 相关测试里的布局/CSS 细节断言，保留用户可感知行为断言。
- 更新 Jest node project 显式测试文件列表，移除已删除的 `aevatarWorkbench` 测试入口。

## 影响路径

- `apps/aevatar-console-web/jest.config.ts`
- `apps/aevatar-console-web/src/pages/**`
- `apps/aevatar-console-web/src/shared/**`

## 验证

- `git diff --check -- apps/aevatar-console-web`
- `pnpm --dir apps/aevatar-console-web test --runInBand`

完整 Jest 结果：112 suites passed, 897 tests passed。

## 备注

完整 Jest 仍会输出既有 console 噪音，包括 `runs/index.test.tsx` 的 `NaN height` / React Query undefined data、Ant Design `autoInsertSpaceInButton` deprecation、以及 jsdom 解析 Ant Design CSS 的警告；本 PR 没有引入新的测试失败。


apps/aevatar-console-web/src/pages/studio/components/StudioBootstrapGate.test.tsx
```diff
@@ -1,28 +0,0 @@
-import { render, screen } from '@testing-library/react';
-import React from 'react';
-import StudioBootstrapGate from './StudioBootstrapGate';
-
-describe('StudioBootstrapGate', () => {
-  it('renders a single bootstrap summary banner and keeps children mounted', () => {
```
potter-sun, line 6: [P2] This removes the only test that exercises `StudioBootstrapGate` itself. The page-level Studio tests mock this component out (`studio/index.test.tsx` replaces it with `children`), while production still wraps the whole Studio shell in it. That leaves the degraded bootstrap behavior unguarded: app/workspace/auth errors should produce one warning/error banner while keeping the workbench mounted. Please keep a small behavior test here, or move the same assertions into a non-mocked integration test, so a partial bootstrap failure cannot regress silently.

apps/aevatar-console-web/src/pages/runs/components/RunsMessagesView.test.tsx
```diff
@@ -1,53 +0,0 @@
...
-import React from "react";
-import { setLocale } from "@umijs/max";
-import RunsMessagesView from "./RunsMessagesView";
-
-describe("RunsMessagesView", () => {
-  beforeEach(() => {
-    setLocale("en-US");
-  });
-
-  it("renders message cards with role, status, and content", () => {
```
potter-sun, line 11: [P2] This deletes the only UI-level assertion that run messages render with role, streaming/completed status, id, and content. `runs/index.test.tsx` resets `mockSession.messages = []` in `beforeEach` and only checks that the Conversation surface exists, while `runEventPresentation.test.ts` covers event-row shaping rather than the rendered transcript. Please keep a focused `RunsMessagesView` test or add a page-level test with non-empty session messages, otherwise a regression that drops streamed assistant/user content from the Run Console can pass the suite.

PR #2094, opened 2026-06-15
Fix studio member patch receipt decoding

AbigailDeng:
## Problem

Saving a new Team member workflow could show `StudioMemberDetail.summary must be an object.` The frontend decoded `PATCH /api/scopes/{scopeId}/members/{memberId}` as `StudioMemberDetail`, but the backend contract on `feature/integrate` returns `202 Accepted` with `StudioMemberCommandResponse`.

## Solution

- Add typed `StudioMemberCommandResponse` / status models.
- Decode member PATCH responses as command receipts for implementation-ref and team-assignment updates.
- Update API tests to use the `202` receipt shape from the backend contract.

## Verification

- `pnpm --dir apps/aevatar-console-web test --runInBand src/shared/studio/api.test.ts src/pages/team-member-workflow-studio/index.test.tsx`

## Manual Check

Started the fixed frontend on `http://127.0.0.1:5174` with both API targets set to `https://aevatar-console-backend-api.aevatar.ai`, and verified `/api/studio/context` proxies to the hosted backend.

### DIFF
--- a/apps/aevatar-console-web/src/shared/studio/api.ts
+++ b/apps/aevatar-console-web/src/shared/studio/api.ts
@@ -19,6 +19,8 @@ import type {
   StudioMemberBindingRunStatus,
   StudioMemberBindingRunStatusResponse,
   StudioMemberBindingViewResponse,
+  StudioMemberCommandResponse,
+  StudioMemberCommandStatus,
   StudioMemberDetail,
   StudioMemberImplementationKind,
   StudioMemberImplementationRef,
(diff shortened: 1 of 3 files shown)


apps/aevatar-console-web/src/shared/studio/api.ts
```diff
@@ -1659,10 +1711,10 @@ export const studioApi = {
     scopeId: string;
     memberId: string;
     teamId: string | null;
-  }): Promise<StudioMemberDetail> {
+  }): Promise<StudioMemberCommandResponse> {
     return requestDecodedJson(
       `/api/scopes/${encodeURIComponent(input.scopeId.trim())}/members/${encodeURIComponent(input.memberId.trim())}`,
-      decodeStudioMemberDetail,
+      decodeStudioMemberCommandResponse,
```
potter-sun, line 1717: `updateMemberTeamAssignment`/`updateMemberImplementationRef` now require `PATCH /api/scopes/{scopeId}/members/{memberId}` to return a command receipt, but this PR does not change the backend contract on the target branch. In `origin/pr/2094`, `StudioMemberEndpoints.HandlePatchAsync` still calls `memberService.UpdateAsync(...)` and returns `Results.Ok(detail)`, and `StudioMemberEndpointsTests.HandlePatchAsync_ShouldReturnOk_OnImplementationRefPatch` still asserts `Ok<StudioMemberDetailResponse>`. So merged as-is, the frontend will still fail against the repo backend, just with a new decode error (`StudioMemberCommandResponse.status/scopeId/memberId` missing instead of `StudioMemberDetail.summary` missing). Please either update the PATCH endpoint/tests in this PR to return the 202 receipt shape, or keep this decoder compatible with both the existing detail response and the new receipt until the backend contract lands on this branch.

AbigailDeng: Thanks for checking this. This PR is intentionally aligned with the `feature/integrate` backend contract, where `PATCH /api/scopes/{scopeId}/members/{memberId}` returns `202 Accepted` with `StudioMemberCommandResponse`.

Given the backend contract is expected to land on the workflow-studio branch before this frontend PR is merged, I am going to keep the decoder receipt-only instead of adding a temporary dual-shape compatibility path. The merge order should be: backend contract lands/rebases into the target branch first, then this frontend PR.

PR #2219, opened 2026-06-17
Handle scoped workflow draft accepted polling

AbigailDeng:
## Summary
- Decode scoped workflow draft create `202 Accepted` responses as accepted/readiness receipts.
- Poll the single draft detail endpoint by returned `workflowId`, treating interim `404` as projection pending.
- Apply saved workflow selection and Team member workflow linking only after the draft detail read model returns a real `WorkflowDraftResponse`.

## Validation
- `pnpm --dir apps/aevatar-console-web tsc --pretty false`
- `git diff --check`

## Notes
- Frontend-only change.
- Unit tests intentionally deferred until manual validation passes.

### DIFF
--- a/apps/aevatar-console-web/src/pages/team-member-workflow-studio/hooks/useTeamMemberWorkflowStudio.ts
+++ b/apps/aevatar-console-web/src/pages/team-member-workflow-studio/hooks/useTeamMemberWorkflowStudio.ts
@@ -1243,6 +1243,22 @@ export function useTeamMemberWorkflowStudio(): TeamMemberWorkflowStudioState {
     },
     [queryClient, route.scopeId],
   );
+  const invalidateSavedWorkflowDraft = React.useCallback(
+    (saved: SavedWorkflowDraft) => {
+      const savedWorkflowId = trimOptional(saved.workflow.workflowId);
+      if (!route.scopeId || !savedWorkflowId) {
+        return;
+      }
+
+      void queryClient.invalidateQueries({
+        queryKey: getTeamMemberWorkflowStudioWorkflowQueryKey(
+          route.scopeId,
+          savedWorkflowId,
+        ),
+      });
+    },
+    [queryClient, route.scopeId],
+  );
   const markSavedDraft = React.useCallback(
     (saved: SavedWorkflowDraft) => {
       cacheSavedWorkflowDraft(saved);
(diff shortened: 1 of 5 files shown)


apps/aevatar-console-web/src/pages/team-member-workflow-studio/hooks/useTeamMemberWorkflowStudio.ts
```diff
@@ -1330,6 +1431,7 @@ export function useTeamMemberWorkflowStudio(): TeamMemberWorkflowStudioState {
     },
     onSuccess: ({ memberId, savedDraft, workflowId }) => {
       setPendingCreatedWorkflowMemberLink(null);
+      cacheSavedWorkflowDraft(savedDraft);
```
potter-sun, line 1450: This now seeds the exact workflow query that the route enables after `history.replace`. Because the query above also has a 30s `staleTime`, React Query treats this optimistic saved draft as fresh and never calls `studioApi.getWorkflow` after entering the new member route. The existing `reloads route state after creating a workflow member` test now fails with `Expected: "wf-untitled-member", "scope-1" / Number of calls: 0` when run as `pnpm --dir apps/aevatar-console-web jest src/pages/team-member-workflow-studio/index.test.tsx --runInBand --testNamePattern="reloads route state after creating a workflow member"`. Please either invalidate/refetch the workflow query after the route swap, or avoid marking this route query fresh, so the editor still reconciles with the route/read-model state after member creation.

AbigailDeng: Fixed in dc8ef0a92 by invalidating the freshly seeded workflow draft query after member creation. The editor still applies the saved draft immediately, but the route query is marked stale so the new member route refetches/reconciles with the read-model state instead of treating the optimistic cache as fresh for the 30s stale window.

Verified locally:
- `pnpm --dir apps/aevatar-console-web jest src/pages/team-member-workflow-studio/index.test.tsx --runInBand --testNamePattern="reloads route state after creating a workflow member"`
- `pnpm --dir apps/aevatar-console-web jest src/pages/team-member-workflow-studio/index.test.tsx --runInBand`
- `pnpm --dir apps/aevatar-console-web test --runInBand`
- `pnpm --dir apps/aevatar-console-web build`

PR #2236, opened 2026-06-17
Allow draft workflow runs before save

AbigailDeng:
## Problem

Workflow Studio draft run was gated on an existing Team member identity. That meant a user could not run the currently edited workflow until saving/creating the member first, even though the draft-run endpoint accepts the current workflow YAML directly.

## Solution

- Treat the run action as a current draft workflow run instead of an active member run.
- Allow the draft-run panel to open for any loaded current workflow document in a scoped workspace.
- Serialize the current editable workflow document and call the scope workflow draft-run endpoint directly.
- Keep member identity only as optional local execution context for display/id generation.
- Add a regression test that runs an unsaved new workflow draft without calling save/member creation/binding APIs.

## Impact paths

- `apps/aevatar-console-web/src/pages/team-member-workflow-studio/hooks/useTeamMemberWorkflowStudio.ts`
- `apps/aevatar-console-web/src/pages/team-member-workflow-studio/index.tsx`
- `apps/aevatar-console-web/src/pages/team-member-workflow-studio/components/WorkflowStudioHeader.tsx`
- `apps/aevatar-console-web/src/pages/team-member-workflow-studio/index.test.tsx`

## Verification

- `pnpm --dir apps/aevatar-console-web install --frozen-lockfile`
- `pnpm --dir apps/aevatar-console-web test src/pages/team-member-workflow-studio/index.test.tsx --runInBand`
- `git diff --check`

## Notes

This PR is intentionally based directly on `dev` and contains only the draft-run fix, avoiding the broader `feat/2026-06-08_workflow-studio` branch delta.


apps/aevatar-console-web/src/pages/team-member-workflow-studio/hooks/useTeamMemberWorkflowStudio.ts
```diff
@@ -2157,37 +2156,31 @@ export function useTeamMemberWorkflowStudio(): TeamMemberWorkflowStudioState {
...
-  const canRunActiveMember = Boolean(
-    route.mode === "existing" &&
-      route.scopeId &&
-      route.memberId &&
+  const canRunCurrentDraft = Boolean(
+    route.scopeId &&
       editableDocument &&
       workflowHasSteps &&
-      !activeMemberRunMutation.isPending,
+      !currentDraftRunMutation.isPending,
```
potter-sun, line 2164: [P1] `canRunCurrentDraft` now enables unsaved new drafts without considering `selectedStepConfigurationError`. On `/members/new/workflow`, a user can add a node, type invalid or unapplied inspector JSON, then click Run; `openDraftRunPanel` clears the error and hides the inspector, and `runCurrentDraft` serializes the last valid `editableDocument`. That starts a draft run with stale YAML while Save/Publish are blocked for the same edit. Please gate `canRunCurrentDraft`/`canOpenDraftRunPanel` and the `runCurrentDraft` guard on `!selectedStepConfigurationError`, and surface that same disabled reason instead of clearing it when opening the panel.

apps/aevatar-console-web/src/pages/team-member-workflow-studio/index.test.tsx
```diff
@@ -579,6 +579,73 @@ describe("TeamMemberWorkflowStudioPage", () => {
...
+      within(draftRunPanel).getByRole("button", { name: "Start draft run" }),
+    );
+
+    await waitFor(() => {
+      expect(runtimeRunsApi.streamDraftRun).toHaveBeenCalledWith(
+        "scope-1",
+        expect.objectContaining({
+          prompt: "Run the unsaved workflow",
+          workflowYamls: [
+            expect.stringContaining("name: Untitled member"),
```
potter-sun: [P2] This regression test does not assert that the unsaved edit is actually included in the draft-run payload. The test inserts an LLM node above, but the endpoint assertion only checks `name: Untitled member`; a broken implementation that serializes a blank document with the same title would still pass. Please also assert the emitted YAML contains the inserted step identity/type, for example `id: ...` and `type: llm_call`, so the test proves the current editable draft is what reaches `streamDraftRun`.

AbigailDeng: [from Codex]: Addressed in 2f60d0475. The unsaved draft-run regression now asserts the emitted YAML contains the inserted step id (`llm_step`) and type (`llm_call`), proving the editable draft reaches `streamDraftRun`.

apps/aevatar-console-web/src/pages/team-member-workflow-studio/hooks/useTeamMemberWorkflowStudio.ts
AbigailDeng: [from Codex]: Addressed in 2f60d0475. Draft runs are now blocked while the selected node configuration is invalid, and the same configuration error is surfaced on Run instead of opening the panel and clearing it.

PR #2610, opened 2026-07-06
Integrate audit trail, security hardening, and admin visibility

eanz17:
## Summary
- Integrates the platform audit trail stack across abstractions, core materialization, hosting endpoints, endpoint capture middleware, and document storage.
- Wires audit translation coverage through channel registration/callbacks, Studio lifecycle, GAgent service flows, CQRS observatory endpoints, workflow artifacts, and tool execution.
- Adds security hardening around JWT/DPoP validation, issuer key resolution, replay protection, secret redaction, OAuth client ES ACL probing, tool approval defaults, and mainnet scripting opt-in composition.
- Adds the Admin audit trail viewer with cross-scope filtering and updates workflow board total-step derivation from definition steps.
- Updates architecture docs, audit trail canon/ADR, CI guards, and broad unit/integration coverage for the integrated paths.

## Impact Paths
- `src/Aevatar.Audit.*`, `src/Aevatar.Bootstrap`, `src/Aevatar.Mainnet.Host.Api`
- `src/Aevatar.AI.*`, `src/Aevatar.Studio.*`, `src/platform/Aevatar.GAgentService.*`
- `src/workflow/Aevatar.Workflow.*`, `agents/channels/*`, `agents/Aevatar.GAgents.Channel.Identity`
- `test/*` coverage for audit trail, security/auth, endpoint capture, workflow artifacts, service translators, and channel flows
- `docs/adr/0039-platform-audit-trail.md`, `docs/canon/audit-trail.md`, and `tools/ci/*` guards

## Validation
- Not rerun locally while opening this integration PR.
- Branch includes focused test and guard coverage for the integrated audit/security/workflow changes; CI should run the full validation against `dev`.


src/Aevatar.Authentication.Hosting/AevatarAuthenticationHostExtensions.cs
```diff
@@ -99,6 +113,19 @@ public static WebApplicationBuilder AddAevatarAuthentication(this WebApplication
                     ConfigureScopeServiceTokenValidation(jwt, options, keyProvider));
         }
 
+        // DPoP (RFC 9449) sender-constraint validation. The validator and a no-op replay
+        // guard are always registered so hosts can enable the check via configuration alone;
+        // the OnTokenValidated hook is a no-op unless Aevatar:Authentication:DPoP:Enabled=true,
+        // so default bearer behavior is unchanged.
+        builder.Services.TryAddSingleton<IDPoPReplayGuard, NoOpDPoPReplayGuard>();
```
eanz17, line 139: [P1] Enabling DPoP currently provides no replay defense. This is the only production `IDPoPReplayGuard` registration in the repository, and `NoOpDPoPReplayGuard.TryRegisterAsync` always returns true, so the same signed proof and `jti` can be replayed throughout the accepted freshness window. Either wire a shared TTL-backed guard before allowing `DPoP:Enabled=true`, or fail startup when DPoP is enabled with the no-op implementation.

src/Aevatar.Authentication.Hosting/PerIssuerSigningKeyResolver.cs
```diff
@@ -0,0 +1,98 @@
...
+
+    private IEnumerable<SecurityKey> AllConfiguredKeys(TokenValidationParameters validationParameters)
+    {
+        IEnumerable<SecurityKey> configured = validationParameters.IssuerSigningKeys ?? [];
+        if (validationParameters.IssuerSigningKey is not null)
+            configured = configured.Append(validationParameters.IssuerSigningKey);
+
+        // Include this resolver's own scope keys too, so the fallback is a genuine superset
+        // even if the parameters were rebuilt without them.
+        return configured.Concat(_scopeKeys).Distinct();
```
eanz17: [P1] Keep scope-token keys out of the authority-issuer branch. `ValidIssuers` contains both the OIDC authority and the scope issuer, and `ValidAlgorithms` contains HS256 when scope tokens are enabled. For an authority-issued token this fallback concatenates `_scopeKeys`, so a holder of the scope-service HMAC key can sign an HS256 token with `iss` set to the authority and arbitrary user claims, and it passes the intended issuer/key boundary. Please model the authority issuers explicitly and return only JWKS keys for them; unknown issuers should fail closed instead of receiving a superset that includes scope keys.

src/Aevatar.Foundation.Runtime.Persistence.Implementations.Garnet/SecretStoreKeyringDocument.cs
```diff
@@ -0,0 +1,147 @@
...
+
+        var fullPath = Path.GetFullPath(path);
+        var directory = Path.GetDirectoryName(fullPath);
+        if (!string.IsNullOrWhiteSpace(directory))
+            Directory.CreateDirectory(directory);
+
+        var tempPath = $"{fullPath}.{Guid.NewGuid():N}.tmp";
+        try
+        {
+            File.WriteAllText(tempPath, JsonSerializer.Serialize(this, JsonOptions));
```
eanz17: [P1] Create this keyring with owner-only permissions before moving it into place. The document contains every AES data key plus the HMAC fingerprint key in plaintext Base64, but `File.WriteAllText` uses the process umask and commonly leaves the temp/final file mode as 0644. Any local user could then decrypt all Garnet vault records. The local secrets writer in this PR already clamps its temp file to 0600; this path needs the same treatment and should reject or repair an existing over-permissive keyring.

src/Aevatar.Mainnet.Host.Api/Hosting/MainnetHostBuilderExtensions.cs
```diff
@@ -174,6 +184,11 @@ public static WebApplicationBuilder AddAevatarMainnetHost(
         builder.Services.TryAddSingleton<IResponsesCallerScopeResolver, NyxIdResponsesCallerScopeResolver>();
         builder.Services.Configure<ResponsesNyxIdIdentityAssertionOptions>(
             builder.Configuration.GetSection(ResponsesNyxIdIdentityAssertionOptions.SectionName));
+        // Single-use jti replay guard for NyxID identity assertions. In-memory + node-local by
+        // default (bounded by each assertion's short lifetime); swap for a distributed backing
+        // (e.g. Garnet) to make replay detection cluster-wide without touching callers.
+        builder.Services.TryAddSingleton<IIdentityAssertionReplayGuard>(
```
eanz17: [P1] The Mainnet default must enforce replay atomically across replicas. The same valid identity assertion can be sent to two host instances and both node-local dictionaries accept its `jti`, so the documented single-use guarantee disappears in the distributed deployment this host composes. No distributed `IIdentityAssertionReplayGuard` is registered elsewhere in the repo. Please back this with a shared set-if-absent plus TTL store, and keep the in-memory guard only for development/tests.

tools/secret-store/RedisSecretStoreSweepTarget.cs
```diff
@@ -0,0 +1,119 @@
...
+        CancellationToken ct = default)
+    {
+        ArgumentException.ThrowIfNullOrWhiteSpace(pattern);
+        if (cursor < 0)
+            throw new ArgumentOutOfRangeException(nameof(cursor));
+        if (count <= 0)
+            throw new ArgumentOutOfRangeException(nameof(count));
+        ct.ThrowIfCancellationRequested();
+
+        var result = await _server.ExecuteAsync(
```
eanz17: [P1] Run SCAN against the configured database. `_database` uses `GetDatabase(database)`, but this raw server command has no database selection and executes against the connection default database. With `--database 1`, for example, scanning happens in DB 0 while GET/CAS happen in DB 1, so the reencryption sweep can find zero keys and exit successfully without rotating anything. Use a database-aware scan API such as `IServer.Keys(database, ...)` or a connection whose default database matches the requested value.

src/Aevatar.Audit.Core/Projection/ProjectionAuditTrailAppender.cs
```diff
@@ -0,0 +1,96 @@
...
+            var existing = await _store.GetAsync(auditId, ct);
+            if (existing != null)
+            {
+                return string.Equals(existing.ContentHash, contentHash, StringComparison.Ordinal)
+                    ? AuditTrailAppendResult.Duplicate(auditId)
+                    : AuditTrailAppendResult.Conflict(auditId, "Audit id already exists with different content.");
+            }
+
+            var observedAt = record.OccurredAt?.ToDateTimeOffset() ?? DateTimeOffset.UtcNow;
+            var document = AuditTrailDocumentFactory.Create(record, auditId, contentHash, observedAt);
```
eanz17: [P1] Apply `AuditRecordSanitizer` before hashing and creating the production document. The sanitizer is registered in DI but its only consumer is `InMemoryAuditTrailStore`; the Elasticsearch artifact store persists `document.Record` verbatim. Production can therefore retain secret-shaped annotations and unbounded/error text that development rejects, which also means tests exercise different semantics. Sanitization and validation need to happen once in this appender before every store implementation.

src/Aevatar.Mainnet.Host.Api/Hosting/HttpOAuthClientEsAclProbe.cs
```diff
@@ -0,0 +1,254 @@
...
+        {
+            return EsAclProbeResult.Unverifiable(
+                $"Elasticsearch _has_privileges returned {(int)response.StatusCode}; cannot classify the '{IndexName}' read grant.");
+        }
+
+        var hasRead = TryReadIndexHasRead(body);
+        // Security layer is reachable and authenticated. Whether the internal
+        // identity itself has read or not, the presence of an enforcing security
+        // layer means the index is not world-readable at the cluster level.
+        return EsAclProbeResult.Restricted(
```
eanz17: [P1] This does not verify that the OAuth-client index read grant is restricted. `_has_privileges` only tells us whether the current internal identity has `read`; it cannot tell whether other identities or wildcard roles also have it, and this branch returns `Restricted` even when `hasRead` is false. `ProbeRoleMappingsAsync` only lists mapping names and does not inspect role/index privileges. A security-enabled cluster with a broad read role therefore passes Strict. Please inspect the effective role definitions/mappings for this index, or report `Unverifiable` rather than claiming a restricted grant.

src/platform/Aevatar.GAgentService.Infrastructure/Schedules/ScheduledServiceInvocationDispatchPort.cs
```diff
@@ -90,15 +99,51 @@ private async Task<ServiceInvocationRequest> BuildInvocationRequestAsync(
...
+        CredentialRole role,
+        string token,
+        CancellationToken ct)
+    {
+        if (_secretVault == null)
+            throw new InvalidOperationException("Scheduled workflow caller credential vault is not configured.");
+
+        var ownerScopeKey = ResolveOwnerScopeKey(dispatch);
+        var subjectId = ResolveSubjectId(dispatch.Auth, role);
+        var stored = await _secretVault.PutAsync(new StoreSecretRequest(
```
eanz17, line 286: [P1] Bound the lifetime and cleanup of this per-fire bearer secret. Every recurring workflow fire mints a new access token and `PutAsync` creates a fresh `sec_` reference, but this request provides no `ExpiresAt` and there is no revoke call for `WorkflowCallerDurableBearerToken` anywhere in the repository. This leaves an unbounded set of durable bearer ciphertexts and permits vault resolution long after the JWT lifetime. Carry the exchange expiry (or parse `exp`), set it on the secret, and give run completion a revoke/cleanup owner.

src/platform/Aevatar.GAgentService.Application/Schedules/ScheduledDispatchApplicationService.cs
```diff
@@ -41,10 +52,19 @@ public async Task<ScheduledDispatchMutationReceipt> CreateAsync(
...
-        var normalized = NormalizeConfiguration(configuration, requireScheduleId: true);
-        ValidateSchedule(normalized);
+        var normalized = await NormalizeAndAdmitMutationAsync(configuration, context, requireScheduleId: true, ct);
+        // A deleted schedule is a permanent tombstone: the actor rejects any
+        // reconfigure, and the admission-only dispatch would swallow that
+        // rejection — an unguarded ensure would return an accepted receipt for
+        // a schedule that never materializes. Surface the tombstone as the same
+        // typed not-found the mutators throw, so callers can pick a fresh id.
+        await EnsureMutableAsync(normalized.ScheduleId, ct);
+        normalized = AdmitCredentialRequirement(
```
eanz17, line 74: [P1] This admission now rejects the shipped `self_reschedule` path before the actor can preserve or establish credentials. `WorkflowSelfRescheduleModule` builds a `WorkflowScheduleConfiguration` with `Auth == null` and no mutation context; the mapper classifies it as `WorkflowService`, and the default policy denies credential source `None`. As a result the checked-in Firecrawl workflows fail every `self_reschedule` ensure, while existing schedules never reach `PreserveExistingServiceInvocationAuth` in `ScheduledDispatchGAgent`. Please make the module supply a typed source for new schedules and allow an auth-omitting ensure/update to preserve the existing actor-owned source.

src/workflow/Aevatar.Workflow.Core/Execution/WorkflowExecutionKernel.cs
```diff
@@ -1141,14 +1205,91 @@ await WorkflowRuntimeCallbackLeaseSupport.TryCancelAsync(
...
+        };
+        if (!requestPublishSucceeded)
+        {
+            await TryStartCompensationOrPublishTerminalFailureAsync(
+                ctx,
+                terminalFailure,
+                state,
+                terminalStep: null,
+                ct,
+                knownNoCompensableLedger: true);
```
eanz17: [P1] Do not assert that the whole run has no compensable ledger merely because this step was not published. A concrete run can have step A complete with a recorded compensation entry, then fail while preparing/publishing step B. This flag fabricates `NoCompensableLedger`, skips `_stateHost.TryStartCompensationAsync`, cleans up, and publishes terminal failure without compensating A. Please query the run-level ledger here; only the unpublished current step is known to have no new entry.

src/Aevatar.Mainnet.Host.Api/Voice/voice-console.html
```diff
@@ -808,29 +959,68 @@ async function startVoice(){
...
+  };
+  if(reason && typed[reason]) return {status:"error",detail:typed[reason]};
+  if(code===1000) return {status:"ended",detail:"会话已结束"};
+  if(reason) return {status:"error",detail:"连接已关闭："+reason};
+  if(code===1006) return {status:"error",detail:"连接被服务器拒绝或异常中断"};
+  return {status:"error",detail:"连接已关闭（code "+code+"）"};
+}
+
+function dialVoice(token, retriesRemaining){
+  if(voice.status==="ended") return;
```
eanz17, line 1115: [P1] Tear down audio when the user stops while the microphone permission prompt is open. `stopVoice()` runs before `voice.micStream` exists, but after permission is granted `startVoice()` assigns the stream and creates both AudioContexts, then this early return leaves the microphone track recording indefinitely. Recheck `voice.status` immediately after `getUserMedia` and call `teardownAudio()` on this early-exit path.

src/workflow/Aevatar.Workflow.Application/Observatory/WorkflowRunObservatoryQueryService.cs
```diff
@@ -284,6 +335,250 @@ private static ObservatoryRunStatistics ToStatistics(WorkflowRunStatistics summa
...
+
+        if (!string.IsNullOrWhiteSpace(snapshot.LastError))
+        {
+            AppendDistinct(diagnostics, new ObservatoryRunDiagnostic
+            {
+                TimestampUtc = NonDefault(snapshot.LastUpdatedAt),
+                Severity = "error",
+                Code = "current_state_last_error",
+                Source = "current-state",
+                Message = snapshot.LastError,
```
eanz17: [P1] Sanitize current-state errors before adding them to the observatory response. `WorkflowRunState.FinalError` and `DeadLetterError` are projected verbatim, while this PR applies `WorkflowAuditTextSanitizer` to the equivalent run-report fields. On the report-lag/fallback path an upstream exception that embeds an Authorization token or API key is therefore returned unredacted here, including through the new admin cross-scope endpoint. Apply the same sanitizer to both current-state diagnostic messages.

agents/Aevatar.GAgents.Authoring.Lark/ScheduledAgentCreateRequestMapper.cs
```diff
@@ -349,10 +368,76 @@ private static bool ValidateOneShotDelay(TimeSpan delay, out string error)
...
+
+    private static UserAgentCatalogUpsertCommand BuildCatalogUpsertCommand(
+        ScheduledAgentCreatePlannedRequest request,
+        ScheduledAgentApiKeyIssueResult issuedKey) =>
+        new()
+        {
+            AgentId = request.AgentId,
+            ConversationId = request.ConversationId,
+            NyxProviderSlug = request.PrimaryOutboundSlug,
+            NyxApiKey = issuedKey.FullKey,
```
eanz17: [P1] Pass `storedKey.Reference` into the catalog command instead of the raw key. This new producer sets only deprecated `NyxApiKey`, while `UserAgentCatalogGAgent.HandleUpsertAsync` deliberately writes that field as empty and copies only `NyxApiKeyReference`. Consequently every newly created scheduled agent gets no credential projection, and `UserAgentDeliveryTargetReader` returns null for outbound delivery/revocation. `BuildCatalogUpsertCommand` should accept the `storedKey` result and populate `NyxApiKeyReference` without putting `FullKey` in the command.

src/Aevatar.AI.Core/Auditing/ToolCallReceiptFinalizer.cs
```diff
@@ -0,0 +1,152 @@
...
+            return new FinalizedToolCallReceipt(NormalizeReceipt(context, context.Receipt), IsSynthetic: false);
+
+        if (exception != null)
+        {
+            return new FinalizedToolCallReceipt(
+                CreateMinimalReceipt(
+                    context,
+                    AgentToolReceiptStatus.Error,
+                    errorCode: "tool_execution_exception",
+                    errorMessage: exception.Message),
```
eanz17: [P1] Do not copy `exception.Message` into a persisted governance record. This flows through `ToolAuditRecordFactory` into `AuditRecord.ErrorSummary`; `AuditRecordSanitizer` only normalizes/truncates summaries and does not redact tokens, URLs, response bodies, or credentials. Provider exceptions routinely contain those values. Persist an allowlisted error code/class here and keep the full exception in operational logs.

src/Aevatar.Bootstrap/Hosting/EndpointAuditCaptureMiddleware.cs
```diff
@@ -0,0 +1,354 @@
...
+                _logger.LogError(
+                    "Endpoint audit summary capture failed for operation {OperationName}.",
+                    metadata.OperationName);
+            }
+
+            var outcome = EndpointAuditOutcomeClassifier.Classify(context, capturedException);
+            await AppendBestEffortAsync(
+                _appender,
+                () => BuildRecord(context, metadata, _identityHasher, metadata.OperationName, outcome),
+                context.RequestAborted);
```
eanz17: [P1] The terminal append cannot use `RequestAborted`. On disconnect/cancellation this token is already cancelled when `finally` runs; `AppendAsync` throws and `AppendBestEffortAsync` swallows it, so the audit trail retains only `.attempted` and loses the terminal fact. Use a short host-owned timeout here, and classify `OperationCanceledException` as `AuditOutcome.Cancelled` rather than `Error`.

src/Aevatar.Studio.Application/Studio/Services/StudioMemberWorkflowBindingPort.cs
```diff
@@ -0,0 +1,223 @@
...
+        var receipt = await _memberService.BindAsync(
+            request.ScopeId,
+            request.MemberId,
+            new UpdateStudioMemberBindingRequest(
+                Workflow: new StudioMemberWorkflowBindingSpec(
+                    workflowId,
+                    [request.WorkflowYaml])),
+            ct);
+
+        var terminalRun = await WaitForTerminalBindingRunAsync(receipt, ct);
```
eanz17: [P1] This receipt only guarantees `dispatch_accepted`; waiting up to 175 seconds on an eventually consistent read model can report `invalid_arguments`/timeout after the binding command was already accepted and may still succeed, prompting a duplicate retry. Return the accepted receipt with `bindingRunId`/status URL here and observe completion asynchronously (event/push or a later query) instead of assembling a synchronous workflow inside the tool call.

agents/Aevatar.GAgents.Channel.Identity/Provisioning/AevatarOAuthClientResources.cs
```diff
@@ -0,0 +1,29 @@
...
+/// RFC 8707 resource indicators required by aevatar's NyxID OAuth flows.
+/// </summary>
+public static class AevatarOAuthClientResources
+{
+    public const string RequiredServiceSlug = "aevatar";
+
+    public static string RequiredServiceResourceUri(string nyxIdAuthority)
+    {
+        ArgumentException.ThrowIfNullOrWhiteSpace(nyxIdAuthority);
+        return $"{nyxIdAuthority.TrimEnd('/')}/api/v1/proxy/s/{RequiredServiceSlug}";
```
eanz17: [P1] Build this resource from NyxID's canonical API base discovered from its issuer/resource metadata, not the configured login authority. The default authority is `https://nyx.chrono-ai.fun`, while production discovery and `BASE_URL` use `https://nyx-api.chrono-ai.fun`; NyxID matches the exact `BASE_URL/api/v1/proxy/s/` prefix. These new authorize/token/exchange requests therefore send an invalid resource and return `invalid_target`. Please add a split-origin test.

src/Aevatar.Authentication.Hosting/AevatarAuthenticationHostExtensions.cs
```diff
@@ -130,6 +157,14 @@ private static void ConfigureScopeServiceTokenValidation(
         var signingKeys = keyProvider.ValidationKeys.Select(key => key.ValidationKey).ToArray();
         jwt.TokenValidationParameters.IssuerSigningKeys = signingKeys;
 
+        // Defense-in-depth: bind the scope-token issuer to the scope-token keys only. Tokens
+        // from an unknown/OIDC issuer fall back to all configured keys, so nothing that
+        // validates today stops validating (see PerIssuerSigningKeyResolver).
+        var resolver = new PerIssuerSigningKeyResolver(
+            [keyProvider.Issuer],
+            signingKeys);
+        jwt.TokenValidationParameters.IssuerSigningKeyResolver = resolver.Resolve;
```
eanz17: [P1] Installing the four-argument resolver bypasses IdentityModel's normal `BaseConfiguration.SigningKeys` lookup. In the current JwtBearer pipeline discovery is attached as a configuration manager; JWKS keys are not copied into `IssuerSigningKeys`. This resolver returns the non-empty scope-key set for an authority token, so the discovery key is never tried and normal NyxID tokens fail whenever scope tokens are enabled. Use `IssuerSigningKeyResolverUsingConfiguration` and include `configuration.SigningKeys`, with an HTTP-pipeline discovery test.

src/Aevatar.Authentication.Hosting/DPoPProofValidator.cs
```diff
@@ -0,0 +1,349 @@
...
+        // (e) jti must be present; replay is delegated to the injectable guard.
+        var jti = proof.Payload.TryGetValue("jti", out var jtiValue) ? jtiValue?.ToString() : null;
+        if (string.IsNullOrWhiteSpace(jti))
+            return DPoPValidationResult.Fail("dpop_jti_missing", "DPoP proof is missing jti.");
+
+        var fresh = await _replayGuard.TryRegisterAsync(jti, cancellationToken);
+        if (!fresh)
+            return DPoPValidationResult.Fail("dpop_jti_replayed", "DPoP proof jti has already been used.");
+
+        return DPoPValidationResult.Success;
```
eanz17, line 177: [P1] A protected-resource DPoP proof must include and verify `ath` against the presented access token (RFC 9449 section 7). `ValidateAsync` never receives that token and returns success after `htm`/`htu`/`iat`/`jti`, so an unused proof minted for token A can authorize token B when both are bound to the same proof key. Pass the raw access token, require `ath`, and fixed-time compare it with `base64url(SHA-256(access_token))`.

src/Aevatar.Authentication.Hosting/AevatarAuthenticationHostExtensions.cs
```diff
@@ -61,8 +67,16 @@ public static WebApplicationBuilder AddAevatarAuthentication(this WebApplication
...
+                // Pin the accepted signing algorithms to block alg=none and downgrade attacks.
+                // Default is a safe superset covering every algorithm actually accepted; when
+                // scope service tokens are enabled it also allows their HS256/RS256 keys.
+                jwt.TokenValidationParameters.ValidAlgorithms =
+                    ResolveValidAlgorithms(options, scopeTokenOptions.Enabled);
+
                 jwt.Events = new JwtBearerEvents
                 {
+                    OnTokenValidated = OnTokenValidatedValidateDPoP,
                     OnMessageReceived = context =>
```
eanz17, line 82: [P1] Extract `Authorization: DPoP <access-token>` here. JwtBearer's default extraction accepts only the `Bearer` scheme, while NyxID returns `token_type=DPoP` for `cnf.jkt` tokens and compliant clients use the DPoP scheme. Such requests currently produce `NoResult`, so `OnTokenValidated` and the new proof validator are never reached. Please cover this through the actual HTTP authentication pipeline.

src/Aevatar.AI.ToolProviders.NyxId/ObservatoryAdminAuthorizationOptions.cs
```diff
@@ -1,16 +1,16 @@
 namespace Aevatar.AI.ToolProviders.NyxId;
 
-// 06-20-observatory-admin-cross-scope (G8): tuning for the platform-admin authorizer's per-token cache.
 public sealed class ObservatoryAdminAuthorizationOptions
 {
-    public const string ConfigSection = "Aevatar:Observatory";
+    public const string ConfigSection = "Aevatar:AdminAccess";
```
eanz17, line 5: [P1] This silently relocates the existing security kill switch. A deployment with `Aevatar__Observatory__CrossScopeEnabled=false` now ignores that value and falls back to the new default `true`, re-enabling cross-scope admin access. Migrate the deployment value atomically, bind the old section during the transition, or fail startup when the retired setting is present.

agents/channels/Aevatar.GAgents.Channel.NyxIdRelay/Outbound/RemoteToolApprovalMessageMapper.cs
```diff
@@ -0,0 +1,70 @@
...
+        };
+
+        if (notification.Submission.ExpiresAt is { } expiresAt)
+            lines.Add($"Expires: `{expiresAt:O}`");
+
+        if (!string.IsNullOrWhiteSpace(request.ArgumentsJson))
+        {
+            lines.Add(string.Empty);
+            lines.Add("Arguments:");
+            lines.Add($"```json\n{request.ArgumentsJson}\n```");
```
eanz17: [P1] Do not send `ArgumentsJson` verbatim. Tool arguments can contain passwords, API keys, authorization headers, SSH commands, or private request bodies, and this delivery target can be a Lark group conversation. Render an allowlisted/redacted summary with a structured JSON scrubber before composing the card.

src/Aevatar.Audit.Core/Stores/InMemoryAuditTrailStore.cs
```diff
@@ -0,0 +1,242 @@
...
+    public Task<AuditTrailAppendResult> AppendAsync(
+        AuditRecord record,
+        CancellationToken cancellationToken = default)
+    {
+        cancellationToken.ThrowIfCancellationRequested();
+
+        var sanitized = _sanitizer.Sanitize(record);
+        lock (_records)
+        {
+            _records.Add(sanitized.Clone());
```
eanz17: [P2] Direct development appends must preserve the same audit-id idempotency as production. This type is registered directly as `IAuditTrailAppender`, but every retry unconditionally adds another record and returns `Appended`; the production appender returns `Duplicate` or `Conflict`. Route this path through the existing content-hash/`UpsertAsync` check so development and tests do not hide retry semantics or produce duplicate governance facts.

src/Aevatar.Authentication.Hosting/DPoPProofValidator.cs
```diff
@@ -0,0 +1,349 @@
...
+
+        iat = DateTimeOffset.FromUnixTimeSeconds(seconds);
+        return true;
+    }
+
+    private static bool FixedTimeEquals(string left, string right)
+    {
+        var leftBytes = Encoding.UTF8.GetBytes(left);
+        var rightBytes = Encoding.UTF8.GetBytes(right);
+        return CryptographicOperations.FixedTimeEquals(leftBytes, rightBytes);
```
eanz17, line 380: [P2] Guard length before `CryptographicOperations.FixedTimeEquals`.

`FixedTimeEquals` throws `ArgumentException` when the two byte arrays differ in length. A crafted / truncated `cnf.jkt` (or mismatched thumbprint encoding) therefore escapes the normal `DPoPValidationResult.Fail(...)` path and bubbles out of `ValidateAsync` into `OnTokenValidated` as an unhandled exception — typically a 500 instead of a clean auth failure, and potentially a timing/exception oracle around thumbprint shape.

Match `AuditActorIdentityHasher`: return false when lengths differ (after encoding), then call `FixedTimeEquals` only on equal-length buffers.

src/Aevatar.Audit.Core/Stores/InMemoryAuditTrailStore.cs
```diff
@@ -0,0 +1,242 @@
...
+                string.Equals(candidate.AuditId, document.AuditId, StringComparison.Ordinal));
+            if (existing is not null)
+            {
+                return Task.FromResult(string.Equals(existing.ContentHash, document.ContentHash, StringComparison.Ordinal)
+                    ? AuditTrailArtifactWriteResult.Duplicate()
+                    : AuditTrailArtifactWriteResult.Conflict());
+            }
+
+            _records.Add(record.Clone());
+            _documents.Add(document.Clone());
```
eanz17: [P2] Persist the sanitized record into the document, not the raw input.

`Sanitize(document.Record)` only feeds `_records`; `_documents.Add(document.Clone())` still stores the original (possibly secret-bearing) `document.Record`. Any path that reads via `GetAsync` / document store therefore bypasses the sanitizer that `AppendAsync` applies.

Build the stored document from `record` (sanitized) and recompute content hash from that payload so query and get share one redacted fact.

src/Aevatar.Audit.Hosting/AuditTrailEndpoints.cs
```diff
@@ -0,0 +1,326 @@
...
+        string targetScope,
+        PlatformCaller admin,
+        string? reason) =>
+        logger.LogInformation(
+            "audit_trail_admin_read outcome={Outcome} action={Action} adminUserId={AdminUserId} adminEmail={AdminEmail} " +
+            "role={Role} grantSource={GrantSource} callerScope={CallerScope} targetScope={TargetScope} reason={Reason} correlationId={CorrelationId}",
+            outcome,
+            action,
+            admin.UserId,
+            admin.Email,
```
eanz17: [P2] Avoid logging admin email in application logs.

`adminEmail={AdminEmail}` lands in host stdout / centralized log sinks (ES, Loki, etc.) for every cross-scope audit read. Governance identity should stay on the hashed `audit_actor_id` / stable `adminUserId` plane; email is PII and often used as a recovery identifier.

Prefer `adminUserId` + `grantSource` only; if email is required for forensics, hash it with the same host-owned HMAC key id used for audit actor identity.

src/Aevatar.Mainnet.Host.Api/Hosting/MainnetAgentProjectionDocumentStoresExtensions.cs
```diff
@@ -43,6 +57,18 @@ public static IServiceCollection AddMainnetAgentProjectionDocumentStores(
         if (documentProvider.ElasticsearchEnabled)
         {
             AddElasticsearchStores(services, configuration);
+            services.Configure<AevatarOAuthClientEsAclOptions>(options =>
+            {
+                options.EnforcementMode = AevatarOAuthClientEsAclEnforcementMode.Strict;
```
eanz17: [P1] Enabling `EnforcementMode=Strict` here does not actually fail closed in production.

Mainnet also hardcodes `GrantMatchesGrainEventStoreInternal = true` (composition), and `HttpOAuthClientEsAclProbe` reports `Restricted` whenever the security API is reachable — not when the index grant is exclusive to grain/event-store identities (already flagged on the probe). Strict then only throws on a definitive `Unrestricted` or a *false* attestation.

Net effect: a world-readable `aevatar-oauth-clients` index (HMAC material / vault refs) can still pass startup. Either:
1. Probe must positively prove exclusivity (not just “security on”), **and**
2. Strict must fail closed on `Unverifiable`/`Unavailable` unless an *operator-supplied* attestation is present (not code-hardcoded true).

agents/Aevatar.GAgents.Channel.Identity/Provisioning/AevatarOAuthClientEsAclStartupGuard.cs
```diff
@@ -49,17 +69,86 @@ public Task StartAsync(CancellationToken cancellationToken)
...
+
+                if (!attestationDeclaredRestricted)
+                {
+                    throw new InvalidOperationException(
+                        BuildFailureMessage(
+                            $"the operator attestation {AevatarOAuthClientEsAclOptions.SectionName}:GrantMatchesGrainEventStoreInternal is false. " +
+                            $"probeStatus={probeResult.Status} observedState={probeResult.ObservedState}"));
+                }
+
+                if (probeResult.Status is EsAclProbeStatus.Unavailable or EsAclProbeStatus.Unverifiable)
```
eanz17: [P1] Strict mode must not accept `Unverifiable` / `Unavailable` with only attestation.

When the live probe cannot classify the grant, this branch logs a warning and continues as long as `GrantMatchesGrainEventStoreInternal` is true. Mainnet sets that flag in code, so the “fail closed” Strict path degrades to “fail open with a warning” on every transport/auth/security-API hiccup — exactly when you most need to stop serving secrets from ES.

For Strict: throw on anything other than a positive `Restricted` result. Keep attestation as an *additional* check, never a substitute for a live restricted classification.

PR #2631, opened 2026-07-08
Fix missing member published-runs redirect

AbigailDeng:
## Problem

Published-runs member routes kept rendering the replay shell when the routed Team member no longer existed. The page then continued to request member runs and audit data, surfacing `STUDIO_MEMBER_NOT_FOUND` / unavailable error panels instead of leaving the stale member route.

Fixes #2630.

## Solution

- Treat `STUDIO_MEMBER_NOT_FOUND` from `studioApi.getMember` as a stale member route and replace navigation to `/scopes`.
- Gate member runs and audit loading behind a successful member lookup so missing members do not trigger follow-up runs/audit requests.
- Preserve deep-link behavior by only loading audit from a `runId`-only URL after the matching run is present in the runs catalog; URLs that include `actorId` can still load directly.
- Add a regression test for missing member redirect and tighten the loading test around the new member-first flow.

## Impact paths

- `apps/aevatar-console-web/src/pages/runtime-published-runs/MemberPublishedRunsReplay.tsx`
- `apps/aevatar-console-web/src/pages/runtime-published-runs/index.test.tsx`

## Verification

- `./node_modules/.bin/jest --selectProjects jsdom --runInBand --runTestsByPath src/pages/runtime-published-runs/index.test.tsx`
- `./node_modules/.bin/tsc --noEmit`
- `git diff --check`

Note: direct `pnpm --dir apps/aevatar-console-web ...` initially stopped in this fresh worktree at pnpm's `ERR_PNPM_IGNORED_BUILDS` install gate. After dependencies were present, the same checks were run through local `node_modules/.bin`.

## Recurrence prevention

- Immediate symptom: missing Team member published-runs routes showed unavailable error panels.
- Root cause: runs/audit loaded before confirming the routed member existed.
- General failure pattern: child resource pages can leak child API failures when parent resource existence is not established first.
- Similar locations searched: `STUDIO_MEMBER_NOT_FOUND`, member routes, published runs, invoke/workflow studio member-loading call sites.
- Guard added: member lookup gates runs/audit queries; missing member routes replace to console home.
- Regression test added: missing member redirect test asserts `/scopes` and no runs/audit calls.
- Hard gate: targeted jsdom test, TypeScript, production build, and diff whitespace check.
- Remaining risks: no live browser session against the deployed backend in this PR.

## FKST provenance

- Issue: #2630
- Worktree: /Users/xiezixin/Documents/work/aevatar-fix-missing-member-redirect
- Branch: fix/2026-07-08_missing-member-redirect
- Operator: AbigailDeng
- Freshness: Current user reproduction from July 8, 2026 on `origin/dev`.

### DIFF
--- a/apps/aevatar-console-web/src/pages/runtime-published-runs/MemberPublishedRunsReplay.tsx
+++ b/apps/aevatar-console-web/src/pages/runtime-published-runs/MemberPublishedRunsReplay.tsx
@@ -1448,7 +1448,9 @@ const MemberPublishedRunsReplay: React.FC<MemberPublishedRunsReplayProps> = ({
               loading={auditQuery.isFetching}
               onClick={() => {
                 runsQuery.refetch();
-                auditQuery.refetch();
+                if (auditCanLoad) {
+                  auditQuery.refetch();
+                }
               }}
             >
               {t("pages.runs.memberPublishedRuns.refresh", "Refresh")}
--- a/apps/aevatar-console-web/src/pages/runtime-published-runs/MemberPublishedRunsReplay.tsx
+++ b/apps/aevatar-console-web/src/pages/runtime-published-runs/MemberPublishedRunsReplay.tsx
@@ -36,6 +36,7 @@ import {
   buildTeamDetailHref,
   buildTeamMemberPublishedRunsHref,
   buildTeamMemberWorkflowStudioHref,
+  buildTeamsHref,
 } from "@/shared/navigation/teamRoutes";
 import GraphCanvas from "@/shared/graphs/GraphCanvas";
 import type {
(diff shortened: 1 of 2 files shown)


apps/aevatar-console-web/src/pages/runtime-published-runs/MemberPublishedRunsReplay.tsx
```diff
@@ -1098,7 +1135,7 @@ const MemberPublishedRunsReplay: React.FC<MemberPublishedRunsReplayProps> = ({
   }, [memberId, normalizedInitialRunId, normalizedTeamId, scopeId, selectedCatalogRun]);
 
   const auditQuery = useQuery({
-    enabled: Boolean(scopeId && memberId && selectedRunId),
+    enabled: auditCanLoad,
```
potter-sun, line 1138: [P1] This only gates the automatic query. `auditQuery.refetch()` in the Refresh handler still calls the query function even when `enabled` is false in TanStack Query v5, so a `runId`-only URL whose run is not present in the catalog can still hit `getMemberRunAudit(..., { actorId: undefined })` when the user clicks Refresh. That bypasses the catalog-match guard added here. Please guard the refresh handler with `auditCanLoad` or disable the audit refresh until the catalog match exists, and add a regression for `runId`-only + catalog miss + Refresh.

AbigailDeng: [from Codex]: Fixed in 89c70b845a0fa239d7fed3b75c1518f16fb51ac7. The stage Refresh handler still refreshes the run catalog, but now only calls auditQuery.refetch() when auditCanLoad is true, so a runId-only catalog miss no longer calls getMemberRunAudit with actorId undefined. Added a regression for runId-only + catalog miss + Refresh.

PR #2640, opened 2026-07-08
Fix Mission Wall workflow retention

AbigailDeng:
## Problem

Mission Wall could keep the selected workflow card visible while the right-side Step Flow turned into an empty grid after the page had been open for a while.

Root cause:
- `WorkflowReplayCanvas` treated ReactFlow viewport fitting as a one-shot operation. If the first `fitView` ran before the node/container measurement settled, the viewport could stay stranded on an empty grid for the same graph revision.
- The Mission Wall data hook also consumed transient empty workflow-board snapshots directly, so a short read-model hiccup could clear visible workflow state.

## Solution

- Retry the Mission Wall ReactFlow `fitView` across several animation frames for each graph revision, then clean up queued frame/timeout work when the graph changes or unmounts.
- Cache the last non-empty workflow-board snapshot per `scopeId + teamId` for a short 60s fallback window.
- Mark the wall as `Degraded` while serving that stale fallback, so the UI stays useful without pretending the read model is fresh.
- Add regression coverage for both the ReactFlow fit retry and the transient empty snapshot case.

## Impact paths

- `apps/aevatar-console-web/src/pages/MissionWall/components/WorkflowReplayCanvas.tsx`
- `apps/aevatar-console-web/src/pages/MissionWall/hooks/useMissionWallData.ts`
- `apps/aevatar-console-web/src/pages/MissionWall/components/WorkflowReplayCanvas.test.tsx`
- `apps/aevatar-console-web/src/pages/MissionWall/index.test.tsx`

## Verification

- `./node_modules/.bin/tsc --noEmit --pretty false`
- `./node_modules/.bin/jest --runInBand --watchman=false src/pages/MissionWall/index.test.tsx`
- `./node_modules/.bin/jest --runInBand --watchman=false src/pages/MissionWall/components/WorkflowReplayCanvas.test.tsx`
- `./node_modules/.bin/jest --selectProjects jsdom --runInBand --watchman=false`  
  Result: 112 suites passed, 924 tests passed.
- `git diff --check`

Note: direct `pnpm --dir apps/aevatar-console-web ...` initially hit the local pnpm ignored-builds install gate in this fresh worktree, so validation used the already materialized local package binaries. Build and tests passed.

## FKST provenance

- Issue: #2639
- Worktree: /Users/xiezixin/Documents/work/aevatar-mission-wall-retention
- Branch: fix/2026-07-08_mission-wall-workflow-retention
- Operator: AbigailDeng
- Freshness: Current user-reported Mission Wall issue from 2026-07-08, fixed against current origin/dev evidence.


apps/aevatar-console-web/src/pages/MissionWall/hooks/useMissionWallData.ts
```diff
@@ -196,34 +221,66 @@ export function useMissionWallRuntimeData(): MissionWallRuntimeData {
...
+  const shouldUseCachedSnapshot =
+    Boolean(cachedSnapshot) &&
+    snapshotCacheRef.current !== undefined &&
+    nowMs - snapshotCacheRef.current.cachedAtMs <=
+      MISSION_WALL_STALE_SNAPSHOT_FALLBACK_MS &&
+    ((snapshotQuery.isSuccess && queriedRunCount === 0) || snapshotQuery.isError);
+  const effectiveSnapshot = shouldUseCachedSnapshot
+    ? cachedSnapshot
+    : queriedSnapshot;
+  React.useEffect(() => {
```
potter-sun, line 263: [P1] This effect keeps refreshing the fallback TTL during a persistent refetch outage. TanStack Query keeps the previous `data` after a failed refetch while flipping the query to `isError`, so `queriedSnapshot` can remain the last non-empty snapshot here. Because `nowMs` changes every second and is in this effect dependency list, `cachedAtMs` is rewritten on every tick, which means the intended 60s stale fallback window never expires and Mission Wall can show old workflows indefinitely. Please only update `snapshotCacheRef` from a successful fresh snapshot, for example by gating this effect on `snapshotQuery.isSuccess`, and add a test where repeated snapshot errors continue past `MISSION_WALL_STALE_SNAPSHOT_FALLBACK_MS`.

AbigailDeng: [from Codex]: Fixed in e2b106c98ce0e61b4c913ed9baaf92f41826084c. The Mission Wall snapshot cache is now refreshed only from successful fresh snapshot data using dataUpdatedAt, so the one-second clock no longer extends the stale fallback TTL. Refetch errors no longer keep showing TanStack Query retained data after the fallback window expires, and I added a regression that keeps errors past MISSION_WALL_STALE_SNAPSHOT_FALLBACK_MS.

PR #2641, opened 2026-07-08
github-devloop implementation for #2635

louis4li:
github-devloop implementation PR for issue #2635

### DIFF
--- a/src/Aevatar.AI.ToolProviders.NyxId/Tools/NyxIdCodeExecuteTool.cs
+++ b/src/Aevatar.AI.ToolProviders.NyxId/Tools/NyxIdCodeExecuteTool.cs
@@ -68,9 +68,9 @@ public sealed class NyxIdCodeExecuteTool : IAgentTool
         if (string.IsNullOrWhiteSpace(language) || string.IsNullOrWhiteSpace(code))
             return """{"error":"Both 'language' and 'code' are required."}""";
 
-        // Resolve sandbox slug: context → API discovery → known slugs → give up
-        var slug = ResolveSandboxSlugFromContext()
-                   ?? await DiscoverSandboxSlugAsync(token, ct);
+        // Resolve sandbox slug: context -> API discovery -> known slugs -> give up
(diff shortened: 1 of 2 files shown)


louis4li:
github-devloop PR review decision: reject
Three-angle verdicts: teleology=comment parsimony=comment fidelity=reject
Blocking gap: Proxy-over-truth: recovery is inferred from a different live-discovered slug instead of verifying the real post-recovery `code_execute` target succeeds.

teleology:
This should advance, with one advisory concern: the retry classifier is broader than the purpose strictly demands because `ShouldRetryWithLiveDiscoveredSlug` retries any proxy error carrying a `message` when no `status` is present, so a follow-up should narrow that to the known transient/proxy failure shape or add an idempotency/replay-safety argument for `code_execute`; however, the PR does address the issue’s bounded recovery purpose by invalidating a context-selected stale sandbox slug through live discovery and verifying the same-run retry path in `ExecuteAsync_ContextSandboxReturnsTransientError_RetriesLiveDiscoveredSandbox`.

parsimony:
This is mergeable in shape and does not violate parsimony enough to reject: it deletes the incident-specific temptation, avoids `1006` branching, and uses live service discovery as the recovery authority after a context slug fails. The one advisory cleanup is to narrow `TryGetProxyErrorMessage(proxyResponse, out _)` in `ShouldRetryWithLiveDiscoveredSlug`, because retrying merely because an error object has a `"message"` is a symptom branch; explicit transient status or equivalent transient code would be cleaner.

fidelity:
The PR should not advance as merge-ready because it replaces the source requirement with a stale-slug proxy: `ExecuteAsync` now retries only when the context slug fails and `DiscoverSandboxSlugAsync` returns a different slug, while the issue requires `code_execute` to recover after service recovery when same-run `nyxid_proxy` succeeds; the diff provides no source evidence that recovery equals slug change, and if the recovered service keeps the same slug the new code still returns the original `HTTP 500 internal_error (1006)`.



⟦AI:FKST⟧

Issue #2679, opened 2026-07-09
Studio workflow provisioning reports accepted while failed binds leave enabled schedules and duplicate resources


louis4li:
## Problem

Studio workflow provisioning currently behaves like a false success:

- Binding is asynchronous, but the Studio agent/facade reports `accepted` / `provisioned` without observing the binding result.
- In failing cases, the actual binding run can be `failed` while the user-facing text says provisioning succeeded.
- Each retry creates a new member plus schedule instead of reusing or reconciling the previous attempt.
- Failed retries leave enabled schedules behind; repeated attempts can accumulate orphaned members/schedules that keep firing.

## Evidence

- The provisioning service intentionally does not poll or observe binding completion; see `src/Aevatar.Studio.Application/Studio/Services/StudioWorkflowProvisioningService.cs` and `test/Aevatar.Studio.Tests/StudioWorkflowProvisioningServiceTests.cs` assertions that `GetBindingRunAsync` is never called.
- The response always returns `BindingStatus = accepted` once create/bind/schedule requests are accepted, even though the binding result is still pending and may later fail.
- `BuildScheduleConfiguration` creates schedules with `Enabled: true` before binding success is known.
- In observed retries, seven attempts left 14 resources (member + schedule per retry) while schedules remained enabled.

## Expected behavior

Provisioning should be honest about asynchronous state and should not accumulate active garbage resources after failed binds.

Recommended product/architecture behavior:

- Return an explicit pending/accepted state that is not presented as final success.
- Surface binding run status and failure reason to Studio/agent output.
- Create or enable schedules only after the workflow binding is actually usable, or create a disabled schedule and enable it via binding-success continuation.
- Retries should reuse/reconcile the prior member/binding attempt when appropriate, or perform cleanup/disablement of failed-attempt schedules.
- Provide an observability path from provisioning response to binding run/readmodel state.

## Acceptance criteria

- Studio user-facing output distinguishes `accepted/pending` from `bound/provisioned` and reports failed binding runs as failures.
- Schedules are not left enabled for failed or unusable workflow bindings.
- Retrying the same demo workflow does not create unbounded duplicate member+schedule resources.
- Tests cover failed binding after accepted receipt and verify schedule/resource cleanup or delayed enablement behavior.

PR #2681, opened 2026-07-09
github-devloop implementation for #2679

louis4li:
github-devloop implementation PR for issue #2679


eanz17:
Closing as superseded by feature/integrate commit 748f98e7d, which moved workflow schedule provisioning to an actor-owned durable intent and lifecycle-driven executor. This PR still has two correctness blockers against the current contract: a pending bind can leave a valid bind without ever creating the schedule, and a failed rebind can retain the prior enabled schedule. Rebasing this implementation would reintroduce behavior that the current base has already replaced. The branch is preserved for history; no code was deleted.

PR #2696, opened 2026-07-10
Fix unified loading animation

AbigailDeng:
## Summary

- Add a shared Aevatar loading component for dots, pulse dots, streaming cursor, and page loading states.
- Move loading keyframes into global namespaced styles so reused chat/runtime components animate outside page-local style blocks.
- Replace page-level loading and chat/AGUI streaming indicators with the shared component.

Closes #2695

## Impact paths

- apps/aevatar-console-web/src/shared/ui/AevatarLoading.tsx
- apps/aevatar-console-web/src/global.less
- apps/aevatar-console-web/src/pages/chat/*
- apps/aevatar-console-web/src/shared/agui/runtimeConversationPresentation.tsx
- apps/aevatar-console-web/src/app.tsx
- apps/aevatar-console-web/src/loading.tsx
- apps/aevatar-console-web/src/pages/auth/callback/index.tsx
- apps/aevatar-console-web/src/shared/auth/ProtectedRouteRedirectGate.tsx
- apps/aevatar-console-web/src/pages/scopes/invoke.tsx

## Validation

- `./node_modules/.bin/biome lint src/shared/ui/AevatarLoading.tsx src/shared/ui/AevatarLoading.test.tsx`
- `./node_modules/.bin/tsc --noEmit -p tsconfig.json`
- `./node_modules/.bin/jest --selectProjects jsdom --runTestsByPath src/shared/ui/AevatarLoading.test.tsx src/pages/chat/index.test.tsx --runInBand`
- `./node_modules/.bin/jest --selectProjects jsdom --runInBand`

Note: `pnpm --dir apps/aevatar-console-web install --frozen-lockfile` populated dependencies but exited with `ERR_PNPM_IGNORED_BUILDS`; validation above used the installed local binaries.

## FKST provenance

- Issue: #2695
- Worktree: /Users/xiezixin/Documents/work/aevatar-pr-loading
- Branch: fix/2026-07-10_unified-loading-animation
- Operator: AbigailDeng
- Freshness: current user report on 2026-07-10

### DIFF
--- a/apps/aevatar-console-web/src/global.less
+++ b/apps/aevatar-console-web/src/global.less
@@ -468,7 +468,9 @@ body.aevatar-studio-host
   .aevatar-loading-dot,
   .aevatar-loading-pulse-dot,
   .aevatar-loading-cursor {
-    animation-duration: 2.4s;
+    animation: none;
+    opacity: 1;
+    transform: none;
   }
 }
 
--- a/apps/aevatar-console-web/src/global.less
+++ b/apps/aevatar-console-web/src/global.less
@@ -360,6 +360,118 @@ body.aevatar-studio-host
   pointer-events: none;
 }
 
+@keyframes aevatarLoadingPulse {
+  0%,
+  100% {
+    opacity: 1;
+  }
+
+  50% {
+    opacity: 0.45;
+  }
+}
+
+@keyframes aevatarLoadingBlink {
+  50% {
+    opacity: 0;
+  }
+}
+
+@keyframes aevatarLoadingBounce {
+  0%,
+  80%,
+  100% {
+    opacity: 0.7;
+    transform: translateY(0);
+  }
+
+  40% {
+    opacity: 1;
+    transform: translateY(-3px);
+  }
+}
+
+.aevatar-loading-dots {
+  align-items: center;
+  color: var(--aevatar-loading-color, currentColor);
+  display: inline-flex;
+  gap: var(--aevatar-loading-gap, 6px);
+  line-height: 1;
+  min-height: var(--aevatar-loading-dot-size, 6px);
+  vertical-align: middle;
+}
+
+.aevatar-loading-dot {
+  animation: aevatarLoadingBounce 1s ease-in-out infinite;
+  background: var(--aevatar-loading-color, currentColor);
+  border-radius: 999px;
+  display: block;
+  flex: 0 0 auto;
+  height: var(--aevatar-loading-dot-size, 6px);
+  width: var(--aevatar-loading-dot-size, 6px);
+}
+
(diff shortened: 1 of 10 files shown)


apps/aevatar-console-web/src/shared/ui/AevatarLoading.tsx
```diff
@@ -0,0 +1,189 @@
...
+      aria-busy="true"
+      className={joinClassNames(
+        'aevatar-page-loading',
+        fullscreen && 'aevatar-page-loading-fullscreen',
+        className,
+      )}
+      role="status"
+      style={style}
+    >
+      <AevatarLoadingDots color="#2563eb" decorative size="large" />
```
potter-sun, line 187: [P2] This `role="status"` is silent for the current default path. Most new call sites (`app.tsx`, `loading.tsx`, `ProtectedRouteRedirectGate`) render `AevatarPageLoading` without `tip`, and the only child is `AevatarLoadingDots decorative`, which sets `aria-hidden`. That leaves a live status with no announced text. Please give the page loader a default accessible label or hidden fallback text, and add a test for the no-tip path.

apps/aevatar-console-web/src/global.less
```diff
@@ -360,6 +360,118 @@ body.aevatar-studio-host
...
+
+.aevatar-page-loading-tip {
+  line-height: 1.5;
+}
+
+@media (prefers-reduced-motion: reduce) {
+  .aevatar-loading-dot,
+  .aevatar-loading-pulse-dot,
+  .aevatar-loading-cursor {
+    animation-duration: 2.4s;
```
potter-sun: [P2] This still runs all loading animations for users who asked for reduced motion; it only slows them down. The new shared loader is now used in page bootstrap and streaming states, so `prefers-reduced-motion: reduce` should make the dots/cursor static or otherwise remove the translate/blink/pulse animation instead of keeping infinite motion.

apps/aevatar-console-web/src/shared/ui/AevatarLoading.tsx
AbigailDeng: Fixed in 5bbb3b6c6. `AevatarPageLoading` now has a default `ariaLabel` and renders a visually hidden fallback label when no visible `tip` is provided. I also added a no-tip test covering the default page loading path.

apps/aevatar-console-web/src/global.less
AbigailDeng: Fixed in 5bbb3b6c6. The `prefers-reduced-motion: reduce` rule now disables the shared loading animations instead of slowing them down, and pins opacity/transform so dots and cursors render statically.

PR #2698, opened 2026-07-10
github-devloop implementation for #2666

potter-sun:
github-devloop implementation PR for issue #2666


AbigailDeng (approved):
Reviewed the Team Test entry-selection change. No blocking findings.

The updated row action now separates the two intents correctly: when the team has no entry and the prompt is empty, it only sets the entry member; when a prompt is present, it still sets the entry and starts Team Test after the entry read model is visible. The main Start test button remains gated on an existing runnable entry member plus a non-empty prompt, and the implementation keeps using `row.memberId` for entry selection.

Verified locally:
- `./node_modules/.bin/tsc --noEmit`
- `./node_modules/.bin/jest --runInBand src/pages/teams/detail.test.tsx` (75 tests passed)

Note: invoking through `pnpm --dir apps/aevatar-console-web ...` in the temporary worktree still triggers the local pnpm ignored-builds policy before execution, so I ran the installed binaries directly after dependency materialization.

PR #2864, opened 2026-07-20
Integrate server-backed Chat History in Console

AbigailDeng:
## Problem

The Console still treated chat history as browser-owned state and used the legacy chat payload shape. That conflated `sessionId` with Chat History identity, prevented authoritative list/detail/delete behavior, and made newly committed turns disappear while the read model was catching up.

Related to #2834.

## Solution

- Add authenticated, runtime-decoded clients for scoped Chat History list, detail, and delete endpoints, and route those endpoints to the Studio host in local proxy configuration.
- Send the typed `/api/chat` intent exactly as `conversation: {}` for a new history and `conversation: { conversationId }` for continuation. Keep `sessionId` independent and stable, send only the current prompt, and omit legacy `chatHistory` and body `scopeId` fields.
- Accept Chat History identity only from the exact `aevatar.chat.context` SSE payload; fail completed streams without valid context and retry continuation-only `404 CONVERSATION_NOT_FOUND` responses for exactly three total attempts.
- Replace browser persistence and local rename/save controls with server-backed history, projection-delay overlays and reconciliation, delete tombstones, and stale scope/detail/delete protection.
- Bind History reads to the scope resolved by `/api/auth/me`, block route/auth scope splits, and require a trusted authenticated scope before starting or continuing `/api/chat`.
- Add a desktop history rail, mobile History Drawer, loading/empty/error/retry/delete states, open-role author attribution, latest-turn status recovery, and Studio Files rendering for author, role, thinking, status, and assistant errors.

## Impacted paths

- `apps/aevatar-console-web/src/pages/chat/**`: request contract, SSE identity, server-backed history state/UI, scope trust boundaries, and regression tests.
- `apps/aevatar-console-web/src/pages/studio/components/StudioFiles*`: Chat History tree, detail semantics, deletion operation isolation, and tests.
- `apps/aevatar-console-web/config/proxy.ts` and `src/shared/config/proxyConfig.test.ts`: scoped Chat History Studio-host proxy routing.
- `apps/aevatar-console-web/src/global.less` and locale catalogs: responsive layout and English/Chinese copy.

## Verification

- Frozen install with the declared `pnpm@10.2.1`: passed.
- Changed-file Biome lint: passed. Full-tree Biome remains blocked by 81 pre-existing `origin/dev` diagnostics outside this change.
- `pnpm --dir apps/aevatar-console-web tsc`: passed.
- `pnpm --dir apps/aevatar-console-web test --runInBand`: 129/129 suites and 1106/1106 tests passed on final HEAD.
- `pnpm --dir apps/aevatar-console-web build`: passed on final HEAD.
- `git diff --check`, frontend-only path boundary, and credential-pattern scan: passed.
- FKST pre-commit and pre-push hooks: passed without `--no-verify`; pre-push also passed docs lint for 77 files with 0 errors.
- Independent final review: passed with no remaining P1/P2 findings.

Local Direct Playwright QA passed 10/10 journeys at `1440x900`, `1024x768`, and `390x844`: desktop list/detail, list retry, detail retry with stable selection, streaming/Stop/error, projection-delay live history, three-attempt continuation retry, delete cancel/loading/success/failure, tablet layout, mobile drawer, and Studio Files message semantics.

Post-review Direct Playwright QA passed 3/3 additional journeys: route/auth scope mismatch blocked both read and write requests, generic roles preserved author attribution, and delayed Studio deletion remained bound to the original conversation.

This is local browser evidence, not a GitHub-hosted E2E check.

- Full report: `/Users/xiezixin/.fkst/artifacts/aevatar-frontend/20260720T040404Z-c7db0aec/logs/browser-qa-playwright.json`
- Review-fix report: `/Users/xiezixin/.fkst/artifacts/aevatar-frontend/20260720T040404Z-c7db0aec/logs/browser-qa-review-fixes.json`
- Real-backend report: `/Users/xiezixin/.fkst/artifacts/aevatar-frontend/20260720T040404Z-c7db0aec/logs/browser-qa-real-backend.json`
- Screenshots: `/Users/xiezixin/.fkst/artifacts/aevatar-frontend/20260720T040404Z-c7db0aec/screenshots/`

## Known backend limitations

- In the temporary `origin/feature/integrate@dddca3b` backend environment, the scoped list remained empty while the same server-owned conversation detail returned committed turns. Populated list and delete journeys therefore used a deterministic contract fixture; the real backend was used for auth, stream context, pending state, continuation identity, and detail verification.
- The temporary backend could not authenticate to its upstream LLM route, so the real stream ended in the expected persisted error-turn path rather than a successful model answer.
- `conversationId` currently binds Chat History identity but does not rehydrate prior turns into Workflow/LLM execution. The frontend intentionally sends only the raw current prompt and does not reconstruct a transcript.

## FKST provenance

- Issue: #2834
- Worktree: /Users/xiezixin/.fkst/runtime/aevatar-frontend/manual/feat-2026-07-20_chat-history-api-integration
- Branch: feat/2026-07-20_chat-history-api-integration
- Operator: AbigailDeng
- Freshness: frontend migration for origin/feature/integrate@dddca3b, rebased on origin/dev@dc641555f
- Workflow run: 20260720T040404Z-c7db0aec


apps/aevatar-console-web/config/proxy.ts
```diff
@@ -58,6 +58,8 @@ const studioProxyEntries = [
 }, {});
 
 const studioScopeProxyEntries = {
+  '^/api/scopes/[^/]+/chat-history(?:/.*)?$':
```
potter-sun, line 61: [P1] This route exposes a cross-scope IDOR in the current backend. ChatHistoryActorIds.Conversation builds `chat-{scopeId}-{conversationId}` without unambiguous encoding, so `(tenant, admin-c1)` and `(tenant-admin, c1)` resolve to the same actor. The scope guard validates only the path scope, and GetMessagesAsync/DeleteConversationAsync do not verify the loaded actors stored scope/conversation identity. Please make the tuple collision-free and enforce ownership on reads/deletes before routing this surface; add a regression test for the two colliding tuples.

apps/aevatar-console-web/src/pages/chat/index.tsx
```diff
@@ -625,41 +1013,75 @@ const ChatPage: React.FC = () => {
...
+          }
           rawFrames.push(frame.raw);
+          const chatHistoryContext = extractChatHistoryContext(frame.raw);
+          if (chatHistoryContext) {
+            receivedChatHistoryContext = true;
+            if (chatHistoryContext.scopeId !== scopeId) {
+              throw new Error("Chat History context does not match the active scope.");
+            }
+            if (
+              conversation.conversationId &&
```
potter-sun, line 1167: [P1] For a create flow, `conversation.conversationId` remains undefined, so every later valid-looking context frame passes this guard and overwrites the accepted conversation/turn identity. A duplicated or replayed conflicting frame can therefore bind subsequent turns to the wrong conversation. Capture the first accepted context and reject any later context whose scope, conversationId, or turnId differs; cover both wrong-scope and wrong-conversation frames.

apps/aevatar-console-web/src/pages/chat/index.tsx
```diff
@@ -438,160 +729,254 @@ const ChatPage: React.FC = () => {
...
+              const next = {
+                ...current,
+                expectedTurnCount: serverMeta.messageCount,
+                title: serverMeta.title || current.title,
+              };
+              activeConversationRef.current = next;
+              return next;
+            });
+            return;
+          } catch {
```
potter-sun: [P1] All four reconciliation failures, and exhaustion without an exception, are silent. The optimistic entry then remains in memory indefinitely while the UI says history is saved; after navigation/reload it can disappear with no retry or warning. Model reconciliation as pending/failed and keep a background retry or explicit retry action until the authoritative turn/version is observed.

apps/aevatar-console-web/src/pages/chat/chatHistoryApi.ts
```diff
@@ -1,44 +1,268 @@
...
+  const record = asRecord(value, path);
+  const message = {
+    content: readString(record, "content", path),
+    id: readString(record, "id", path),
+    role: readString(record, "role", path),
+    status: readString(record, "status", path),
+    timestamp: readNumber(record, "timestamp", path),
+  } as StoredChatMessage & Record<string, unknown>;
+
+  for (const key of ["error", "thinking", "authorId", "authorName"] as const) {
```
potter-sun: [P2] The backend response already carries typed `turnId`, but this decoder drops it and reconciliation later derives identity from the display ID convention `${turnId}:assistant`. That violates the typed contract and will strand conversations in pending state if message ID formatting changes. Preserve `turnId` on StoredChatMessage and match `turnId` plus role instead.

apps/aevatar-console-web/src/pages/chat/chatHistoryApi.ts
```diff
@@ -1,44 +1,268 @@
...
+    payload = await response.json();
+  } catch {
+    throw new ChatHistoryContractError("$response", "valid JSON");
+  }
+
+  return decoder(payload);
+}
 
 export const chatHistoryApi = {
   async listConversationMetas(scopeId: string): Promise<ConversationMeta[]> {
```
potter-sun, line 273: [P1] This list contract has no pagination, while ActorBackedChatHistoryStore hard-caps the query with `Take = ChatConversationGAgent.MaxTurns` (250). Once a workspace has 251 conversations, older histories become unreachable from both Chat and Studio and cannot be opened or deleted. Add a cursor/continuation contract and a load-more path rather than reusing the per-conversation turn limit as a workspace list limit.

apps/aevatar-console-web/src/pages/chat/index.tsx
```diff
@@ -760,27 +1209,32 @@ const ChatPage: React.FC = () => {
           ),
           status: "error",
         };
-        setActiveConversation(failedConversation);
-        await persistConversation(failedConversation);
+        if (receivedChatHistoryContext && failedConversation.conversationId) {
```
potter-sun: [P1] If the command is accepted but the SSE connection drops before the context frame reaches the browser, this error path has no conversation or command identity to recover. Retrying the same draft sends another create intent with a new server command ID and can execute side effects twice. The create contract needs a stable client idempotency key plus recovery/read-model lookup before this can be treated as retryable chat.

apps/aevatar-console-web/src/pages/chat/index.tsx
```diff
@@ -438,160 +729,254 @@ const ChatPage: React.FC = () => {
...
+      const controller = new AbortController();
+      reconciliationControllersRef.current.set(conversationId, controller);
+      const reconciliationScopeEpoch = scopeEpochRef.current;
+      const delaysMs = [0, 300, 900, 1_800];
+
+      void (async () => {
+        for (const delayMs of delaysMs) {
+          try {
+            await abortableDelay(delayMs, controller.signal);
+            const [nextConversations, storedMessages] = await Promise.all([
```
potter-sun: [P2] Every reconciliation attempt downloads both the full 250-item index and the full transcript, for up to eight growing payloads per completed turn. The metadata count usually proves observation by itself. Fetch the index first and load detail only when the count cannot prove the expected turn, or expose a lightweight latest-turn/version read model.

potter-sun (changes requested):
Requesting changes because the new server-backed history surface exposes an actor-ID ownership collision and has several durability/identity gaps. The active design doc is also stale: docs/design/2026-07-16-studio-chat-history-terminal-append.md:80 still promises local browser fallback, and its create example sends conversationId:null while this client intentionally sends conversation:{}.

apps/aevatar-console-web/src/pages/chat/index.tsx
AbigailDeng: Fixed in `798e4ffa1`. The stream handler now locks the first accepted `aevatar.chat.context` and rejects any later frame whose scope, conversation ID, or turn ID differs. Added regression coverage for wrong-scope and wrong-conversation frames. Focused Chat suites pass 39 tests; full frontend Jest passes 129 suites / 1109 tests.

apps/aevatar-console-web/src/pages/chat/index.tsx
AbigailDeng: Fixed in `798e4ffa1`. Reconciliation now has explicit pending/failed state, keeps the optimistic conversation visible, stops claiming history is saved while unresolved, and exposes retry after exhaustion. Tests cover visible exhaustion and successful explicit retry. Focused Chat suites pass 39 tests; full frontend Jest passes 129 suites / 1109 tests.

apps/aevatar-console-web/src/pages/chat/chatHistoryApi.ts
AbigailDeng: Fixed in `798e4ffa1`. The decoder preserves the backend typed `turnId`, and reconciliation matches assistant role plus typed turn ID instead of deriving identity from the display message ID. Added coverage with an opaque message ID. Focused Chat suites pass 39 tests; full frontend Jest passes 129 suites / 1109 tests.

apps/aevatar-console-web/src/pages/chat/index.tsx
AbigailDeng: Fixed in `798e4ffa1`. Reconciliation now fetches index metadata first and uses `messageCount` to prove observation; full conversation detail is loaded only when metadata cannot prove the expected turn. Tests assert that the normal metadata-proven path does not fetch detail. Focused Chat suites pass 39 tests; full frontend Jest passes 129 suites / 1109 tests.

apps/aevatar-console-web/config/proxy.ts
AbigailDeng: Tracked in #2888 and assigned to @jason-aelf. This requires backend work on `feature/integrate` for collision-free actor identity and stored ownership checks, so it is intentionally outside the frontend-only commit `798e4ffa1`.

apps/aevatar-console-web/src/pages/chat/chatHistoryApi.ts
AbigailDeng: Tracked in #2888 and assigned to @jason-aelf. The issue requires a typed cursor/continuation contract and deterministic backend paging on `feature/integrate`; the frontend load-more path will depend on that contract, so this is intentionally outside `798e4ffa1`.

apps/aevatar-console-web/src/pages/chat/index.tsx
AbigailDeng: Tracked in #2888 and assigned to @jason-aelf. Stable create idempotency and scope-bound recovery require typed backend contracts and a materialized recovery read model on `feature/integrate`, so this is intentionally outside the frontend-only commit `798e4ffa1`.

potter-sun (approved):

PR #2930, opened 2026-07-23
fix(console): align chat history continuation contract

AbigailDeng:
## Problem

Fixes #2920.

Restored chat conversations still used the previous continuation contract. The frontend replayed history in the prompt, had no authoritative server watermark, and could either retry with stale state or leave the conversation in the wrong interaction state after an ambiguous request failure.

Backend contract dependency: #2923 defines the required conversation-detail and SSE watermark contract. It was merged into `auto-work-dev-0701`, not `dev`, so integration and deployment of this PR require that backend contract to be present in the target environment.

## Solution

- Decode conversation detail as `{ messages, stateVersion }` and keep the server transcript authoritative.
- Send `{ conversationId: null }` for a new conversation and `{ conversationId, minimumStateVersion }` for a continuation.
- Send only the current prompt; remove client-side `<conversation_history>` replay.
- Read typed `stateVersion` from `aevatar.chat.context` and retain watermarks monotonically.
- Retry only typed `503 CHAT_HISTORY_RESERVATION_UNAVAILABLE` responses after refreshing authoritative history.
- Distinguish definitive rejection, retry-preparation failure, and potentially accepted POST failure.
- Recover accepted creates when the stream disconnects before context, and fail closed for ambiguous continuations without typed context.
- Keep composer and confirmation actions disabled while reconciliation is in progress.
- Update Studio Files detail loading for the new conversation-detail response shape.

## Impact paths

- Console chat request and SSE handling
- Conversation history loading and reconciliation
- Restored-conversation composer state
- Studio Files conversation detail view
- English and Chinese error copy

## Verification

- Focused affected suites: 4/4 suites, 77/77 tests passed.
- TypeScript: passed.
- Biome lint for all changed frontend files: passed.
- `git diff --check`: passed.
- Frontend production build: passed.
- A prior local full Jest run completed 129/130 suites and 1147/1148 tests. The only failure was an unrelated timeout in `src/pages/runs/index.test.tsx:762`; the selected test passed in isolation in 471 ms. Full-suite validation is left to PR CI per repository workflow.

## Recurrence prevention

- Root cause: the client lacked a typed authoritative watermark and inferred request acceptance from broad error classes.
- General failure pattern: eventually consistent retries mixed authoritative state, transient preparation failures, and potentially accepted POST failures.
- Guard added: typed detail decoding, positive continuation validation, pre-dispatch abort classification, monotonic watermark retention, phase-aware retry errors, authoritative refresh merging, create recovery, and an ambiguous-continuation lock.
- Regression coverage: new and continuation payloads, `0 -> 6 -> 8` refresh, transient and terminal refresh failures, refresh-to-dispatch cancellation, stale context rollback, refreshed transcript retention after stream failure, create recovery after disconnect, unreadable error bodies, cross-tab turns, abort propagation, ambiguous no-context handling, and watermark reuse on the next turn.

### DIFF
--- a/apps/aevatar-console-web/src/pages/chat/chatApi.ts
+++ b/apps/aevatar-console-web/src/pages/chat/chatApi.ts
@@ -7,6 +7,10 @@ import type {
   ChatStudioTarget,
   ChatUsageSummary,
 } from "./chatTypes";
+import {
+  ChatHistoryApiError,
+  ChatHistoryContractError,
+} from "./chatHistoryApi";
 
 type JsonRecord = Record<string, unknown>;
 
@@ -45,6 +49,23 @@ export class ChatApiError extends Error {
     this.code = code;
     this.status = status;
   }
+}
+
+export class ChatRetryPreparationError extends Error {
+  readonly cause: unknown;
+  readonly mayHaveBeenAccepted = false;
+
+  constructor(cause: unknown) {
+    super(cause instanceof Error ? cause.message : String(cause));
+    this.name = "ChatRetryPreparationError";
+    this.cause = cause;
+  }
+}
+
+export function chatRequestMayHaveBeenAccepted(error: unknown): boolean {
+  return !(
+    error instanceof ChatApiError || error instanceof ChatRetryPreparationError
+  );
 }
 
 function compactObject<T extends Record<string, unknown>>(value: T): T {
@@ -376,7 +397,18 @@ export async function startChatStream(
   });
 
   if (!response.ok) {
-    const details = await readResponseErrorDetails(response);
+    let details: Awaited<ReturnType<typeof readResponseErrorDetails>>;
+    try {
+      details = await readResponseErrorDetails(response);
+    } catch {
+      const statusText = response.statusText.trim();
+      throw new ChatApiError(
+        statusText
+          ? `HTTP ${response.status} ${statusText}`
+          : `HTTP ${response.status}`,
+        response.status
+      );
+    }
     throw new ChatApiError(details.message, details.status, details.code);
   }
 
@@ -404,8 +436,42 @@ function abortableDelay(delayMs: number, signal: AbortSignal): Promise<void> {
   });
 }
 
+function isAbortError(error: unknown, signal: AbortSignal): boolean {
+  return (
+    signal.aborted ||
+    (error instanceof DOMException && error.name === "AbortError")
+  );
+}
+
+function throwIfAbortedBeforeDispatch(signal: AbortSignal): void {
+  if (!signal.aborted) {
+    return;
+  }
+
+  throw new ChatRetryPreparationError(
+    signal.reason ?? new DOMException("Aborted", "AbortError")
+  );
+}
+
+function isRetryableHistoryRefreshError(error: unknown): boolean {
+  if (error instanceof ChatHistoryContractError) {
+    return false;
+  }
+
+  if (error instanceof ChatHistoryApiError) {
+    return error.status >= 500 && error.status < 600;
+  }
+
+  if (error && typeof error === "object" && "status" in error) {
+    const status = (error as { status?: unknown }).status;
+    return typeof status === "number" && status >= 500 && status < 600;
+  }
+
+  return true;
+}
+
 export type ChatHistoryRefreshRetryOptions = {
-  refreshMinimumStateVersion?: () => Promise<number>;
+  refreshMinimumStateVersion?: (signal: AbortSignal) => Promise<number>;
   retryDelaysMs?: readonly number[];
 };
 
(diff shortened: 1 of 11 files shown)


apps/aevatar-console-web/src/pages/chat/chatHistoryApi.ts
```diff
@@ -204,16 +205,40 @@ export function decodeChatCreateRecovery(value: unknown): ChatCreateRecovery {
...
   return value.map((message, index) =>
-    decodeStoredChatMessage(message, `$messages[${index}]`)
+    decodeStoredChatMessage(message, `${path}[${index}]`)
   );
 }
 
+export function decodeChatConversationDetail(
+  value: unknown
+): ChatConversationDetail {
+  const record = asRecord(value, "$conversation");
```
potter-sun, line 224: [P1] Ship the conversation-detail contract with this decoder

The target `dev` controller still returns `IReadOnlyList<StoredChatMessage>` as a JSON array, while this code now requires `{ messages, stateVersion }`. Every history restore, Studio Files detail load, and continuation refresh will throw `ChatHistoryContractError`. Please include or merge #2923 into the target first, retarget this PR, or keep the array decoder until the backend contract ships.

apps/aevatar-console-web/src/pages/chat/chatApi.ts
```diff
@@ -231,10 +257,17 @@ export function extractChatHistoryContext(
   const context: ChatHistoryContext = {
     conversationId: readString(payload, "conversationId"),
     scopeId: readString(payload, "scopeId"),
+    stateVersion: readNumber(payload, "stateVersion") ?? Number.NaN,
```
potter-sun, line 260: [P1] Do not require a watermark that the target backend never emits

On the target `dev`, `WorkflowChatContextPayload` contains only `scope_id`, `conversation_id`, and `turn_id`, and `BuildChatContextFrame` populates only those fields. `stateVersion` is therefore always absent, so this extractor returns `null` for every valid context and successful streams end as missing-context failures. The typed backend field and authoritative value need to ship in the same target before this requirement is enabled.

apps/aevatar-console-web/src/pages/chat/chatApi.ts
```diff
@@ -301,19 +334,44 @@ function normalizeConversationIntent(
...
+    minimumStateVersion <= 0
+  ) {
+    throw new ChatApiError(
+      "Conversation state version is invalid.",
+      400,
+      "INVALID_CONVERSATION_STATE_VERSION"
+    );
+  }
+
+  return { conversationId, minimumStateVersion };
```
potter-sun, line 374: [P1] The target backend rejects every continuation payload

This newly serialized `minimumStateVersion` is not a field on the target branch `ChatConversationInput`, which uses `JsonUnmappedMemberHandling.Disallow`. Consequently every continuation POST is rejected as `INVALID_CONVERSATION_INPUT` before dispatch. Please land the end-to-end typed backend contract from #2923 first, or include it here, rather than merging this client change onto `dev` alone.

apps/aevatar-console-web/src/pages/chat/chatApi.ts
```diff
@@ -42,6 +51,23 @@ export class ChatApiError extends Error {
...
+  readonly mayHaveBeenAccepted = false;
+
+  constructor(cause: unknown) {
+    super(cause instanceof Error ? cause.message : String(cause));
+    this.name = "ChatRetryPreparationError";
+    this.cause = cause;
+  }
+}
+
+export function chatRequestMayHaveBeenAccepted(error: unknown): boolean {
```
potter-sun, line 65: [P1] Treat post-accept HTTP 5xx as ambiguous

`ChatApiError` includes every HTTP response, but an HTTP response is not proof that dispatch was rejected. The server marks the command accepted before awaiting `BindAcceptedAsync`; if that bind throws, the SSE writer has not started and the endpoint returns JSON 500. This function then reports not accepted, unlocks the continuation, and a retry has no stable command ID, so it can execute duplicate side effects. Treat 5xx as ambiguous and reconcile, or emit a durable accepted receipt before fallible binding and use a stable continuation command ID.

apps/aevatar-console-web/src/pages/chat/index.tsx
```diff
@@ -950,7 +1033,10 @@ const ChatPage: React.FC = () => {
           try {
             await abortableDelay(delayMs, controller.signal);
             const nextConversations =
-              await chatHistoryApi.listConversationMetas(scopeId);
+              await chatHistoryApi.listConversationMetas(
```
potter-sun, line 1036: [P2] Avoid scanning the complete history on every reconciliation attempt

`listConversationMetas` follows every `nextCursor` before resolving. Calling it on each 0/300/900/1800 ms reconciliation attempt makes a completed turn issue up to `4 * pageCount` list requests and retain the full history, so cost grows with account age. The turn can be observed through the targeted conversation detail; refresh or invalidate the index separately, or add a targeted metadata lookup.

potter-sun (commented):
Review summary: blocking against the current `dev` target. This frontend-only PR assumes the #2923 backend contract, but that contract is not present on the base branch. The affected frontend tests mock the new response and frame shapes, so green CI does not exercise the cross-layer failures called out inline.

[P2] The exhausted-reconciliation test at `src/pages/chat/index.test.tsx:1664` waits on the real 0/300/900/1800 ms retry schedule with a 4.5-second ceiling. The focused chat suite took 38.7 seconds locally. Please use Jest fake timers and advance the retry schedule deterministically.

PR #2943, opened 2026-07-23
fix: show all workflow nodes in execution overview

AbigailDeng:
## Problem

The Workflow Studio execution Overview was derived only from runtime logs. At run start, untouched workflow nodes were missing and the UI emphasized one oversized first-node detail instead of presenting the submitted graph. Row height was also content-dependent, so the Overview could shift as statuses and long labels arrived.

## Solution

- Capture a run-scoped snapshot of every submitted workflow node before YAML serialization and keep it independent from subsequent canvas edits.
- Reconcile submitted nodes with runtime attempts: definition nodes appear immediately as Pending, matching runtime attempts replace them, and runtime-only nodes append after the definition order.
- Keep distinct retries for the same step selectable through stable per-attempt log-index membership.
- Render every Overview row at a fixed 80px height with clipped, ellipsized content.
- Skip disabled Pending rows during keyboard navigation and disable Clear logs while a draft run is still streaming.
- Stack Overview and details below 720px without changing the desktop information architecture.
- Add localized Pending labels for English and Chinese.

## Impacted paths

- `apps/aevatar-console-web/src/pages/team-member-workflow-studio/components/WorkflowStudioExecutionPanel.tsx`
- `apps/aevatar-console-web/src/pages/team-member-workflow-studio/hooks/useTeamMemberWorkflowStudio.ts`
- `apps/aevatar-console-web/src/pages/team-member-workflow-studio/index.tsx`
- `apps/aevatar-console-web/src/pages/team-member-workflow-studio/index.test.tsx`
- `apps/aevatar-console-web/src/locales/en-US.ts`
- `apps/aevatar-console-web/src/locales/zh-CN.ts`

## Verification

- Touched-file Biome lint for all six changed files: passed, 6 files checked with no fixes.
- `pnpm --dir apps/aevatar-console-web tsc`: passed on the latest `dev` base.
- Focused Workflow Studio and i18n suites: 97/97 passed.
- `pnpm --dir apps/aevatar-console-web test --runInBand --no-watchman`: 130/130 suites and 1148/1148 tests passed.
- After fast-forwarding the latest unrelated Chat commit from `dev`, Chat and i18n tests passed. A later combined rerun hit two existing Workflow Studio 30-second timeouts under host contention; both unrelated cases passed immediately in isolated reruns, 2/2.
- `pnpm --dir apps/aevatar-console-web build`: passed on the latest `dev` base.
- `git diff --check`: passed.

The full-repository Biome scan has a pre-existing baseline of 89 errors, 82 warnings, and 9 infos outside these touched files. This PR does not change that baseline.

## Coverage

```text
CODE PATH COVERAGE
===========================
[+] Definition reconciliation
    [+] All submitted nodes render Pending at RUN_STARTED
    [+] Runtime entries replace matching Pending nodes
    [+] Runtime-only nodes append after definition nodes
    [+] Repeated step attempts remain separate

[+] Run-scoped lifecycle
    [+] Live canvas edits do not mutate active-run Overview
    [+] Failed runs retain submitted-node context
    [+] New runs clear previous logs before serialization completes
    [+] Clear logs removes the completed execution UI and snapshot
    [+] Clear logs is disabled while a run is streaming

[+] Interaction and layout
    [+] Pending rows are disabled and skipped by keyboard navigation
    [+] Retry selection survives completion frames
    [+] Rows stay at 80px and hide overflowing content
    [+] Mobile breakpoint stacks Overview and details

COVERAGE: 13/13 paths tested (100%)
Coverage gate: PASS
```

## Review results

- Structured review found retry selection could jump to the first attempt. Fixed with per-attempt `logIndexes` membership and a completion-frame regression test.
- Run-lifecycle review found a new snapshot could temporarily pair with previous-run logs. Fixed by resetting execution detail, error, and selection in `onMutate`.
- Adversarial review found live Clear logs could be repopulated by the next SSE frame. Fixed by disabling Clear logs while the mutation is pending.
- Adversarial review found the original 76px geometry could clip Ant Design tags. Fixed by using a measured fixed height of 80px.
- A suggestion to label terminal unvisited branches as Skipped was intentionally not implemented. The backend currently provides no authoritative skipped-node state; step-level `nextStepId` or `branchKey` is not sufficient to infer all terminal node outcomes.

## Manual browser QA

| Viewport | Result |
| --- | --- |
| 1440x900 | All nodes visible; fixed rows; side-by-side details; no overflow |
| 1024x768 | All nodes visible; fixed rows; no overlap |
| 768x1024 | Responsive layout stable; no document overflow |
| 390x844 | Overview stacks above details; no overlap or console errors |

Browser QA observed four nodes, fixed row geometry, no horizontal document overflow, correct mobile stacking, and zero console errors.

## Recurrence prevention

- Immediate symptom: Overview initially omitted untouched nodes and emphasized one large node detail.
- Root cause: Overview was derived only from runtime logs and read the mutable editor graph.
- General failure pattern: execution-scoped UI state was derived from incomplete events and live editor state.
- Similar locations searched: WorkflowStudioExecutionPanel callers, execution selection helpers, and run mutation lifecycle; no sibling implementation found.
- Guard added: run-scoped node snapshot plus unique per-attempt log membership.
- Regression test added: Pending nodes, live graph edits, retry selection, failure context, rerun isolation, live-clear protection, fixed rows, and responsive CSS.
- Hard gate: touched-file Biome, TypeScript, Jest, production build, i18n audit, and test-stability guard.
- Remaining risk: no backend-authored Skipped state for terminal unvisited branches.

## Documentation

No public API, route, configuration, or operator workflow changed. Documentation review is expected to require no repository documentation update.

### DIFF
--- a/apps/aevatar-console-web/src/pages/team-member-workflow-studio/hooks/useTeamMemberWorkflowStudio.ts
+++ b/apps/aevatar-console-web/src/pages/team-member-workflow-studio/hooks/useTeamMemberWorkflowStudio.ts
@@ -33,17 +33,18 @@ import {
 } from "@/shared/studio/document";
 import {
   buildExecutionTrace,
+  buildWorkflowExecutionNodeSnapshots,
   createStudioExecutionFrame,
   decorateEdgesForExecution,
   decorateNodesForExecution,
   findExecutionLogIndexForStep,
   type ExecutionTrace,
+  type WorkflowExecutionNodeSnapshot,
 } from "@/shared/studio/execution";
 import {
   buildStudioGraphElements,
   buildStudioWorkflowLayout,
   STUDIO_GRAPH_CATEGORIES,
-  type StudioGraphNodeData,
 } from "@/shared/studio/graph";
 import { isStudioApiStatus, studioApi } from "@/shared/studio/api";
 import {
@@ -67,10 +68,6 @@ import type {
 type TeamMemberWorkflowStudioMode = "new" | "existing";
 type WorkflowPublishTone = "default" | "processing" | "success" | "warning" | "error";
 type WorkflowExecutionStatus = "idle" | "running" | "succeeded" | "failed";
-type WorkflowExecutionNode = Pick<
-  StudioGraphNodeData,
-  "stepId" | "stepType" | "subtitle" | "targetRole" | "title"
->;
 
 type SaveWorkflowDraftVariables = {
   readonly document: StudioWorkflowDocument;
@@ -215,7 +212,7 @@ type TeamMemberWorkflowStudioState = {
   readonly currentDraftRunPlaceholderReason: string;
   readonly executionDetail: StudioExecutionDetail | null;
   readonly executionError: string;
-  readonly executionWorkflowNodes: readonly WorkflowExecutionNode[];
+  readonly executionWorkflowNodes: readonly WorkflowExecutionNodeSnapshot[];
   readonly draftRunFiles: readonly File[];
   readonly executionRunMessage: string;
   readonly executionStatus: WorkflowExecutionStatus;
@@ -1098,7 +1095,7 @@ export function useTeamMemberWorkflowStudio(): TeamMemberWorkflowStudioState {
     React.useState<StudioExecutionDetail | null>(null);
   const [executionError, setExecutionError] = React.useState("");
   const [executionWorkflowNodes, setExecutionWorkflowNodes] = React.useState<
-    WorkflowExecutionNode[]
+    WorkflowExecutionNodeSnapshot[]
   >([]);
   const [activeExecutionLogIndex, setActiveExecutionLogIndex] =
     React.useState<number | null>(null);
@@ -2353,9 +2350,7 @@ export function useTeamMemberWorkflowStudio(): TeamMemberWorkflowStudioState {
       setExecutionDetail(null);
       setExecutionError("");
       setActiveExecutionLogIndex(null);
-      setExecutionWorkflowNodes(
-        buildStudioGraphElements(document).nodes.map((node) => node.data),
-      );
+      setExecutionWorkflowNodes(buildWorkflowExecutionNodeSnapshots(document));
     },
     onSuccess: (detail) => {
       setExecutionDetail(detail);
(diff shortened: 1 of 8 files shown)


apps/aevatar-console-web/src/pages/team-member-workflow-studio/hooks/useTeamMemberWorkflowStudio.ts
```diff
@@ -2340,8 +2349,13 @@ export function useTeamMemberWorkflowStudio(): TeamMemberWorkflowStudioState {
           : "Failed to run workflow draft.",
       );
     },
-    onMutate: () => {
+    onMutate: ({ document }: RunCurrentDraftVariables) => {
+      setExecutionDetail(null);
       setExecutionError("");
+      setActiveExecutionLogIndex(null);
+      setExecutionWorkflowNodes(
+        buildStudioGraphElements(document).nodes.map((node) => node.data),
```
potter-sun: [P2] Build the execution snapshot from an execution-specific contract

This run-scoped snapshot is currently derived through buildStudioGraphElements, which returns React Flow view data (Node<StudioGraphNodeData>). The panel and hook also duplicate the same Pick contract. That makes execution history depend on presentation mapping, so a later graph filtering/labeling change can silently alter which submitted nodes the Overview records. Please define one shared WorkflowExecutionNodeSnapshot plus a document-to-snapshot mapper in the execution/model module, and use it here and in the panel.

AbigailDeng: Fixed in c15f919d4.

- Added the shared `WorkflowExecutionNodeSnapshot` contract and `buildWorkflowExecutionNodeSnapshots` document mapper in `shared/studio/execution`.
- The run hook now captures the submitted workflow document directly instead of deriving execution history from React Flow view nodes.
- The execution panel consumes the same shared snapshot contract, removing the duplicated `Pick<StudioGraphNodeData, ...>` dependency.
- Added focused mapper coverage for document order, compatibility-field normalization, and filtering steps without an id.

Verified with the focused execution-model test, the Workflow Studio SSE overview test, frontend TypeScript checking, the test-stability guard, and the production frontend build.

PR #2948, opened 2026-07-24
fix(console): scope team automation schedule queries

AbigailDeng:
## Problem

Team Automation pages queried the generic `/api/schedules` collection and filtered results only in the client. After backend PR #2937 added scoped schedule reads, the console still did not pass the authoritative scope/team/member query context.

Closes #2942.

## Solution

- Extend `ScheduledDispatchListQuery` with `scopeId`, optional `teamId`, and optional `memberId`, and serialize those values into `/api/schedules`.
- Preserve the scoped filters through every `listAll` cursor page while leaving unscoped callers unchanged.
- Pass route-derived `scopeId` and `teamId` from Team Automation pages, and pass `routeMemberId` only on member-scoped routes.
- Include the route member identity in the React Query key so team-level and member-level results cannot share a cache entry.
- Keep schedule create/update authorization and owner payloads unchanged.
- Use a semantic `article` for each labeled automation row so its existing accessible name is valid.

## Impact paths

- `apps/aevatar-console-web/src/shared/api/scheduledDispatchApi.ts`
- `apps/aevatar-console-web/src/pages/teams/detail.tsx`
- `apps/aevatar-console-web/src/pages/teams/tabs/TeamAutomationsTab.tsx`
- Focused API and Team detail tests

No backend, root configuration, dependency, or documentation changes are required.

## Validation

- [x] `pnpm --dir apps/aevatar-console-web jest --runInBand src/shared/api/scheduledDispatchApi.test.ts src/pages/teams/detail.test.tsx` - 94 passed
- [x] `pnpm --dir apps/aevatar-console-web jest --runInBand src/shared/api/scheduledDispatchApi.test.ts` - 17 passed on final source
- [x] Focused Team Automation page cases - 2 passed on final source
- [x] `pnpm --dir apps/aevatar-console-web tsc`
- [x] `pnpm --dir apps/aevatar-console-web build`
- [x] Targeted Biome lint for all 5 changed files - exit 0; 2 existing warnings and 2 existing infos in `detail.tsx`
- [ ] Full frontend Jest suite was not run, per the machine-local Incremental Unit Test Policy.
- [ ] Full frontend Biome lint is blocked by 89 pre-existing `origin/dev` errors outside this diff.

## Browser QA

The debug app started at `http://localhost:5174`, but `/scopes` redirected to `/login?redirect=%2Fscopes`. No authenticated Console session was available, so the Team Automation user journey was not substituted with jsdom evidence or reported as browser E2E.

- Console errors on the login surface: none
- Blocker screenshot: `/Users/xiezixin/.fkst/artifacts/aevatar-frontend/20260724T022022Z-97ab53bc/screenshots/browser-qa-auth-blocker.png`

## FKST provenance

- Issue: #2942
- Worktree: `/Users/xiezixin/.fkst/runtime/aevatar-frontend/manual/fix-2026-07-24_scoped-schedule-queries`
- Branch: `fix/2026-07-24_scoped-schedule-queries`
- Operator: AbigailDeng
- Freshness: `origin/dev@af89911ebe844045b54d80af4852050037b3a9a1`, fetched 2026-07-24 before commit
- Workflow run: `20260724T022022Z-97ab53bc`

## Backend dependency

This frontend contract depends on backend PR #2937 (`Support scoped schedule queries`), merged as `51d6e2fef` into `feature/integrate`. That merge is not yet an ancestor of `origin/dev` as of 2026-07-24. The frontend branch intentionally remains based on `dev` per the requested workflow; deploy/release this change with #2937 so `/api/schedules?scopeId=...&teamId=...&memberId=...` is handled by the scoped backend query path. The parameter names and generic fallback behavior were verified against #2937 endpoint tests.

### DIFF
--- a/apps/aevatar-console-web/src/pages/teams/detail.tsx
+++ b/apps/aevatar-console-web/src/pages/teams/detail.tsx
@@ -1977,6 +1977,11 @@ const TeamDetailPage: React.FC = () => {
   const renderAutomationsTab = () => {
     return (
       <TeamAutomationsTab
+        key={JSON.stringify([
+          scopeId,
+          selectedTeamId,
+          routeState.routeMemberId,
+        ])}
         members={teamRosterRows.map((row) => ({
           automationsHref: row.automationsHref,
           canAutomateMember: row.canAutomateMember,
@@ -1993,7 +1998,7 @@ const TeamDetailPage: React.FC = () => {
           serviceRevisionId: row.serviceRevisionId,
           workflowSupported: row.workflowSupported,
         }))}
-        routeMemberId={routeState.memberId}
+        routeMemberId={routeState.routeMemberId}
         scopeId={scopeId}
         serviceIdentitiesLoading={automationsServicesQuery.isLoading}
         teamId={selectedTeamId}
--- a/apps/aevatar-console-web/src/pages/teams/detail.tsx
+++ b/apps/aevatar-console-web/src/pages/teams/detail.tsx
@@ -1993,6 +1993,7 @@ const TeamDetailPage: React.FC = () => {
           serviceRevisionId: row.serviceRevisionId,
           workflowSupported: row.workflowSupported,
         }))}
+        routeMemberId={routeState.memberId}
         scopeId={scopeId}
         serviceIdentitiesLoading={automationsServicesQuery.isLoading}
         teamId={selectedTeamId}
--- a/apps/aevatar-console-web/src/pages/teams/tabs/TeamAutomationsTab.tsx
+++ b/apps/aevatar-console-web/src/pages/teams/tabs/TeamAutomationsTab.tsx
@@ -71,6 +71,7 @@ export type TeamAutomationMemberRow = {
 
 type TeamAutomationsTabProps = {
   readonly members?: readonly TeamAutomationMemberRow[];
+  readonly routeMemberId?: string;
   readonly scopeId: string;
   readonly serviceIdentitiesLoading?: boolean;
   readonly teamId: string;
@@ -680,6 +681,7 @@ function hasBackendObservedManualRun(
 
 const TeamAutomationsTab: React.FC<TeamAutomationsTabProps> = ({
   members = [],
+  routeMemberId: routeMemberIdInput = "",
   scopeId,
   serviceIdentitiesLoading = false,
   teamId,
@@ -687,6 +689,7 @@ const TeamAutomationsTab: React.FC<TeamAutomationsTabProps> = ({
   const intl = useIntl();
   const queryClient = useQueryClient();
   const { token } = theme.useToken();
+  const routeMemberId = trimText(routeMemberIdInput);
   const automatableMembers = React.useMemo(
     () => members.filter((member) => member.canAutomateMember),
     [members],
@@ -726,8 +729,9 @@ const TeamAutomationsTab: React.FC<TeamAutomationsTabProps> = ({
     timezone: resolveDefaultTimezone(),
   }));
   const scheduleQueryKey = React.useMemo(
-    () => ["scheduled-dispatches", "team", scopeId, teamId] as const,
-    [scopeId, teamId],
+    () =>
+      ["scheduled-dispatches", "team", scopeId, teamId, routeMemberId] as const,
+    [routeMemberId, scopeId, teamId],
   );
   const serviceKeyToMember = React.useMemo(() => {
     const next = new Map<string, TeamAutomationMemberRow>();
@@ -750,7 +754,10 @@ const TeamAutomationsTab: React.FC<TeamAutomationsTabProps> = ({
     queryFn: () =>
       scheduledDispatchApi.listAll({
         includeTotalCount: true,
+        ...(routeMemberId ? { memberId: routeMemberId } : {}),
+        scopeId,
         take: scheduleListTake,
+        teamId,
       }),
     queryKey: scheduleQueryKey,
     retry: (failureCount) => failureCount < scheduleListRetryLimit,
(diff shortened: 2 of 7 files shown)


apps/aevatar-console-web/src/pages/teams/tabs/TeamAutomationsTab.tsx
```diff
@@ -750,7 +754,10 @@ const TeamAutomationsTab: React.FC<TeamAutomationsTabProps> = ({
     queryFn: () =>
       scheduledDispatchApi.listAll({
         includeTotalCount: true,
+        ...(routeMemberId ? { memberId: routeMemberId } : {}),
```
potter-sun, line 768: [P1] Keep the mutation target inside the member-scoped surface

When `routeMemberId` is present, this query is restricted to member A, but the create form still populates its selector from every `automatableMembers` entry and `selectedMember` falls back to the first member when the route ID is stale. A user on A’s canonical page can therefore create for B (or an invalid member route can create for the fallback member); `showCreatedScheduleFeedback` then keeps B as an optimistic row, while every refresh remains filtered to A and can never reconcile it. Resolve the route member against the roster and restrict/lock creation to that member on member-scoped routes (or show not-found for an invalid member), with distinct A/B IDs in the regression test.

apps/aevatar-console-web/src/pages/teams/detail.tsx
```diff
@@ -1993,6 +1993,7 @@ const TeamDetailPage: React.FC = () => {
           serviceRevisionId: row.serviceRevisionId,
           workflowSupported: row.workflowSupported,
         }))}
+        routeMemberId={routeState.memberId}
```
potter-sun: [P1] Derive this filter only from the canonical member path

`routeState.memberId` is not path-only: `readTeamDetailRouteState` still falls back to `params.get("memberId")`. This means `/scopes/scope-1/teams/t-alpha?tab=automations&memberId=m-alpha` now sends `memberId=m-alpha` to `/api/schedules`, silently turning the Team collection into a member collection and reactivating the forbidden hidden compatibility route. It also contradicts this PR description, which says `memberId` is passed only on member-scoped routes. Please remove the query fallback or expose a separate path-derived member ID for this prop, and add a regression test proving the Team URL omits `memberId` while the canonical `/members/:memberId/automations` path includes it.

apps/aevatar-console-web/src/pages/teams/tabs/TeamAutomationsTab.tsx
```diff
@@ -726,8 +729,9 @@ const TeamAutomationsTab: React.FC<TeamAutomationsTabProps> = ({
     timezone: resolveDefaultTimezone(),
   }));
   const scheduleQueryKey = React.useMemo(
-    () => ["scheduled-dispatches", "team", scopeId, teamId] as const,
-    [scopeId, teamId],
+    () =>
+      ["scheduled-dispatches", "team", scopeId, teamId, routeMemberId] as const,
```
potter-sun, line 744: [P1] Scope optimistic rows together with the query cache

The server cache key is now member-scoped, but `pendingCreatedSchedules` is still component-global and `teamSchedules` merges every pending row into the current query result by schedule ID only. If the route component is reused while navigating member A -> member B after a create, A’s pending schedule resurfaces on B’s page (B’s backend result cannot contain that ID), including edit/run/delete actions. Key/reset the optimistic state on `[scopeId, teamId, routeMemberId]`, or record the owner member on each pending row and filter it against the active route; add a navigation regression test with distinct member/service IDs.

apps/aevatar-console-web/src/pages/teams/detail.tsx
AbigailDeng: Fixed in d620e69bd.

`readTeamDetailRouteState` now exposes a separate path-only `routeMemberId`. The existing query-derived member candidate remains separate for the older Team selection behavior, while the schedule query, cache key, member surface, and component scope key consume only `routeMemberId`.

Regression coverage now proves `/scopes/scope-1/teams/t-alpha?memberId=member-team-alpha&tab=automations` omits `memberId` from `listAll`, while `/members/member-team-alpha/automations` includes it.

apps/aevatar-console-web/src/pages/teams/tabs/TeamAutomationsTab.tsx
AbigailDeng: Fixed in d620e69bd.

On member-scoped automation routes, the member roster is now reduced to the exact path-derived `routeMemberId` before resolving `automatableMembers`, `selectedMember`, schedule ownership, and form options. The target selector is locked on that surface. An unknown or non-automatable route member now resolves to no target, so creation stays disabled instead of falling back to the first Team member.

The regression uses distinct `member-team-alpha` / `member-team-beta` and `alpha-service` / `beta-service` identities, verifies the member target is locked, and verifies the route transition resolves the beta target rather than retaining alpha.

apps/aevatar-console-web/src/pages/teams/tabs/TeamAutomationsTab.tsx
AbigailDeng: Fixed in d620e69bd.

`TeamAutomationsTab` is now keyed by `[scopeId, teamId, routeMemberId]`. A surface transition remounts the local automation state as one unit, clearing pending-created rows together with delete/enable overrides, manual-run feedback, form state, modal state, highlights, and timers, while React Query continues to use its member-scoped cache key.

The new same-render navigation regression creates an unreconciled optimistic row on member alpha, navigates through `history.push` to member beta, verifies the beta-scoped backend query, and verifies the alpha row and target do not leak into beta.

PR #2996, opened 2026-07-28
Complete canonical NyxID operation admission

eanz17:
## 问题与方案

Closes #2984.
Fixes #2944.

- 将 workflow authoring 收敛为 step 级 typed NyxID selector：只保存 exact user_service_id 与 operation_id；slug、method、path template、schema、source stamp 与 digest 均由服务端 proof 生成。
- 引入 external-capability-admission.v3，以 call-site scoped invocation_admissions 作为唯一当前事实；持久化 v2 plan 在 prepare/publish/rebind 时返回 typed CAPABILITY_ADMISSION_REBIND_REQUIRED。
- 把 proof 从 definition actor 贯通到 run actor state、WorkflowToolExecutionRequest、typed tool context 与 NyxIdOperationRequestBuilder。ordinary、nested、foreach/for_each/foreach_llm、while/loop 共用同一 admission/runtime 主链。
- NyxID 请求只从 proof template 与 path_params/query/headers/body/response_mode 构造；缺 proof、identity/route/digest drift 或非法参数均在 HTTP dispatch 前 fail closed。NyxID Proxy wire 不携带 operation_id 或 contract_digest。
- 保持 x-aevatar-tool allowlist fail closed，并区分 missing/duplicate operationId、unknown/unallowlisted operation、header/body policy rejection。
- Studio member bind 返回安全的 typed readiness/remediation，不泄漏 credential、raw OpenAPI、外部 response body 或内部异常。
- 增加 #2944 message resource、approval instance 与 foreach fixture acceptance coverage，并保留通用 sub_param_ 语义。

## 影响路径

- NyxID connected-service/OpenAPI discovery、proof generation 与 proxy request builder。
- Workflow capability proto、definition/run actor state、compiler、foreach/while、tool execution adapter。
- Studio bind HTTP error mapping，以及 service revision persisted revalidation。
- Canon 文档：NyxID authority chain、author selector、v3 migration、proof handoff、runtime parameter model 与 A-prime update option。

## 验证

以下命令均在 rebase 到最新 origin/feature/integrate 后执行并返回 EXIT=0：

- dotnet build aevatar.slnx --nologo（0 errors）
- dotnet test aevatar.slnx --nologo（所有执行项目 0 failures）
- bash tools/ci/architecture_guards.sh
- bash tools/ci/workflow_binding_boundary_guard.sh
- bash tools/docs/lint.sh（82 files，0 errors）
- bash tools/ci/projection_route_mapping_guard.sh
- git diff --check

额外定向验证：Ornn 230/230、Workflow Application 440/440、Integration 412 passed（4 个既有显式跳过）、Workflow Host API 886/886、Studio 1421/1421、AI 1838/1838。

构建仍报告仓库已有的非阻断告警类别，包括 NU1903 Microsoft.OpenApi 2.0.0、NU1510、obsolete/nullability、unused proto import 与 analyzer complexity/coupling；无构建错误。

## Review follow-up (2026-07-29)

- Fail closed when a proof-bound `file_artifact` request is not an admitted GET.
- Enforce required headers and validate path/query/header scalar values against the admitted type/enum schema.
- Reparse root and inline workflow YAML through the canonical parser during Enforce startup inventory; pure internal workflows may omit an admission plan.
- Describe proof-bound runtime slots in `nyxid_proxy`'s tool contract without forcing legacy route fields.
- Keep the broad PR composition explicit: this branch also carries content-artifact authority, channel/NyxID chat recovery, managed Codex recovery, backend observatory/approval, and console authorization workstreams already listed in the changed canon/spec/runbook files.


eanz17 (commented):
## Code review — PR #2996

Overall direction of the admission/proof chain is solid: step-scoped v3 plan → run state → tool context → `NyxIdOperationRequestBuilder` is the right shape, and compiled external call sites fail closed in `ToolCallModule` independent of Shadow/Enforce.

Found several concrete correctness gaps that should be fixed (or explicitly accepted) before merge. (GitHub blocks REQUEST_CHANGES on own PRs and this PR's diff is too large for reliable inline line anchors, so notes are file-referenced.)

### Should-fix before merge

#### 1. Admitted `file_artifact` ignores HTTP method
- `src/Aevatar.AI.ToolProviders.NyxId/Tools/NyxIdProxyTool.cs` — `ExecuteAdmittedFileArtifactAsync` (~L367–408)
- Always calls `ProxyGetBinaryResponseAsync` (GET) without checking `request.Method`
- Legacy path correctly rejects non-GET (`file_artifact_requires_get` ~L432)
- `NyxIdOperationRequestBuilder` (~L116) only rejects body + `file_artifact`, not non-GET methods
- **Risk:** non-GET proof + binary response policy + `response_mode=file_artifact` silently GETs the path → method drift vs proof
- **Fix:** fail closed unless admitted method is GET; add regression test

#### 2. Required headers are not enforced
- `NyxIdOperationRequestBuilder.BuildHeaders` (~L351–396)
- Early-return on empty `headers` slot; never checks `HeaderParameters.Required`
- Query/body enforce required; headers do not
- **Fix:** mirror query required validation + unit test for missing required header

#### 3. Path/query/header values skip schema validation
- Proof stores type/enum schema; only body uses `ValidateSchema`
- Path only safety-encodes; query/header only require scalar text
- Weakens the PR claim “build only from proof + typed values”
- **Fix:** validate scalar path/query/header values against declared schema (enum/integer/number/boolean at minimum)

#### 4. Enforce startup inventory is circular + null-plan harsh
- `NyxIdWorkflowAdmissionEnforcementStartupGuard.HasValidV3Plan` (~L177–195)
- Passes `plan.InvocationAdmissions.Select(ToExpectedInvocation)` as expected → call-site↔YAML matching is tautological
- Does not re-parse workflow YAML (unlike `RevalidatePersistedAsync`)
- `plan is null` fails every definition/run without a v3 plan, including pure non-NyxID workflows → flipping Enforce may refuse host start without full backfill
- **Fix:** parse YAML for real expected invocations; allow null/empty plan when there are no external call sites

### Medium

#### 5. `ParametersSchema` still free-form while admitted path is slot-based
- `NyxIdProxyTool.ParametersSchema` (~L123–160) still requires `service_id`/`slug`/`path` and free-form `method`/`body`
- Admitted path only accepts `path_params`/`query`/`headers`/`body`/`response_mode` and rejects everything else (`NYXID_OPERATION_ARGUMENT_NOT_SUPPORTED`)
- Agents will emit rejected shapes when proof is present (fail-closed, but looks broken in prod)
- **Fix:** document admitted slots or emit proof-derived schema on admitted call sites

#### 6. PR composition / blast radius
- ~35k LOC / ~93 commits / 300+ files
- Title/body focus on NyxID operation admission, but diff also includes content-artifacts, channel recovery, admin observatory, managed codex, NyxID chat recovery, etc.
- Hard to review or roll back as one atomic unit — please expand PR description workstreams or split follow-ups

### Positive
- Fail-closed argument slots, path traversal defenses, sensitive-header rejection look deliberate
- Workflow-side admission from actor-owned run state (no process registry) matches architecture rules
- Core CI gates green at review time (coverage still pending)

**Recommendation:** fix 1–3 before merge; land 4–5 with Enforce rollout; address 6 in description or structure.

src/Aevatar.AI.ToolProviders.NyxId/Tools/NyxIdProxyTool.cs
```diff
@@ -239,6 +320,103 @@ public async Task<string> ExecuteAsync(string argumentsJson, CancellationToken c
...
+            return FileArtifactError("file_artifact_ingress_unavailable", "Host has not registered workflow file artifact ingress.");
+
+        var context = AgentToolRequestContext.Current;
+        var workflowRuntime = context?.WorkflowRuntime ?? AgentWorkflowRuntimeContext.Empty;
+        var callerScopeId = Normalize(context?.Caller.ScopeId);
+        var ownerRunId = Normalize(workflowRuntime.ParentRunId);
+        if (!workflowRuntime.HasManagedParent || callerScopeId == null || ownerRunId == null)
+            return FileArtifactError("managed_workflow_context_required", "response_mode=file_artifact requires a managed workflow runtime context and caller scope.");
+
+        var response = await _client.ProxyGetBinaryResponseAsync(
```
eanz17, line 448: **Bug:** admitted `file_artifact` path always GETs via `ProxyGetBinaryResponseAsync` and never checks `request.Method`. Legacy path rejects non-GET (`file_artifact_requires_get`). Fail closed unless admitted method is GET; add a non-GET proof + file_artifact regression test.

src/Aevatar.AI.ToolProviders.NyxId/Tools/NyxIdOperationRequestBuilder.cs
```diff
@@ -0,0 +1,559 @@
...
+    }
+
+    private static Dictionary<string, string>? BuildHeaders(
+        AgentToolOperationAdmission admission,
+        JsonElement root,
+        out NyxIdOperationRequestFailure? failure)
+    {
+        failure = null;
+        var supplied = ReadObjectSlot(root, HeadersSlot, out failure);
+        if (failure is not null || supplied.Count == 0)
```
eanz17: **Required headers not enforced:** early-return when `headers` slot empty; never checks `HeaderParameters.Required`. Query/body enforce required — headers should too.

src/Aevatar.Mainnet.Host.Api/WorkflowAdmission/NyxIdWorkflowAdmissionEnforcementStartupGuard.cs
```diff
@@ -0,0 +1,246 @@
...
+        }
+
+        try
+        {
+            WorkflowCapabilityAdmissionPlanIntegrity.ValidateOrThrow(
+                plan,
+                workflowYaml ?? string.Empty,
+                inlineWorkflowYamls,
+                plan.ExecutionMode,
+                plan.InvocationAdmissions.Select(ToExpectedInvocation));
```
eanz17: **Circular Enforce inventory:** `expectedInvocations` is rebuilt from `plan.InvocationAdmissions`, so call-site↔YAML matching always passes. Re-parse YAML like revalidate; also `plan is null` blocks pure non-NyxID definitions on Enforce.

src/Aevatar.AI.ToolProviders.NyxId/Tools/NyxIdOperationRequestBuilder.cs
```diff
@@ -0,0 +1,559 @@
...
+            var name = template[(open + 1)..close];
+            if (!supplied.TryGetValue(name, out var value))
+            {
+                failure = new NyxIdOperationRequestFailure(
+                    "NYXID_OPERATION_PATH_PARAMETER_MISSING",
+                    $"path parameter '{name}' is required by the admitted operation.");
+                return string.Empty;
+            }
+
+            var segment = EncodePathSegment(name, value, out failure);
```
eanz17, line 252: **Path/query/header skip schema validation:** proof stores type/enum on each parameter, but only body runs ValidateSchema. Path only safety-encodes; query/header only require scalar text. An admitted integer/enum parameter can be any string. Validate scalars against declared schema before encoding (at least enum/integer/number/boolean).

src/Aevatar.AI.ToolProviders.NyxId/Tools/NyxIdOperationRequestBuilder.cs
```diff
@@ -0,0 +1,559 @@
...
+
+            var headers = BuildHeaders(admission, root, out var headerFailure);
+            if (headerFailure is not null)
+                return new NyxIdOperationRequestBuildResult(null, headerFailure);
+
+            var body = BuildBody(admission, root, out var bodyFailure);
+            if (bodyFailure is not null)
+                return new NyxIdOperationRequestBuildResult(null, bodyFailure);
+
+            if (responseMode == FileArtifactResponseMode && body is not null)
```
eanz17, line 124: **Missing GET-only guard for file_artifact on admitted builder path.** This rejects body + file_artifact but not non-GET methods. Downstream ExecuteAdmittedFileArtifactAsync hard-codes GET, so a non-GET proof can still reach a GET call. Fail closed when response_mode=file_artifact and admitted method is not GET.

src/Aevatar.AI.ToolProviders.NyxId/Tools/NyxIdProxyTool.cs
eanz17: Fixed in dc0674146. Proof-bound file_artifact now fails closed in the shared request builder unless the admitted method is GET; the regression test confirms no HTTP request or artifact ingress occurs.

src/Aevatar.AI.ToolProviders.NyxId/Tools/NyxIdOperationRequestBuilder.cs
eanz17: Fixed in dc0674146. BuildHeaders now enforces required admitted headers before request construction; a missing required header returns NYXID_OPERATION_HEADER_MISSING and sends no HTTP request.

src/Aevatar.Mainnet.Host.Api/WorkflowAdmission/NyxIdWorkflowAdmissionEnforcementStartupGuard.cs
eanz17: Fixed in dc0674146. The Enforce guard now uses the canonical IWorkflowDefinitionParser to reparse root and inline YAML and validates the persisted plan against those parsed call sites. A missing or empty plan is allowed only when the YAML has no external call sites.

src/Aevatar.AI.ToolProviders.NyxId/Tools/NyxIdOperationRequestBuilder.cs
eanz17: Fixed in dc0674146. Path, query, and header values now reuse the existing schema validator with location-specific error codes, covering string enums, integer, number/enums, and boolean values before encoding or dispatch.

src/Aevatar.AI.ToolProviders.NyxId/Tools/NyxIdOperationRequestBuilder.cs
eanz17: Fixed in dc0674146. The shared admitted request builder rejects response_mode=file_artifact for every non-GET proof before token resolution or downstream work; the new regression test covers the method-drift case.

src/Aevatar.AI.ToolProviders.NyxId/Tools/NyxIdOperationRequestBuilder.cs
```diff
@@ -0,0 +1,665 @@
...
+        if (extra is not null)
+        {
+            failure = new NyxIdOperationRequestFailure(
+                "NYXID_OPERATION_PATH_PARAMETER_UNKNOWN",
+                $"path parameter '{extra}' is not declared by the admitted operation.");
+            return string.Empty;
+        }
+
+        var builder = new StringBuilder();
+        var template = admission.PathTemplate ?? string.Empty;
```
eanz17, line 204: **[P1] Normalize proof-bound templates before building the NyxID URL.** The OpenAPI path is copied into the admission proof without validating static segments, and this builder preserves those segments verbatim. A service spec can publish a path such as `../../../api-keys` (or encoded `%2e%2e`); `BuildProxyUrl` then creates `/api/v1/proxy/s/{slug}/../../../api-keys`, which `HttpRequestMessage` normalizes to `/api/v1/api-keys` while retaining the NyxID Bearer token. The dynamic-parameter checks do not cover this static part. Reuse the existing `NormalizeExactProxyPath`/dot-segment guard for the completed resource path (before appending query values), and add a malicious OpenAPI-template test that asserts zero HTTP requests.

src/Aevatar.AI.ToolProviders.NyxId/Tools/NyxIdOperationRequestBuilder.cs
```diff
@@ -0,0 +1,665 @@
...
+        if (missing is not null)
+        {
+            failure = new NyxIdOperationRequestFailure(
+                "NYXID_OPERATION_QUERY_PARAMETER_MISSING",
+                $"query parameter '{missing.Name}' is required by the admitted operation.");
+            return string.Empty;
+        }
+
+        var pairs = new List<string>();
+        foreach (var name in supplied.Keys.OrderBy(static key => key, StringComparer.Ordinal))
```
eanz17, line 381: **[P1] Reject NyxID-reserved query names on the proof-bound path.** This loop accepts every query name present in the external OpenAPI proof, including `_nyxid_*`. The exact-service request path already rejects that prefix case-insensitively, and the canon says callers may not provide any such control query. Here `BuildProxyUrl` only removes the exact lowercase `_nyxid_via` spelling before appending its own route, so `_NYXID_VIA` or any other reserved control survives to the NyxID proxy. Fail admission/build when a declared or supplied query name starts with `_nyxid_` (ordinal-ignore-case), and cover an OpenAPI-declared reserved name with zero HTTP dispatch.

src/Aevatar.AI.ToolProviders.NyxId/Tools/NyxIdProxyTool.cs
```diff
@@ -239,6 +329,103 @@ public async Task<string> ExecuteAsync(string argumentsJson, CancellationToken c
...
+
+        var request = build.Request!;
+        var token = AgentToolRequestContext.NyxIdAccessToken;
+        var orgToken = AgentToolRequestContext.NyxIdOrgToken;
+        if (string.IsNullOrWhiteSpace(token))
+        {
+            return request.FileArtifact
+                ? FileArtifactError("missing_nyxid_access_token", "No NyxID access token available. User must be authenticated.")
+                : """{"error":"No NyxID access token available. User must be authenticated."}""";
+        }
```
eanz17, line 366: **[P1] Revalidate the admitted operation against live NyxID authority before dispatch.** This only selects a token: with no distinct org token it performs no live read, and otherwise it merely scans a list for the service ID. A persisted plan can therefore outlive a UserService disable/reroute/slug change or effective-OpenAPI change; the old proof (including its old approval/risk decision) is then used to send its method/path to the new backend. `ContractDigest` is never consumed anywhere in the runtime path. Before any proxy request, reuse the exact `/keys/{id}` authority revalidation (active/access/slug/route) and re-fetch/recompute the selected operation digest, failing closed on drift; add route and contract-drift tests asserting zero proxy dispatch.

agents/Aevatar.GAgents.Channel.Identity/Endpoints/IdentityOAuthEndpoints.cs
```diff
@@ -681,12 +752,98 @@ internal static async Task<IResult> HandleAevatarOAuthClientRebuildCoreAsync(
...
+        var authorization = await AuthorizeRebuildAsync(http, adminAuthorizer, logger, ct)
+            .ConfigureAwait(false);
+        if (authorization.Rejection is not null)
+            return authorization.Rejection;
+
+        CommandDispatchResult<ChannelIdentityOAuthAcceptedReceipt, ChannelIdentityOAuthDispatchError> accepted;
+        try
+        {
+            accepted = await rotateDispatch
+                .DispatchAsync(new RotateAevatarOAuthClientHmacKeyCommand(), ct)
```
eanz17: [P1] Make the destructive rotation request idempotent

This POST has no caller-supplied idempotency key or expected current `kid`, and `ChannelIdentityOAuthCommandDispatch` generates a fresh command id for every invocation. If the 202 response is lost and the operator/proxy retries the same request, the actor rotates twice: v1→v2 demotes v1, then v2→v3 replaces the only previous-key slot with v2. Every still-valid state token signed by v1 then fails immediately, violating the documented 5-minute grace window precisely on the recovery path. Please bind the operation to a stable idempotency key / expected current kid and have the actor no-op an already-applied request.

src/Aevatar.Mainnet.Host.Api/BackendConsole/admin.html
```diff
@@ -3211,19 +3093,208 @@
...
+  var feedback=(state.feedback||'').trim();
+  if(!approved&&!feedback){ state.error='请填写驳回反馈'; if(rerender) rerender(); return false; }
+  state.submitting=true; state.error=''; if(rerender) rerender();
+  var body={stepId:approval.stepId,approved:!!approved};
+  if(!approved) body.userInput=feedback;
+  try{
+    await adminJson('/api/scopes/'+encodeURIComponent(run.scope)+'/runs/'+encodeURIComponent(run.id)+':resume',{
+      method:'POST',headers:{'Content-Type':'application/json'},body:JSON.stringify(body)
+    });
+    state.accepted=true; state.rejecting=false; return true;
```
eanz17, line 3251: [P1] Clear the optimistic approval latch from committed state

A successful call here is only the resume endpoint’s 202 inbox-admission ACK, but `state.accepted` is keyed only by `run.id` and is never cleared while any approval remains active. If the command loses a race with the approval timeout (and is ignored) or this run later reaches a second `human_approval` step, polling continues to return an active approval yet the buttons remain permanently disabled with “decision accepted”. The operator cannot approve the real pending step without reloading the whole page. Key the transient latch by the approval/step identity and clear it when refreshed committed state still shows that step pending or changes to a new step; 202 alone must not become UI completion state.

agents/Aevatar.GAgents.NyxidChat/NyxIdChatConversationGAgent.cs
```diff
@@ -54,7 +65,47 @@ protected override NyxIdChatConversationGAgentState TransitionState(
...
+                throw;
+            }
+            catch (Exception exception)
+            {
+                Logger.LogWarning(
+                    "NyxIdChat pending history reservation recovery failed: actor={ActorId} delivery={DeliveryId} exceptionType={ExceptionType}",
+                    Id,
+                    pendingReservation.DeliveryId,
+                    exception.GetType().Name);
+                return;
```
eanz17: [P1] Schedule a durable retry instead of swallowing reservation recovery failure

This catch returns from activation without recording/scheduling any retry. Because activation then succeeds, the actor can remain active indefinitely with `HistoryDeliveryReservation.Dispatched == false`; the return also skips both the pending-terminal continuation below and `NyxIdChatRecoveryRequestedSignal` for a requested/running operation. A transient chat-history outage during recovery therefore leaves a terminal delivery or the whole turn wedged until an unrelated message/passivation happens. The initialization and terminal outboxes already advance a typed attempt and schedule a durable self-timeout on failure; reservation recovery needs the same actor-owned retry path (and must not silently skip operation recovery).

agents/Aevatar.GAgents.NyxidChat/ConversationReplyGenerator.cs
```diff
@@ -1139,11 +1693,24 @@ private static ConversationHistoryEntry ToConversationHistoryEntry(ChatMessage m
...
         entry.ToolCalls.AddRange((message.ToolCalls ?? []).Select(ToConversationToolCallEntry));
         return entry;
     }
 
+    private static Aevatar.AI.Abstractions.ChatContentPart ToPersistedContentPart(ContentPart part)
+    {
+        var persisted = ContentPartProtoMapper.ToProto(part);
+        if (persisted.Kind == Aevatar.AI.Abstractions.ChatContentPartKind.Text &&
+            HasFileRefIdentity(persisted.FileRef))
+            persisted.Text = string.Empty;
```
eanz17, line 1735: [P1] Keep replayable document text after the artifact expires

This erases the extracted PDF/text payload and leaves durable history with only a workflow-file ref. Those artifacts have a finite TTL (the filesystem backend defaults to one day and rejects expired descriptors), while conversation history has no matching TTL. Every later turn reopens referenced parts from the last 10 history entries in `AgentRunReplyGenerationExecutor.MaterializeFileRefMessagesAsync`; once this ref expires, materialization throws before the LLM call, and the conversation cannot produce successful history to advance the bad entry out of the window. Persist the bounded extracted text (or make expired refs replay as an honest unavailable-attachment marker) instead of removing the only replayable content.

src/workflow/Aevatar.Workflow.Integration.AI/WorkflowRoleGAgent.cs
```diff
@@ -294,6 +302,9 @@ private async Task<WorkflowIntentReplayRecord> ExecuteWorkflowIntentStreamingCha
         var inputParts = ResolveWorkflowRequestInputParts(request);
         var llmControl = LLMControlContextMapper.FromPayload(request.LlmControl);
         var toolContext = llmControl.ToToolContext(AgentToolExecutionContextMapper.FromPayload(request.ToolContext));
+        var turnCatalog = await BuildRequestToolCatalogAsync(intent.AgentToolScope, toolContext, streamCt);
```
eanz17, line 513: [P1] Preserve pending approval instead of completing the LLM step

The request-local tool set added here includes approval-gated tools such as `nyxid_service_update`, `nyxid_service_route`, `nyxid_service_delete`, and write operations. When one yields, `ChatStreamAsync` returns an `ApprovalRequired` receipt and stops, but this custom workflow path only collects that receipt and then `HandleWorkflowLlmExecutionIntent` unconditionally publishes `WorkflowLlmInvocationCompletedEvent { Success = true }`. Unlike the normal `RoleGAgent` handler, it never calls `DetectPendingApproval`, persists `PendingToolApprovalState`, or emits `ToolApprovalRequestEvent`; `LLMCallModule` therefore completes the step (usually with empty output), and the approval cannot resume the call. Route this handler through the existing pending-approval continuation or explicitly surface/suspend on the receipt before publishing success; add a Studio LLM-turn write-tool test that approves and resumes.

src/Aevatar.AI.ToolProviders.NyxId/Tools/NyxIdOperationRequestBuilder.cs
eanz17: Fixed in e53afe559. The completed proof-bound resource path now goes through NormalizeExactProxyPath before query assembly; literal and encoded dot-segment traversal fails with NYXID_OPERATION_PATH_TEMPLATE_INVALID, and regression tests assert zero HTTP requests.

src/Aevatar.AI.ToolProviders.NyxId/Tools/NyxIdOperationRequestBuilder.cs
eanz17: Fixed in e53afe559. Declared or supplied query names starting with _nyxid_ are now rejected case-insensitively before dispatch with NYXID_OPERATION_QUERY_PARAMETER_FORBIDDEN; the tests cover externally declared reserved names and zero HTTP requests.

src/Aevatar.AI.ToolProviders.NyxId/Tools/NyxIdProxyTool.cs
eanz17: Fixed in e53afe559. The current branch's authoritative NyxID surface is the live MCP catalog, so every proof-bound call now re-fetches /api/v1/mcp/config with the caller token, matches the exact service ID and slug, then the exact endpoint ID and contract digest. Authority or contract drift fails closed before proxy or file dispatch, with zero-dispatch tests for both cases.

agents/Aevatar.GAgents.Channel.Identity/Endpoints/IdentityOAuthEndpoints.cs
eanz17: Fixed in e53afe559. Rotation now requires exactly one Idempotency-Key plus a strong quoted If-Match containing the expected current kid. The actor persists the applied idempotency key and no-ops duplicate or stale-kid requests, preserving the previous-key grace slot; endpoint and actor regression tests cover retries and invalid preconditions.

src/Aevatar.Mainnet.Host.Api/BackendConsole/admin.html
eanz17: Fixed in e53afe559. The optimistic latch is keyed by run ID, step ID, and approval request timestamp. Each committed detail refresh reconciles it: a still-pending same approval clears the 202-only accepted latch, and a different approval gets fresh state. The console test covers both the lost-race case and a second approval step.

agents/Aevatar.GAgents.NyxidChat/NyxIdChatConversationGAgent.cs
eanz17: Fixed in e53afe559. Reservation recovery failures now commit an actor-owned retry attempt and schedule a durable self-timeout keyed by delivery ID and attempt. Retry handling rejects stale signals, and activation continues outstanding operation recovery instead of returning early; the regression test exercises failure, continued recovery, and successful retry.

agents/Aevatar.GAgents.NyxidChat/ConversationReplyGenerator.cs
eanz17: Fixed in e53afe559. Historical expired or deleted attachment refs now materialize for the LLM as an honest unavailable-attachment marker instead of throwing and wedging the replay window. Durable history keeps the original ref, expired artifacts are not read, and the regression test verifies both behaviors.

src/workflow/Aevatar.Workflow.Integration.AI/WorkflowRoleGAgent.cs
eanz17: Fixed in e53afe559. The workflow LLM path now detects approval-required receipts, persists the typed continuation, emits the approval request, and suspends before publishing completion. Approval re-resolves the request-local tool, reissues fresh caller tokens without persisting bearer credentials, resumes the original workflow step, and can suspend again for a second gated tool; tests cover approval, denial, token refresh, and the second approval.

eanz17 (commented):
Addressed all eight review threads in e53afe559.

PR #3427, opened 2026-08-11
Preserve committed publication recovery failures

eanz17:
## Problem and solution

A workflow run can finish its steps successfully but fail while publishing the committed terminal state across a rollout. The recovery path treated that infrastructure failure as a completed run, which made committed Observatory state contradict the publication outcome.

This change preserves publication adapter failures as terminal infrastructure failures, makes the failure policy explicit, and covers both the execution boundary and fork-on-failure recovery behavior. It also refreshes the NyxID conformance source pin required by the current feature/integrate baseline.

## Impact paths

- Workflow execution terminal failure handling
- Workflow run recovery after committed publication failure
- NyxID assistant conformance source pin

## Verification

- Publication boundary tests: 18/18 passed
- Aevatar.Workflow.Core.Tests: 981/981 passed
- NyxID conformance guard: passed
- bash tools/ci/test_stability_guards.sh: passed
- bash tools/ci/architecture_guards.sh: passed
- git diff --check origin/feature/integrate...HEAD: passed

## Production acceptance

Production remains on image b58156d2, which does not contain the fix. Final acceptance requires deployment followed by a newer owner committed run; the existing HR-01 failure is intentionally not closed by this PR.

### DIFF
--- a/docs/contracts/nyxid-assistant-conformance/v1/sources.json
+++ b/docs/contracts/nyxid-assistant-conformance/v1/sources.json
@@ -2,15 +2,15 @@
   "schema_version": 1,
   "aevatar": {
     "repository": "https://github.com/AevatarAI/aevatar.git",
-    "revision": "29bf4e60d25cb16138b98584c900debb5a5669a6",
-    "contract_files_sha256": "0bf6cd57359afd7c04374f4dd0f3d368664db5663b2d18b5e209471e4ec2246a",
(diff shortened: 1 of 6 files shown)


eanz17:
Closing as superseded. The core change has an equivalent patch in feature/integrate as commit 29d56c451 (git cherry reports the original implementation commit as already applied). The only remaining unique commit on this branch is an obsolete NyxID conformance source-pin refresh, which must not replace the newer authoritative pin now in the base. No implementation is being discarded.

